PR #2, opened 2013-07-31
Bugfix. unrestrictedTraverse fails when no REQUEST present.

jcbrand:
unrestrictedTraverse shouldn't assume that a REQUEST attribute is always
available in the acquisition chain.

Products such as plone.app.async and collective.xmpp.core sometimes
asynchronously connect to the ZServer without an HTTP request.

They should still be able to use unrestrictedTraverse to traverse namespace adapters.

zope.traversing.namespace.namespacelookup doesn't require a Request object to
be available for it to do namespace adapter lookups.

Thanks to @reinhardt who noticed this issue when testing plone.app.async with an object accessed via a namespace adapter.

### DIFF
--- a/src/OFS/Traversable.py
+++ b/src/OFS/Traversable.py
@@ -221,8 +221,11 @@ class Traversable:
                         # Process URI segment parameters.
                         ns, nm = nsParse(name)
                         try:
-                            next = namespaceLookup(
-                                ns, nm, obj, aq_acquire(self, 'REQUEST'))
+                            request = aq_acquire(self, 'REQUEST')
+                        except AttributeError:
+                            request = None
+                        try:
+                            next = namespaceLookup(ns, nm, obj, request)
                             if IAcquirer.providedBy(next):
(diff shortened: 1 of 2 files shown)


tseaver:
I just [pushed Acquisition 2.13.9  to PyPI](https://pypi.python.org/pypi/Acquisition/2.13.9) with a fix for #1 (and a couple of other backported fixes).  @jcbrand, can you check whether the three-argument form of `aq_acquire` works for you with that fix?

jcbrand:
@tseaver Will do, however I'm currently in bed with the flu, so might take a while.

tseaver:
Hope you feel better soon -- take care!

jcbrand:
Thanks!

jcbrand:
@tseaver This pull request is from so long ago that we don't even use the original code which exposed this issue any more, so it's not something I can easily test without spending a significant amount of time doing some code archaeology.

I therefore propose we close this PR.

PR #5, opened 2013-11-03
Unicode Handling and Additonal complex query support.

flipmcf:
1: Abandoned idea to add 'encoding' parameter to function
- risk of clobbering someone using 'encoding' as a query-string key provided via **kwargs - will rethink
- setting encoding to utf8 if no other encoding is specified.
- utf8 seems hard-coded in ZPublisher marshalling, so...

2: Using 'config.zpublisher_default_encoding' to get configured encoding
- not sure about this.  ZPublisher seems to use utf8 exclusively, so maybe this is not needed
- zpublisher_default_encoding seems to control the html/xml content-type encoding - a different beast.

3: Rigorous testing of Lists-In-Records and other complex data structures:
- for better testing and less changes, I left the original test cases unmodified and added my own for Unicode

4: Re-write of complex_marshal for recursive solution.
- other option is to check value type in multiple places in the code, maintenance nightmares occur.
- extra test cases added for now supported RecordsInLists, ListsInLists, RecordsInRecords
- exercise Recursion Depth Exceeded - for fun.

Note:
   Lists In Lists can produce ambiguous results:  [sublist1, sublist2] produces the same output as sublist1+sublist2.  There are many easy workarounds for this case, tho.

### DIFF
--- a/src/ZTUtils/tests/testZope.py
+++ b/src/ZTUtils/tests/testZope.py
@@ -241,14 +241,6 @@ class UnicodeQueryTests(TestCase):
         self.assertEqual(tag, '<input type="hidden" name="foo:utf8:ustring" value="%s">' % (quote_unicode,))
 
 
-class ComplexMarshalRecursionTests(TestCase):
-    """Test using ComplexMarshal badly..."""
-
-    def testRecursionError(self):
-        list_=['a','b','c']
-        list_.append(list_)
-        self.assertRaises(RuntimeError, complex_marshal, [('this_is_bad', list_)])
-
 class UnicodeEncodingTests(TestCase):
     """ test behavior if encoding is specified """
     # def testMakeQueryUnicodeSpecifyEncoding(self):


flipmcf:
Neglected, but not forgotten.

flipmcf:
Noted the 2.6 failure.

src/ZTUtils/tests/testZope.py
```diff
@@ -35,6 +51,56 @@ def testMarshallListsInRecords(self):
...
+    def testMarshallRecordsInRecords(self):
+        '''Test marshalling records inside records'''
+        sub_record = {'sub1': 'deep', 'sub2': 1}
+        record = {'arg1':'top', 'arg2':sub_record, 'arg3':12}
+        result = complex_marshal([('r_record',record),])
+        self.maxDiff = 1000
+
+        #dictionaries dont' preserve order - use self.assertItemsEqual
+        self.assertItemsEqual(result,
+                              [('r_record.arg1', ':record', 'top'),
```
tseaver: For 2.6 compatibility, try sorting them.

tseaver:
@flipmcf I'd like to review again and merge this week before making a 2.12.23 release.  Can you revisit?

flipmcf:
My fixes are in, finally.

Tres, do we really need that 'recursion check' code?   

This code simply says:
    append a list to itself, recursively.
    throw it into a marshaller
    expect to get a system recursion error
      (in other words, don't expect the marshaller to gracefully deal with that kind of input)

So, we're really testing how python deals with an exceeded recursion depth at this point, not the marshaller at all.  It seems like a pointless test.

If the code improves at some point (the marshaller has it's own way of dealing with recursive data structures) this test will fail.  The best course of action at that point is to simply remove it.  Again, pointless test.

If you agree, we can simply remove it and move on with life.

flipmcf:
I'll do the merge.  Conflicts arise from flake8 and 2to3 compatibility.
I'll do my best to make flake8 not complain after I merge in the changes.

PR #17, opened 2015-02-10
Update setuptools and zc.buildout to currently-installable versions.

tseaver:
- setuptools 12.0.5
- zc.buildout 2.3.0


thet:
+1

This would need a change in versioning policy of all dependencies from the setuptools based one to PEP 440 based one due to this change: https://bitbucket.org/pypa/setuptools/commits/dcd552da643c4448056de84c73d56da6d70769d5#chg-setuptools/tests/test_resources.py

Basically this affects only development versioning numbers like old ``1.2.3.dev0`` to new ``1.2.3-dev0``. At least the Plone community has to switch, and I'll compile a PLIP and promote this idea.

mgedmin:
> Basically this affects only development versioning numbers like old 1.2.3.dev0 to new 1.2.3-dev0.

Huh?  `1.2.3.dev0` is a [PEP 440](https://www.python.org/dev/peps/pep-0440/)-compliant version number.  Why would it need to be changed?

thet:
oh that's absolutely true - i mixed something up. sorry. it's setuptools old ``1.2.3.a1``, pep 440 new ``1.2.3a1``, IIRC. i think my comment above is obsolete.

vincentfretin:
The current version of zc.buildout is 2.3.1, released two days after 2.3.0

tseaver:
I've bumped `zc.buildout` to 2.3.1.

PR #22, opened 2015-02-17
Queue additional warning filters at the beginning of the queue in order to allow overrides.

jensens:
Before this change custom waring filters were appended. Since python 2.7 there are defaults preventing custom filters to get grip if not prepended.
So with the zope.conf setting there is no possibility to override the defaults: The list is processed from the head to tail and first match wins. 
For details please read http://permalink.gmane.org/gmane.comp.web.zope.devel/30472

This PR fixes this.


tseaver:
I don't understand the "non-merge" of this, as the branch is clearly merged to master.

jensens:
all fine, thanks for merging!

PR #27, opened 2015-06-16
support IStreamIterator as WSGI response body

gweis:
I had troubles returning an IStreamIterator instance when using the WSGIPublisher.

This PR makes sure that no non iterator methods like seek, len and str are called when publishing an IStreamIterator.

Patch for master branch see: #28 

cheers

### DIFF
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -8,6 +8,11 @@ http://docs.zope.org/zope2/
 2.13.23 (unreleased)
 --------------------
 
+- Issue #27: Fix publishing of IStreamIterator. This interface does
+  not have seek or tell.
+  Introduce IUnboundStreamIterator to support publishing iterators
+  of unknown length.
+  
 - LP #1465432:  Ensure that WSGIPublisher starts / ends interaction at
   request boundaries (analogous to ZPublisher).  Backport from master.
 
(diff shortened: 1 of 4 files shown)


tseaver:
The `IStreamIterator` interface guarantees a `__len__` method:  we should use that method to set the `Content-Length` header.

The same comment applies to #28.

gweis:
Oh... totally overlooked that. 

I can update the pull request no problem. However, would there be a way to avoid exhausting an iterator with undefined length before passing returning it to the WSGI stack?

Maybe by setting the response.http_chunk attribute, or if there is already a content-length header set on the response?

tseaver:
Perhaps we should define another interface (`IUnboundStreamIterator` maybe?), which does not have a `__len__` method (`IStreamIterator` could then inherit from it, adding `__len__`).  This patch could then test both, adding the `Content-Lenght` header only when appropriate.

tseaver:
I can't merge until Travis is green.  Can you try the sorting fix I suggested?

tseaver:
@gweis I'd like to get your fix merged this week before making a 2.13.23 release.

gweis:
@tseaver This would be great if it cloud go into the next release. (esp. if it is that soon :) ).

I have updated the pull request. I can redo the pull request In case it is too messy by now (I have also merged in the latest changes).

If the pull request is fine, I'll port it to master branch as well.

tseaver:
Thank you!

PR #33, opened 2015-09-16
Don't bail on WindowsError which is not existing on non-windows systems.

icemac:
Pull-Request for master branch.

Discussion see https://github.com/zopefoundation/Zope/commit/87c5992e7e40604df206bb713e42bafd5919fe64


tseaver:
Thanks!

Issue #46, opened 2016-02-01
ipv6 tests fail on Travis


mauritsvanrees:
Zope job 102 from three months ago ran on the legacy Travis structure and passed fine:
https://travis-ci.org/zopefoundation/Zope/jobs/90092264

Two builds from the past few days have this message in Travis: "This job ran on our new platform for Precise builds. Please read [our blog post](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) for more information." Follow the link and one of the things you read is: "The major change that is coming with this migration is that local and external IPv6 networking will no longer be present."

And indeed those two recent builds fail with this error:

```
Error in test test_http_over_ipv6 (ZServer.tests.test_config.ZServerConfigurationTestCase)
Traceback (most recent call last):
  File "/opt/python/2.7.9/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/travis/build/zopefoundation/Zope/src/ZServer/tests/test_config.py", line 148, in test_http_over_ipv6
    server = factory.create()
  File "/home/travis/build/zopefoundation/Zope/src/ZServer/datatypes.py", line 88, in create
    logger_object=access_logger)
  File "/home/travis/build/zopefoundation/Zope/src/ZServer/HTTPServer.py", line 449, in __init__
    http_server.__init__(self, ip, port, resolver, logger_object)
  File "/home/travis/build/zopefoundation/Zope/src/ZServer/medusa/http_server.py", line 574, in __init__
    self.bind ((ip, port))
  File "/opt/python/2.7.9/lib/python2.7/asyncore.py", line 342, in bind
    return self.socket.bind(addr)
  File "/opt/python/2.7.9/lib/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 99] Cannot assign requested address
```

I wondered if the container based setup supports ipv6, but no, they do not:
https://docs.travis-ci.com/user/ci-environment/#Networking

So i guess we should allow this test to fail...

mauritsvanrees:
The related pull request was merged a month ago.

PR #45, opened 2016-02-01
Fix two issues

pbauer:
These were discovered while trying to use the newest zope-packages for Plone (https://github.com/plone/Products.CMFPlone/issues/1351)

1. Cache-invalidation for FactoryDispatcher was broken for packages that are added after first initialization
2. No longer try to get removed help-file PageTemplate_Edit.stx

PR #47, opened 2016-02-01
Remove Python 2.6 support and skip ipv6 test on Travis

mauritsvanrees:
Fixes issue #46.

Once this is accepted and merged, we can try pull request #45 again.


tseaver:
Could we just skip the test on Travis, e.g. via something like:

```python
    if not os.environ.get('TRAVIS'):
        def test_http_over_ipv6(self):
```

thet:
Shouldn't deprecated Python 2.6 support be dropped at all for Zope 4? Then we could use the nicer ``@unittest.skip`` and save some headache when encountering other Python 2.6 incompatibilities.

mauritsvanrees:
I have removed Python 2.6 from `.travis.yml` and `setup.py`. Seems fair enough for Zope 4.

I have updated the pull request to use `unittest.skipIf` and check the os environment for `TRAVIS`, which according to the docs should indeed be there.

tseaver:
@mauritsvanrees Thank you!

mauritsvanrees:
@tseaver Thanks for reviewing and merging!

PR #52, opened 2016-03-04
fix ReST rendering problem in changelog

jensens:
Rendering of RestructuredText on PyPI https://pypi.python.org/pypi/Zope2/2.13.24 is broken. 

This little change fixes it. Looks much better now.

I also found a missing `-r` parameter in the example provided by the change log line and added it.

Hint: [zest.releaser](https://pypi.python.org/pypi/zest.releaser/) is the swiff knife for PyPI releases and besides a streamlined release process it offers the `longtest` script which introspects setup.py and renders README.rst, CHANGES.rst and other involved files as one long file. It reports errors and opens the result in the browser.

### DIFF
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -33,8 +33,8 @@ http://docs.zope.org/zope2/
 
 - Provide a pip-compatible ``requirements.txt`` file for the release.  E.g.::
 
-  $ /path/to/venv/bin/pip install \
-    https://raw.githubusercontent.com/zopefoundation/Zope/2.13.23/requirements.txt
+      $ /path/to/venv/bin/pip install -r \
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
Looks good, I'll merge.

@tseaver: optionally login to pypi and edit the file manually so it looks good again. Just a suggestion.

BTW, I am one of the authors of the `longtest` command from `zest.releaser`. I still use it, but I am leaning towards using the `restview` package, which in turn uses the `readme_renderer` package. pip install it and do `restview --long-description` and it will show the result in the browser and update it live as you edit the underlying files.

tseaver:
@mauritsvanrees I've fixed the PyPI page.

PR #53, opened 2016-03-06
use newer versions of everything (except zope.testbrowser)

tomgross:
Update to ZODB Version 4


versions.cfg
```diff
@@ -43,4 +43,11 @@ repoze.who = 2.2
...
 zope.sequencesort = 4.0.1
 zope.testbrowser = 3.11.1
+ZODB = 4.2.0
+sphinx-rtd-theme = 0.1.9
+alabaster = 0.7.7
+Babel = 2.2.0
+snowballstemmer = 1.2.1
+
+# only for buildout dependencies
 ZODB3 = 3.11.0
```
tseaver: This part makes no sense:  we shouldn't need `ZODB3` for anything.

versions.cfg
```diff
@@ -43,4 +43,11 @@ repoze.who = 2.2
 RestrictedPython = 3.6.0
 zope.sequencesort = 4.0.1
 zope.testbrowser = 3.11.1
+ZODB = 4.2.0
+sphinx-rtd-theme = 0.1.9
+alabaster = 0.7.7
+Babel = 2.2.0
+snowballstemmer = 1.2.1
```
tseaver: I see that `alabaster` is a depenency of `Sphinx`:  are `Babel` and `snowballstemer` also?

tseaver:
Overall, LGTM

versions.cfg
tomgross: yes, they are. Probably it makes sense to keep these dependencies in an entry point of the egg and mark them in the versions.cfg

versions.cfg
tomgross: This is pulled in by the following dependencies:
https://github.com/zopefoundation/Products.BTreeFolder2/
https://github.com/zopefoundation/Products.ExternalMethod/
https://github.com/zopefoundation/Products.ZCatalog/
https://github.com/zopefoundation/Products.ZCTextIndex/

I propose to remove the explicit dependency on ZODB (3) in these packages because they all depend on Zope2 anyways.

thet: I imagine changing the ZODB3 dependency in these packages to ZODB is hairy, because of Zope 2.x depending on ZODB3.

+1 for removing the ZODB3 dependency if above is true.

thet:
Are these changes part of the "Plone on Zope 4" effort? https://github.com/plone/Products.CMFPlone/issues/1351

tomgross:
@thet yes and no. "Plone on Zope 4" is welcome to use it but there is no direct dependency.

thet:
@tomgross sure, will do. Especially, I'm looking forward for incorporating newest ZCatalog with ZODB 4 for Plone, which also means some special care about this None-indexing issue.

versions.cfg
tomgross: I created PRs for the changes proposed above.

setup.py
```diff
@@ -67,7 +67,7 @@ def _read_file(filename):
       'Record',
       'RestrictedPython',
       'ZConfig >= 2.9.2',
-      'ZODB3',
+      'ZODB',
```
tseaver: We should be adding a pin on `BTrees`, since this package imports it explicitly.

tomgross:
This will pass, after ZODB3 dependency is removed from all dependend packages.

tomgross:
Can be merged after
https://github.com/zopefoundation/AccessControl/pull/14
and
https://github.com/zopefoundation/Products.ZCatalog/pull/8

tseaver:
> Can be merged after zopefoundation/AccessControl#14 and zopefoundation/Products.ZCatalog#8
and branches in sources.cfg are adjusted

zopefoundation/AccessControl#15 (replaces zopefoundation/AccessControl#14) and zopefoundation/Products.ZCatalog#8 are merged.

tomgross:
All dependent branches have been merged. This is ready to be merged.

tomgross:
Travis agrees :smile:

tomgross:
Thanks @tseaver for review and merge!

PR #54, opened 2016-05-04
Apply hotfix 20160419 [to 2.13]

mauritsvanrees:
Plus remove a few extra docstrings.
Probably more can be found that are safe and good to remove, but this is a nice start.

### DIFF
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -8,7 +8,9 @@ http://docs.zope.org/zope2/
 2.13.25 (unreleased)
 --------------------
 
-- TBD
+- Removed docstrings from some methods to avoid publishing them.  From
+  Products.PloneHotfix20160419.  [maurits]
+
 
 2.13.24 (2016-02-29)
 --------------------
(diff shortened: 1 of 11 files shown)


leorochael:
+1

tseaver:
Hmm, making a bunch of security-aware methods non-URL-accessible might have some backward-compatibility issues.  Is there a discussion of the threat model underlying the Plone hotfix somewhere?  The `zope-security` mailing list didn't receive any notice of the change.

mauritsvanrees:
This is a minor part of the Plone hotfix. If this was the only part, we probably would not have issued a hotfix. But we got a question about this and since we were creating one for more serious stuff, we thought we might as well include this. We have had more hotfixes which simply removed docstrings from methods that should not have had them, so I am in the mood for shooting docstrings. :-)

The 'threat model' here is basically: an attacker can use these methods to get information about a site, which might, potentially, point to a direction of attach. For example, when you can access `Type`, an attacker may say: ah, this is a Document, let's try attack D, and on this News Item let's try attack N. Probably the `Type` is somewhere in the html code, for example a class `document` on the body, but if not: let's not give potential attackers info that they have no business knowing.

hannosch:
Closed/merged via #63.

Issue #57, opened 2016-05-11
Security tab permission / role matrix security issue


gogobd:
There's a security issue hidden in the "Security" tab of zope. Administrators change permissions by selecting checkboxes in the permission / role matrix. The underlying form uses column and row numbers in its request. If somebody manages to change the number of available permissions while the administrator has not yet saved the changes there's a problem: permissions would not match up any more.
I've solved this by introducing hashes instead of numbers. Initially I thought I could use permission ids, but they may contain blanks. Hashes make sure that just every permission name will work.
The changes went into OFS and AccessControl.

hannosch:
Both PR's have been merged, so I think this is done. Please re-open if I misunderstood.

PR #58, opened 2016-05-11
Fixing permission handling 

gogobd:
Using hashed permission names instead of columns and rows numbers eliminates a security issue when the permission (or role) count changes. See <https://github.com/zopefoundation/Zope/issues/57> for more information.


gogobd:
This merge needs AccessControl to be updated as well, I'm making a new one.

PR #60, opened 2016-06-29
Optimize "OFS.ObjectManager.__contains__" method

malthe:
Optimized the `OFS.ObjectManager.__contains__` method to do the least amount of work necessary.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,9 @@ Bugs Fixed
 Features Added
 ++++++++++++++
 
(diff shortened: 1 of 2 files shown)


tseaver:
Thanks!

davisagli:
A side effect of this that I think is worth mentioning in the changelog is that a subclass that overrides `objectIds` will no longer automatically have its `objectIds` used for a containment check.

malthe:
That's true. It's an inherent risk with object hierarchies that aren't just abstract or final.

PR #72, opened 2016-09-14
Add support for optional 'SameSite' cookie attribute

EcrinDe:
As described in the definition document by the ietf:
https://tools.ietf.org/html/draft-west-first-party-cookies-07

"The 'SameSite' attribute allows servers to assert that a cookie
ought not to be sent along with cross-site requests. This assertion
allows user agents to mitigate the risk of cross-origin information
leakage, and provides some protection against cross-site request
forgery attacks."

### DIFF
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -903,6 +903,8 @@ class HTTPResponse(BaseResponse):
                 # and block read/write access via JavaScript
                 elif name == 'http_only' and v:
                     cookie = '%s; HTTPOnly' % cookie
+                elif name == 'same_site':
+                    cookie = '%s; SameSite=%s' % (cookie, v)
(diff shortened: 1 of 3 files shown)


src/ZPublisher/HTTPResponse.py
```diff
@@ -903,6 +903,8 @@ def _cookie_list(self):
                 # and block read/write access via JavaScript
                 elif name == 'http_only' and v:
                     cookie = '%s; HTTPOnly' % cookie
+                elif name == 'same_site':
+                    cookie = '%s; SameSite=%s' % (cookie, v)
```
tseaver: Please add a comment above line 906 explaining the purpose, and linking to the RFC.

EcrinDe:
requested changes have been made in @1c65616

tseaver:
@EcrinDe Thank you for the patch!

PR #81, opened 2016-12-05
Patch zope.interface to remove docstrings and avoid publishing.

mauritsvanrees:

See https://github.com/zopefoundation/zope.interface/pull/65


icemac:
@mauritsvanrees Should there be any tests?

mauritsvanrees:
There are tests in Plone, but they depend on z3c.form, which I don't want to introduce as test dependency.
I briefly tried to come up with a simpler test case, but that is hard.

icemac:
@mauritsvanrees I thought about a simple test making sure your code is run when Zope starts thus the doc strings of the interface classes are removed after starting Zope.

mauritsvanrees:
Well, that's a bit easier. :-) Done.

mauritsvanrees:
I have rebased to avoid a conflict in the changelog file. The patches for master and for the 2.13 branch have each gotten a positive review from a different person, so I will merge both.

PR #82, opened 2016-12-05
Patch zope.interface to remove docstrings and avoid publishing.

mauritsvanrees:

See https://github.com/zopefoundation/zope.interface/pull/65

### DIFF
--- a/src/Zope2/App/patches/publishing.py
+++ b/src/Zope2/App/patches/publishing.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+
+
+def delete_method_docstring(klass, method_name):
+    # Delete the docstring from the class method.
+    # Objects must have a docstring to be published.
+    # So this avoids them getting published.
+    method = getattr(klass, method_name, None)
+    if (method is not None and hasattr(method, 'im_func') and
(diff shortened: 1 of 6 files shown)


src/Zope2/App/patches/publishing.py
```diff
@@ -0,0 +1,57 @@
...
+for klass in [Element, Attribute, Interface, Method]:
+    try:
+        del klass.__doc__
+    except:
+        pass
+    for method_name in element_methods:
+        delete_method_docstring(klass, method_name)
+
+for method_name in interface_methods:
+    delete_method_docstring(Interface, method_name)
```
tseaver: Rather than patching at import time, please factor this into a module-scope function which gets called inside `Zope2.App.startup.startup()` (like `patch_persistent`, which could also move here).

mauritsvanrees: Done.

mauritsvanrees:
I have added a test to check that our patch indeed has effect.

mauritsvanrees:
I have rebased to avoid a conflict in the changelog file. The patches for master and for the 2.13 branch have each gotten a positive review from a different person, so I will merge both.

PR #83, opened 2016-12-08
Don't copy items the user is not allowed to view. [2.13]

mauritsvanrees:

### DIFF
--- a/src/OFS/CopySupport.py
+++ b/src/OFS/CopySupport.py
@@ -596,9 +596,9 @@ class CopySource(Base):
             # being copied, so it makes no sense to check any of its sub
             # objects.  It probably means we are in a test.
             return ob
-        return self._cleanupCopy(ob)
+        return self._cleanupCopy(ob, container)
 
-    def _cleanupCopy(self, cp):
+    def _cleanupCopy(self, cp, container):
         sm = getSecurityManager()
         ob = aq_base(self)
         if hasattr(ob, 'objectIds'):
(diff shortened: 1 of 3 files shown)


src/OFS/CopySupport.py
```diff
@@ -587,7 +587,35 @@ def _getCopy(self, container):
...
+        sm = getSecurityManager()
+        ob = aq_base(self)
+        if hasattr(ob, 'objectIds'):
+            for k in self.objectIds():
+                v = self._getOb(k)
+                if not sm.checkPermission('View', v):
+                    # We do not use cp._delObject, because this would fire
+                    # events that are needless for objects that are not even in
+                    # an Acquisition chain yet.
+                    cp._delOb(k)
```
tseaver: ISTM we should be logging something about the skipped / deleted objects, so that the user has at least some chance to figure out what happened to their not-quite-a-copy.

mauritsvanrees: Adding a `logging.warn` line should be easy enough. I can do that.
Or are you thinking about showing something to the end user?

tseaver: `logging.warn` was what I was thinking.

mauritsvanrees:
I have added a logger.warn line. Sample output:

"WARNING OFS While copying /Plone/public-folder, removed /Plone/public-folder/private-page from copy because user is not allowed to view the original."

Wait, we can do one better by passing the container along. Done. Sample output:

"WARNING OFS While copying /Plone/public-folder to /Plone/maurits-folder, removed /Plone/public-folder/private-page from copy because user is not allowed to view the original."

PR #84, opened 2016-12-08
Don't copy items the user is not allowed to view. [master]

mauritsvanrees:

Same as #83 but now for master.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 Bugs Fixed
 ++++++++++
 
+- Don't copy items the user is not allowed to view.
+  From Products.PloneHotfix20161129.  [maurits]
 
 Features Added
 ++++++++++++++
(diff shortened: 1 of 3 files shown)


tseaver:
@mauritsvanrees Was there any discussion of just raising `Unauthorized` if the object or any recursive subobject was not accessible?  The tricky dance between the "source" and the "copy" seems like it might be a bug magnet in the future.

mauritsvanrees:
No. Then we would need to change the whole copy procedure. As a normal user on a Linux machine I can also copy the `/etc` folder and will get a few errors about files that I am not allowed to see, but I still get a copy of all the other files. This more or less does the same, or at least with the same end effect.

cc @vangheem @jensens

jensens:
I also don't think just raising Unauthorized is what we want. There are those cases, at least in CMS world, where ones is happy to copy only allowed things and left all other parts behind. 

If called TTW, showing errors (or better warnings) that parts were not copied is probably a good idea. 

But we are deep into the mechanism in here, so we need something very generic like an simple event notify? This way the application can subscribe to it and do appropriate action to tell the user that parts of the tree were skipped.

mauritsvanrees:
The code now logs a warning when doing a partial copy.

PR #88, opened 2017-01-17
In `str.format`, check the security for attributes that are accessed. [2.13]

mauritsvanrees:
This is part of PloneHotfix20170117. See an [article by Armin Ronacher](http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/) for how `str.format` could possibly be abused. The article is about Jinja2, but Zope page templates can be affected too.

Biggest question here is: where do we put this code? I decided to put it in a new file `src/Zope2/App/safe_formatter.py` and to load this in `src/Zope2/App/startup.py:startup`.

I will make a pull request for master when this is accepted.

### DIFF
--- a/src/Testing/ZopeTestCase/zopedoctest/functional.py
+++ b/src/Testing/ZopeTestCase/zopedoctest/functional.py
@@ -237,8 +237,9 @@ class ZopeSuiteFactory:
 
         # If the test_class does not have a runTest method, we add
         # a dummy attribute so that TestCase construction works.
+        # runTest may get called, so set it to something that works.
         if not hasattr(test_class, 'runTest'):
-            setattr(test_class, 'runTest', None)
+            setattr(test_class, 'runTest', lambda x: None)
 
         # Create a TestCase instance which will be used to execute
         # the setUp and tearDown methods, as well as be passed into
@@ -336,4 +337,3 @@ __all__ = [
     'FunctionalDocTestSuite',
     'FunctionalDocFileSuite',
     ]
-
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -8,6 +8,8 @@ http://docs.zope.org/zope2/
 2.13.26 (unreleased)
 --------------------
 
+- In ``str.format``, check the security for attributes that are accessed.
+  Part of PloneHotfix20170117.  [maurits]
 
 
 2.13.25 (2017-01-13)
--- a/src/Zope2/App/safe_formatter.py
+++ b/src/Zope2/App/safe_formatter.py
@@ -1,0 +1,90 @@
+from AccessControl import allow_type
+from AccessControl.ZopeGuards import guarded_getattr
+from collections import Mapping
+
+import string
+
+
+class _MagicFormatMapping(Mapping):
+    """
+    Pulled from Jinja2
+
+    This class implements a dummy wrapper to fix a bug in the Python
+    standard library for string formatting.
+
+    See http://bugs.python.org/issue13598 for information about why
+    this is necessary.
+    """
+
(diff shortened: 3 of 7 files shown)


mauritsvanrees:
Some `AccessControl` tests failed. I have added an [`AccessControl` pull request](https://github.com/zopefoundation/AccessControl/pull/22) to fix that.

That does raise the question: is it maybe better to move the current `Zope2` pull request into `AccessControl`? Sounds good to me.

hannosch:
If I understand the issue correctly, than ``str.format`` has a security hole, which wasn't found when we updated RestrictedPython to support Python 2.6 and later.

To me that sounds like it's a general problem that should be solved in AccessControl/RestrictedPython, as it probably shows up in all restricted environments.

Otherwise we'd need to make sure this is patched in DocumentTemplate (DTML expressions), zope.pagetemplate, Products.Five (viewpagetemplates), Products.PageTemplates and Chameleon.

src/Testing/ZopeTestCase/zopedoctest/functional.py
```diff
@@ -237,8 +237,9 @@ def setup_test_class(self):
 
         # If the test_class does not have a runTest method, we add
         # a dummy attribute so that TestCase construction works.
+        # runTest may get called, so set it to something that works.
         if not hasattr(test_class, 'runTest'):
-            setattr(test_class, 'runTest', None)
+            setattr(test_class, 'runTest', lambda x: None)
```
hannosch: When I run into this problem, it's usually caused by not defining an explict test suite for the test module. IIRC ZopeTestCase never got support for unit test auto-discovery without test suite declarations. But I never looked into this in detail.

hannosch: What I mean is a module level test_suite function, for example https://github.com/zopefoundation/Zope/blob/master/src/Products/Five/browser/tests/test_pages.py#L69

The double under main function isn't required anymore. And if a test module has only test classes inherited from unittest.TestCase one can also leave out the test_suite function. But if there's old style tests around, I think one still needs the extra function.

mauritsvanrees: I and others have removed some `test_suite` functions in Plone without adverse effects the past few years. Whether this works correctly might depend on the test runner that is used.

I hope no one has removed this function and then unnoticedly kept some tests from being discovered. I guess the change is bigger that there *is* a `test_suite` function and that not all test available cases are added to it.

mauritsvanrees: When I use a checkout of Zope2 branch 2.13, do `bin/develop co AccessControl`, run buildout, and switch `AccessControl` to the branch of https://github.com/zopefoundation/AccessControl/pull/23 I see the same problem, which could be solved in the same way by fixing the dummy `runTest`.

It then already goes wrong when doing `bin/alltests -m OFS -m AccessControl`. Part of the problem may be that several `OFS` tests do this:

```

def test_suite():
    from Testing.ZopeTestCase import ZopeDocTestSuite
    return ZopeDocTestSuite()
```

So they call `ZopeDocTestSuite` without any arguments, to let it create a doctest for the current test module. Those tests *are* executed as far as I see, so it seems okay. But when I go in with a pdb, the test case inside this suite *does* have a proper `runTest` method.

mauritsvanrees:
I am closing this one. Instead the following pull request are needed on 2.13 (and similar on master, but I'll get to that later):
- [X] #89 dummy runTest fix in Zope
- [ ] https://github.com/zopefoundation/AccessControl/pull/22, define and use a context manager, needed so the next pull request can work
- [ ] https://github.com/zopefoundation/AccessControl/pull/23, the real fix

PR #91, opened 2017-01-31
Make the WSGI publisher support the test browser's handleErrors = False option

davisagli:
(to raise exceptions rather than rendering a response)

Also make the repr of the Unauthorized exception raised by the WSGI publisher
backwards-compatible.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,11 @@ Bugs Fixed
 - Don't copy items the user is not allowed to view.
   From Products.PloneHotfix20161129.  [maurits]
 
+- Don't render exception views when WSGI publisher is called
+  by the test browser (when 'wsgi.handleErrors' in the request
+  environ is False). This makes it behave more similarly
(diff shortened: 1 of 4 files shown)


hannosch:
This should already be supported, as the testbrowser code conditionally attaches the HTTPException middleware to the WSGI pipeline, based on whether or not `handleErrors` is True (https://github.com/zopefoundation/Zope/blob/master/src/Testing/testbrowser.py#L46).

The problem you are seeing is specific to `Unauthorized` and `Redirection`, which have special code inside the WSGIPublisher and prematurely turn these exceptions into responses. That part should only happen inside the zExceptions HTTPException middleware, rather than inside the WSGIPublisher. If that were true, the exceptions would be raised correctly and not caught.

I meant to clean this part up, and remove the special handling for Unauthorized and Redirection, but completely forgot about it.

Also: The test failures for this PR are intermittent errors and not real ones :(

davisagli:
Closing in favor of #92

PR #92, opened 2017-01-31
render Unauthorized and Redirect exceptions outside of WSGIPublisher

davisagli:
Instead, add info needed for headers to the exception instances so that zExceptions can render them when invoked by the HTTPExceptionHandler middleware (see zopefoundation/zExceptions#2)

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -19,6 +19,7 @@ parts =
     requirements
 sources-dir = develop
 auto-checkout =
+    zExceptions
 
 
 [test]
--- a/sources.cfg
+++ b/sources.cfg
@@ -14,7 +14,7 @@ five.globalrequest = git ${remotes:github}/five.globalrequest pushurl=${remotes:
 MultiMapping = git ${remotes:github}/MultiMapping pushurl=${remotes:github_push}/MultiMapping
 Persistence = git ${remotes:github}/Persistence pushurl=${remotes:github_push}/Persistence
 RestrictedPython = git ${remotes:github}/RestrictedPython pushurl=${remotes:github_push}/RestrictedPython
-zExceptions = git ${remotes:github}/zExceptions pushurl=${remotes:github_push}/zExceptions
+zExceptions = git ${remotes:github}/zExceptions pushurl=${remotes:github_push}/zExceptions branch=davisagli-exception-headers
 zope.globalrequest = git ${remotes:github}/zope.globalrequest pushurl=${remotes:github_push}/zope.globalrequest
 
 # Optional dependencies
--- a/src/Testing/ZopeTestCase/functional.py
+++ b/src/Testing/ZopeTestCase/functional.py
@@ -103,6 +103,10 @@ class Functional(sandbox.Sandboxed):
         wsgi_headers = BytesIO()
 
         def start_response(status, headers):
+            response.setStatus(status.split()[0])
+            for name, value in headers:
+                response.setHeader(name, value)
+
             wsgi_headers.write('HTTP/1.1 %s\r\n' % status)
             headers = '\r\n'.join([': '.join(x) for x in headers])
             wsgi_headers.write(headers)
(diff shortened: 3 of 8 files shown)


davisagli:
FYI, I'm debugging a related issue where WSGIPublisher's call to `response._unauthorized(exc)` fails because PluggableAuthService replaces the response's _unauthorized attribute with a method of the PAS that does not accept an exception instance.

hannosch:
Good to know, I'll hold of merging this. I always wondered how exactly PAS manages to inject itself into the publishing flow.

davisagli:
I don't really see how to move the exception rendering out to middleware while maintaining backwards-compatibility. In addition to the issue with PluggableAuthService expecting to be able to render its own responses when `_unauthorized` is called, a number of tests in Plone rely on response.redirect setting a redirect status and headers without actually raising an exception (so that the redirect is still followed when using a test browser with handleErrors = False)

In the `davisagli-exception-publishing-take-3` branch I've made backwards-compatible changes to get our tests running, without moving the exception rendering to middleware.

hannosch:
Well, backwards compatibility isn't the primary goal and not really achievable while moving to WSGI. The exception hook is gone, standard_error_message is gone, so exception handling in general has to change.

I was assuming that application code that wants to influence exception handling needs to register exception views now. They are the only hook that I know off, which is available both in Zope 2.13 / non-WSGI and Zope 4 / WSGI environments.

And to fully replace what standard_error_message was able to do, you need to provide your own version of the HTTPExceptions middleware, especially if you want to change its HTML output.

davisagli:
Yeah, some combination of making PAS use exception views and making Plone use custom exception middleware might make this possible. I was feeling stuck and trying to find a quick way to let us move on to seeing what other errors we need to deal with. But I should revisit this later.

hannosch:
I've extended my last change via 25461a701c1ad7cbb08f937dd47434844573dc41, which should bring back the old behavior.

The functional/testbrowser code now always inserts the HTTPExceptions middleware and always turns subclasses of HTTPException into responses. The `handle_errors` flag only determines what to do with all other exceptions.

So `handle_errors = True`, means always get a response, maybe even a 500. `handle_errors = False` means, if something unexpected happens like a ValueError, raise it, but for a not found return a response.

davisagli:
Hmm, I think the goal of setting `handleErrors = False` on the test browser is to be able to write tests that check for the original exception that was raised, including HTTPExceptions. But as the test you changed in makesite.txt demonstrates, that's not working yet. Since HTTPExceptions are still turned into responses, the `_checkStatus` function in the testbrowser raises a generic HTTPError so the test sees that instead. I think it would be nice (or at least more backwards compatible) to also skip making responses for HTTPExceptions when `handleErrors` is False; if not then in Plone we have to update a bunch of doctests that look for Unauthorized and NotFound rather than HTTPError. We can do that if necessary (or use our own HTTPExceptionHandler middleware, which might be necessary anyway for other reasons), but I don't understand the reason for treating HTTPExceptions differently here.

hannosch:
Damn, this is more complicated than I thought. I was trying to add more backwards compatibility based on what you said, but likely got it wrong.

hannosch:
I've reverted my commit that introduced the HTTPExceptions/handle_errors special handling again.

davisagli:
Thanks!

As it happens, plone.testing needs its own version of the test browser, so we were already handling this the way you changed back to. But there are still a few ZopeTestCase tests so its nice for it to be consistent.

FWIW, we have a branch of PluggableAuthService which hooks Unauthorized rendering using an exception view: https://github.com/zopefoundation/Products.PluggableAuthService/compare/ed676ae...plonezope4

I still need to take a closer look at what happens with redirects. For now the Plone-on-Zope-4 build is still using a branch of Zope that avoids raising an exception when response.redirect() is called.

hannosch:
Redirects are hard. We can discuss those over in #102. One option might be to distinguish between returning and raising an HTTPException sub-class. But than we'd also change the exception view handling to get called when the returned response is an exception subclass, not just when an exception was raised.

davisagli:
Yeah, I'll follow up on #102 once I've had time to digest the situation a bit more; may not be for a few days.

PR #93, opened 2017-02-01
Test that `str.format` checks security for accessed attributes

mauritsvanrees:
Part of PloneHotfix20170117.
This pull request only adds tests, no other code.

This needs https://github.com/zopefoundation/AccessControl/pull/23 which has the real fix.
This was merged, but not released yet, so we add AccessControl to auto-checkout for now.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -22,7 +22,6 @@ parts =
     wsgi
 sources-dir = develop
 auto-checkout =
-    AccessControl
 versions = versions
 
 
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -22,6 +22,7 @@ parts =
     wsgi
 sources-dir = develop
 auto-checkout =
+    AccessControl
 versions = versions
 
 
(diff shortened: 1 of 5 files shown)


hannosch:
@mauritsvanrees, @icemac what's the status of this PR for Zope 2.13. And does it need to be applied to Zope master as well?

mauritsvanrees:
This is ready and can be merged for Zope 2.13. It tests the code from AccessControl branch 2.13.
Plone uses Zope 2.13 in combination with newer AccessControl 3.0.x, so there

mauritsvanrees:
This is ready and can be merged for Zope 2.13. It tests the code from AccessControl branch 2.13.

Plone uses Zope 2.13 in combination with newer AccessControl 3.0.x, so there it would be good to have a merge of https://github.com/zopefoundation/AccessControl/pull/28.

For master, @icemac was thinking it [may be better to add code in `RestrictedPython` instead](https://github.com/zopefoundation/AccessControl/pull/23#issuecomment-276673953).

PR #95, opened 2017-02-02
normalize zope 3 exceptions

davisagli:
This converts some exceptions from zope.publisher and zope.security that may be raised by application code into the correct exceptions from zExceptions so that they have a chance of being rendered correctly by the exception middleware.

I'd be happy for a suggestion of a cleaner way to handle these.


hannosch:
Hmm. This doesn't look pretty, but I don't see any really good way either. Those exceptions all have different signatures and different required arguments for their `__init__`, so I don't really want to mix them into the zExceptions exceptions as base classes (and let zExceptions depend on zope.publisher).

Do you have an example of what kind of code raises these exceptions in Zope/Plone? The old publisher didn't do anything with them, as far as I can tell, so I wonder how raising a zope.publisher exception actually worked before or what that did.

davisagli:
The case that led me to this was this line in plone.app.uuid (which I think aims to work in both Zope 3 and Zope 2): https://github.com/plone/plone.app.uuid/blob/master/plone/app/uuid/browser.py#L29

In Zope2 2.13, the publisher handles exceptions via HTTPResponse.exception, which calls self.setStatus(t) where t is the exception class. setStatus then does this:

```
        if (isinstance(status, (type, types.ClassType))
         and issubclass(status, Exception)):
            status = status.__name__

        if isinstance(status, str):
            status = status.lower()

        if status in status_codes:
            status = status_codes[status]
        else:
            status = 500
```

so as long as the exceptions have the same name as Zope2 exceptions, they result in the same status.

Not really any less ugly, is it?

hannosch:
Ok, we still have that logic and even invoke ``response.setStatus(exc.__class__)`` inside the block to handle exception views.

Could we break the exception conversion logic out of setStatus into a function, and than call that instead of your if/elif block? That way we'd hopefully make this work for all kinds of exceptions.

It's ugly either way, but I'd rather re-use the code, especially since the exception view block uses it as well.

davisagli:
Yeah, that makes sense now that I understand how this worked before.

davisagli:
Hmm, looking closer I don't think it makes sense to break out the logic from setStatus, because it is for converting an arbitrary exception to an HTTP status code, rather than to another exception. However there is a function in zExceptions, `upgradeException`, which I think can be modified to be useful here.

davisagli:
I updated this branch with the approach based on `zExceptions.upgradeException`

PR #96, opened 2017-02-03
[WIP] Remove ApplicationManager

MatthewWilkes:
As the Control Panel has been removed in Zope4, remove the support classes and special casing in Application, so existing persistent ApplicationManagers are loaded as broken objects.

This fixes problems when loading an older database, as ApplicationManager lost its SimpleItem parent, so could no longer be populated into persistent caches, which happened before the deletion code ran.


hannosch:
The Control Panel is still there in Zope 4. In the left hand ZMI menu, there's a link to it at the very top. It is stripped down, but still shows version information and gets you a list of the databases, incl. their connections and the ability to undo transactions.

If the persistent removal is a problem, I'd rather but back persistent as a base class.

MatthewWilkes:
@hannosch Yeah, David pointed that out to me. I (like many people) have always used /manage_main to get to the ZMI, to avoid the iframe hell. I'm looking at making the remaining features views now.

Adding persistent back in would fix it, but I thought the whole point of the Zope4 effort was around removing un-needed legacy code and modernising. Would you _really_ prefer having multiple persistent objects for exposing that functionality?

hannosch:
I did work on the control panel for two reasons: 1. some of its functionality didn't make sense in WSGI land, as Zope isn't actually controlling the process anymore (no more Lifetime.loop). 2. I wanted to get rid of as many of the default persistent objects I could.

I tried to modernize more of it, but the way the database chooser works and injects fake contexts gave me headaches, so I stopped. You are welcome to go forward, this should all really be views.

MatthewWilkes:
@hannosch It looks like it shouldn't be too bad if we use the modern view pattern, rather than attaching DTML objects to a context, but it looks like nowhere else in Zope is doing that yet. Do you know if that's a deliberate decision, or just a refactor that's low priority?

hannosch:
@MatthewWilkes Nobody really worked on Zope for a long time, so converting to views just never happened. There's probably some test-only views in Products.Five. One other problem you might run into is test support. I think ZopeTestCase has a ZopeLite layer that should load ZCML. But many tests in Zope don't use it and don't load any ZCML as a result.

For larger changes this can be a problem. but for something as isolated as the control panel, it shouldn't be that much of a problem to change them and use ZopeTestCase.

hannosch:
I'm closing this PR for now. The code changes here aren't the way to move forward. I think we can either mix in Persistent, so the migration works or convert to views. But outright removing the App Manager isn't the right thing.

PR #98, opened 2017-02-03
Bring back export/import and drop xml export/import

pbauer:


mauritsvanrees:
I worked with Philip on this yesterday. I have merged this to the `plonezope4` branch that we use in Plone to gather all changes.

mauritsvanrees:
I did two minor fixes. Result is that the message you literally see in the browser after import is no longer this:

```
<em><built-in function id></em> successfully imported
```

but this:

```
"test-folder" successfully imported
```

PR #100, opened 2017-02-07
A couple fixes for WSGI publisher exception handling

davisagli:
* Avoid converting exception type if it's a subclass of the preferred HTTP exception
* Preserve the original traceback on Python 2

### DIFF
--- a/src/ZPublisher/WSGIPublisher.py
+++ b/src/ZPublisher/WSGIPublisher.py
@@ -20,6 +20,7 @@ import sys
 from AccessControl.SecurityManagement import newSecurityManager
 from AccessControl.SecurityManagement import noSecurityManager
 from six.moves._thread import allocate_lock
+from six import reraise
 import transaction
 from transaction.interfaces import TransientError
 from zExceptions import (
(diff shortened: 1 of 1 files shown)


src/ZPublisher/WSGIPublisher.py
```diff
@@ -201,7 +202,9 @@ def _publish_response(request, response, module_info, _publish=publish):
             response.setBody(view())
             return response
 
-        raise exc
+        # Reraise the exception, preserving the original traceback
+        t, v, tb = sys.exc_info()
```
tseaver: Hmm, maybe it is just old scar tissue, but I still get the heebie-jeebies seeing the traceback bound to a local variable:  I would instinctively have written that as:

        reraise(exc.__class__, exc, sys.exc_info()[-1])

davisagli: I guess you mean because of this issue: https://cosmicpercolator.com/2016/01/13/exception-leaks-in-python-2-and-3/

Thanks for catching that; will fix.

Issue #79, opened 2016-12-01
Not compatible with transaction > 2.0


icemac:
Transaction notes are strings instead of unicode which is required by `transaction 2.0.3`. Thus Zope2 cannot be used together with ZODB 5.x.

Example of a traceback see https://builds.whq.gocept.com/job/gocept.httpserverlayer/122/testReport/junit/src.gocept.httpserverlayer.plonetestingz2.tests.zope2.test_zope2/Zope2Tests/test_each_request_gets_a_separate_zodb_connection/

leorochael:
Considering the main use of the transaction note on Zope 2 is for the management screens, I think the note value should be encoded according to the management_page_charset to obtain a unicode string.

icemac:
@pbauer I think you worked on this issue on the sprint is this correct? On which branch did you work?

pbauer:
This is adressed in the plonezope4. The commits are https://github.com/zopefoundation/Zope/commit/924ae93d50a12034653f95597756c919de4a359e and https://github.com/zopefoundation/Zope/commit/e8a1b47702fb02cbcf55167d3a8ed99ca0955485

icemac:
@pbauer Could you cherry-pick these commits to a separate branch so they can be merged before merging `plonezope4`?

PR #105, opened 2017-02-20
Update to transaction >=2.0

hannosch:
This addresses #79, by cherry-picking the two commits from @pbauer, adding some tests, retaining Py3 compatibility and updating the version pins.


pbauer:
Thanks for doing this, it was on my todo-list but I was swamped.

hannosch:
@pbauer Thank the US for presidents day - having all your co-workers on holiday grants you free time ;)

Issue #102, opened 2017-02-11
Zope 4 WSGIPublisher: exception views are rendered outside of pubevents


davisagli:
Currently the WSGIPublisher first publishes the request running within the transaction_pubevents context manager, then catches any exception and tries to render an exception view. This makes for some regressions compared to how exception views used to work:
1. The PubBeforeAbort event is not called after the exception view produces a response. (This is one place where plone.transformchain hooks in, so it means that Diazo themes are not applied to exception view responses.)
2. The PubFailure event is called before the exception view is rendered, which means that the request is no longer registered as active with zope.globalrequest.

I'm thinking maybe rendering exception views actually belongs within the `publish` function.

davisagli:
Rather than adding exception view rendering into `publish`, I propose that we move the `except` block of `_publish_response` into a function called `_err_hook`, and make the following change in publish_module:

```diff
-                response = _publish_response(
-                    request, response, module_info, _publish=_publish)
+                with transaction_pubevents(request):
+                    try:
+                        response = _publish(request, module_info)
+                    except Exception as exc:
+                        tm.doom()
+                        response = _err_hook(exc, request)
```

We should then probably also make `transaction_pubevents` notify PubBeforeAbort and PubFailure rather than PubBeforeCommit and PubSuccess in the case where the transaction was doomed. I think.

hannosch:
Ok, the regression makes sense. The new WSGI publisher is based on the WSGI publisher in Zope 2.13, which didn't support exception views at all, so I added them in the wrong spot.

For doomed transactions, we can make up some new meaning. Zope 2.13 only had minimal support for them as part of its custom transaction manager (https://github.com/zopefoundation/Zope/blob/2.13/src/Zope2/App/startup.py#L299). If you tried to commit a doomed transaction it instead aborted it. But otherwise it didn't change anything and had no effect on the response.

For Zope 4 I've got rid of the custom transaction manager and just use the default one from transaction. That one should raise a DoomedTransaction exception, when trying to commit a doomed transaction. So to get the old behavior back, I've added the "silently abort doomed transaction" logic to the publisher.

[...]

PR #106, opened 2017-02-25
Adjust rendering of exception views

davisagli:
Here are the publisher-related changes which I made on the zopeplone4 branch to get the Plone tests running with Zope 4:

* Return to not raising an exception when response.redirect() is called
* Render exception views within the pubevents context manager

refs #102

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,12 @@ Bugs Fixed
 - Use unicode transaction-notes to support ZODB 5.
   [pbauer]
 
+- Don't raise an exception when `WSGIResponse.redirect` is called
+  (these redirects should be followed even when the test browser uses
+  `handleErrors = False`).
+  [davisagli]
+
+
 Features Added
 ++++++++++++++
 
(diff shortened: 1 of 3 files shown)


hannosch:
We added backwards compatibility for `response.redirect` (#114) and `response._unauthorized` (#112) again, so all the CMF and PAS testbrowser tests work again (or at least the CMF ones don't fail because of this change).

Do you have an example of a project with a failing test for the exception view part of this PR, or a minimal test to replicate the problem?

Related, #115 is still to be done, so redirects which don't abort the transaction are used again.

davisagli:
I'll need to work on a minimal test case, but the gist is that in Plone if the publisher success/failure events are not fired after exception views are rendered, they will not be themed with Diazo. Also firing them before the exception view means the exception view does not have access to the request via zope.globalrequest

hannosch:
If you can give me an example of a package with failing tests, I'm more than happy to have a look there. The exception view part I understand. Though it's a bit weird, as the exception view gets the request passed in directly. But I guess it might call utilities or helper code, which in turn relies on globalrequest.

davisagli:
If you build Plone coredev (5.1 branch) with `-c plip-zope4.cfg`, then switch Zope to master (to avoid my attempted fix from this PR) and run `bin/test -s plone.app.theming` you should see 3 failing tests due to the 404 exception view not being themed.

I'm failing to recall which test was failing due to not having the global request available, but yeah, it was not the exception view itself which needed it, but something included indirectly by the template used for the exception view.

hannosch:
This was superseded by #173.

PR #107, opened 2017-04-03
downgrade coverage version pin

thet:
use coverage 3.7.1, as coverage > 4 doesnt work with z3c.coverage.
Refs: https://github.com/zopefoundation/z3c.coverage/pull/5

IMO, the real fix isn't fixing z3c.coverage but using standard tools to run coverage reports.


hannosch:
Mmh, Zope only has a version pin for coverage and not z3c.coverage. The Zope packages I know of also don't use z3c.coverage, but use a tox/nose/nosexcover combination (e.g. Acquisition).

What project actually uses z3c.coverage? I'm tempted to say that if a project uses z3c.coverage they don't only need to add a pin for z3c.coverage, but also add a different one for coverage itself.

mauritsvanrees:
Plone coredev pins `coverage 4.2`.
@MatthewWilkes wants to change that, see https://github.com/plone/buildout.coredev/pull/312 with short discussion between me and him.

MatthewWilkes:
I don't care which changes, I just need it not to have a conflict so I can run buildouts without having to repin coverage/z3c.coverage.

MatthewWilkes:
In fact, it looks like @thet's change is the same one I proposed, but in a different place. Likely we both decided to downgrade coverage as z3c.coverage explicitly says it doesn't work, but if we don't need z3c.coverage....

thet:
+1 for closing. z3c.coverage should just not be used. z3c.form already got the dependency removed: https://github.com/zopefoundation/z3c.form/commit/4e5f921e15f89f1d887060212c588a07937d4e8e

PR #112, opened 2017-05-05
Restore a `_unauthorized` hook on the response object.

hannosch:
This is used as a hook by PluggableAuthService.

The HTTPResponse module has a new base class and two variants of the HTTP response class. You probably need to update the patch in PAS, to patch both `HTTPResponse.HTTPResponse` and `HTTPResponse.WSGIResponse`, as the `_unauthorized` method is defined on each of those classes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 Bugs Fixed
 ++++++++++
 
+- Restore a `_unauthorized` hook on the response object.
+
 Features Added
 ++++++++++++++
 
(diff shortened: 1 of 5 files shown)


davisagli:
FWIW...At the Alpine City sprint I experimented with making PluggableAuthService use an exception view for this instead of the `_unauthorized` hook, which seemed promising, but I don't think it got pushed since we were still discussing exception views.

davisagli:
Plone has a whole bunch of tests across dozens of packages which expect the Unauthorized exception to actually be raised (we often set `handleErrors=False` on the testbrowser which in Zope 2.13 makes it call `publish` with `debug=1`, which bypasses the `err_hook`). Before this PR (well, in February) I had Plone tests passing in Zope 4 using a generic exception view which bypassed itself when called from the testbrowser with handleErrors=False (as detected via the request environ; see https://github.com/plone/Products.CMFPlone/commit/65ec7177f1377565e11e7883746f5f35ba6b8bb4#diff-fb13ecaf98938e84c27d087c1a876a79R11). Now there doesn't appear to be a way to do this.

davisagli:
My suggestion for how to support this is to add:
```
        if not request.environ.get('wsgi.handleErrors', True):
            reraise(exc.__class__, exc, sys.exc_info()[2])
```
before the `if isinstance(exc, Unauthorized):` block in `_publish_response`

PR #119, opened 2017-05-10
Explicitly require Manager role for `AltDatabaseManager`.

mauritsvanrees:
If you use [`experimental.publishtraverse`](https://pypi.python.org/pypi/experimental.publishtraverse) and try to call `manage_pack`, it warns that the object (the `manage_pack` function) has no roles. In strict mode it will fail.

This is an indication that the function *might* be available for anonymous users. That is not the case here, but being strict seems good.

### DIFF
--- a/src/App/ApplicationManager.py
+++ b/src/App/ApplicationManager.py
@@ -482,6 +482,7 @@ class ApplicationManager(Folder,CacheManager):
 class AltDatabaseManager(DatabaseManager, CacheManager):
     """ Database management DBTab-style
     """
+    __roles__ = ('Manager', )
     db_name = ApplicationManager.db_name.im_func
     db_size = ApplicationManager.db_size.im_func
     manage_pack = ApplicationManager.manage_pack.im_func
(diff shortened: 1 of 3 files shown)


tseaver:
Ugh:

- The current `bootstrap.py` is not Python 2.6 compatible.
- buildout's self-updating seems to have broken with recent versions of setuptools

We may need to `pip install "zc.buildout < 2.4dev" and use its generated buildout to get past this.

mauritsvanrees:
I just saw it too. Locally, `bin/buildout -n` already helped. I have added a commit to try that.

mauritsvanrees:
Well, Travis worked with only three extra commits. :-)

Now we can concentrate on whether the real change is good or not.

tseaver:
Another option might be to call the bootstrap with `--buildout-version=2.3.1` (`bootstrap.py` doesn't know about the pins).

src/App/ApplicationManager.py
```diff
@@ -482,6 +482,7 @@ def objectIds(self, spec=None):
 class AltDatabaseManager(DatabaseManager, CacheManager):
     """ Database management DBTab-style
     """
+    __roles__ = ('Manager', )
```
tseaver: AFAICT this should be fine.

PR #123, opened 2017-05-12
Pass bytes for file content.

tseaver:


src/OFS/Image.py
```diff
@@ -520,6 +520,9 @@ def _read_data(self, file):
         n = 1 << 16
 
         if isinstance(file, str):
+            raise ValueError("Must be bytes")
```
mauritsvanrees: Is this branch for Python 3 only? I ask because this seems wrong for Python 2: `str` and `bytes` are the same there, so in practice both `str` and `bytes` will get this ValueError.
Maybe do it like this:

```
if isinstance(file, bytes):
    ... handle normally
elif isinstance(file, str):
    raise ValueError("Must be bytes")
```

Or actually, since the top of the file has `unicode = str` when on Python 3, the `str` instance check can become:

```
if isinstance(file, unicode):
   raise ValueError("Must be bytes")
```

Note: I don't use Python 3 as much as I should like, so I may be wildly off base here.

stephan-hof: When I look at other packages which share the same code for python2/3 they have something like this.
http://pythonhosted.org/six/#six.binary_type
http://pythonhosted.org/six/#six.text_type
For example flask has its own _compat.py  to avoid the dependency to six. https://github.com/pallets/flask/blob/master/flask/_compat.py

I think to port the Zope2 codebase something like this is needed as well.

hannosch: We should really use six here. When I did the first Python 3 changes on Zope2 itself, I didn't use six. Those were done when I simply ran flake8/pep8 over the codebase since none of the stack was compatible yet. Since we use six in almost all packages incl. Zope2, we should just convert my initial custom `sys.version_info` checks.

PR #128, opened 2017-05-14
More Python 3 fixes.

hannosch:
Special call out to `bytes(3)`, which produces `'3'` in Python 2 and `b'\x00\x00\x00'` in Python 3. I also like Python 3.5 and bringing back % formatting for bytes :(

This fixes another round of test failures. Logging in via basic auth is still broken, and request/response issues around headers and the body being str/bytes.

Python 3.4: 1118 tests, 45 failures, 36 errors and 1 skipped in 8.470 seconds.

### DIFF
--- a/src/OFS/tests/testRanges.py
+++ b/src/OFS/tests/testRanges.py
@@ -186,15 +186,20 @@ class TestRequestRange(unittest.TestCase):
 
         ct = rsp.getHeader('content-type').split(';')[0]
         draftprefix = draft and 'x-' or ''
-        self.assertFalse(ct != 'multipart/%sbyteranges' % draftprefix)
+        self.assertEqual(ct, 'multipart/%sbyteranges' % draftprefix)
         if rsp.getHeader('content-length'):
-            self.assertFalse(rsp.getHeader('content-length') != str(len(body)))
+            self.assertEqual(rsp.getHeader('content-length'), str(len(body)))
 
-        # Decode the multipart message
-        bodyfile = io.BytesIO('Content-Type: %s\n\n%s' % (
-            rsp.getHeader('content-type'), body))
-        partmessages = [part
-                        for part in email.message_from_file(bodyfile).walk()]
+        # Decode the multipart message and force a latin-1 encoding,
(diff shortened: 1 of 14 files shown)


src/OFS/tests/testRanges.py
```diff
@@ -186,15 +186,20 @@ def expectMultipleRanges(self, range, sets, draft=0):
...
+            self.assertEqual(rsp.getHeader('content-length'), str(len(body)))
 
-        # Decode the multipart message
-        bodyfile = io.BytesIO('Content-Type: %s\n\n%s' % (
-            rsp.getHeader('content-type'), body))
-        partmessages = [part
-                        for part in email.message_from_file(bodyfile).walk()]
+        # Decode the multipart message and force a latin-1 encoding,
+        # revert that later after the email was parsed
+        bodyfile = io.StringIO(
```
tseaver: Should this be `BytesIO`?

src/Testing/ZopeTestCase/ZopeLite.py
```diff
@@ -28,14 +28,16 @@
...
 
 from Testing.ZopeTestCase import layer
 
 # Allow code to tell it is run by the test framework
 os.environ['ZOPETESTCASE'] = '1'
 
 # Increase performance on MP hardware
-sys.setcheckinterval(2500)
+if PY2:
+    sys.setcheckinterval(2500)
```
tseaver: Ugh, this is a nasty import time side effect.

src/OFS/tests/testRanges.py
hannosch: I thought so too. But when I tried the ` email.message_from_file` call just below with binary it blew up. Apparently that wants text. That's why I ended up with the forced latin-1 decode, get_payload encode trick.

src/Testing/ZopeTestCase/ZopeLite.py
hannosch: Yeah, to be fair, this is in a test-only module and has been there forever. I've just disabled it for PY3, as that only has the switchinterval and the default value for it is generally good.

PR #130, opened 2017-05-15
More Python 3 fixes

hannosch:
All Python 3 versions now see the same number of test failures:

Total: 1118 tests, 28 failures, 33 errors and 1 skipped in 7.121 seconds.

### DIFF
--- a/src/Products/Five/browser/tests/resource.txt
+++ b/src/Products/Five/browser/tests/resource.txt
@@ -65,7 +65,7 @@ PageTemplateResource's __call__ renders the template
 
   >>> for r in dir_resource_names:
   ...     resource = self.folder.unrestrictedTraverse(base % r)
-  ...     self.assertTrue(isinstance(resource, Resource))
+  ...     self.assertIsInstance(resource, Resource)
   ...     if not isinstance(resource, PageTemplateResource):
   ...         self.assertEquals(resource(), base_url % r)
 
--- a/src/Products/Five/browser/tests/resource.txt
+++ b/src/Products/Five/browser/tests/resource.txt
@@ -65,7 +65,7 @@ PageTemplateResource's __call__ renders the template
 
   >>> for r in dir_resource_names:
   ...     resource = self.folder.unrestrictedTraverse(base % r)
-  ...     self.assert_(isinstance(resource, Resource))
+  ...     self.assertTrue(isinstance(resource, Resource))
   ...     if not isinstance(resource, PageTemplateResource):
   ...         self.assertEquals(resource(), base_url % r)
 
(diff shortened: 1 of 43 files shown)


src/Products/Five/browser/tests/resource.txt
```diff
@@ -65,7 +65,7 @@ PageTemplateResource's __call__ renders the template
 
   >>> for r in dir_resource_names:
   ...     resource = self.folder.unrestrictedTraverse(base % r)
-  ...     self.assert_(isinstance(resource, Resource))
+  ...     self.assertTrue(isinstance(resource, Resource))
```
tseaver: Could use `self.assertIsInstance(resource, Resource)`.

src/Products/PageTemplates/tests/secure.pt
```diff
@@ -1,5 +1,5 @@
 <div xmlns="http://www.w3.org/1999/xhtml"
      xmlns:tal="http://xml.zope.org/namespaces/tal">
   <span tal:define="soup view/tagsoup | options/soup"
-        tal:replace="structure python: modules['cgi'].escape(soup)" />
+        tal:replace="structure python: modules['cgi'].escape(soup, True)" />
```
tseaver: Shouldn't we be avoiding `cgi.escape`?

src/Testing/ZopeTestCase/zopedoctest/testFunctionalDocTest.py
```diff
@@ -89,23 +89,24 @@ def test___str___w_headers(self):
 
 
 SHOW_COOKIES_DTML = '''\
-<dtml-in "REQUEST.cookies.keys()">
+<dtml-in "list(REQUEST.cookies.keys())">
```
tseaver: Just iterating doesn't work under Py3k?

src/Products/PageTemplates/tests/secure.pt
hannosch: We should. But it's still there in Python 3.6, so it's not required for porting.

I couldn't remember how to write a conditional import or the equivalent thereof inside a restricted page template. IIRC someone has to declare the cgi (or html) module and the escape function as safe to use as well.

src/Testing/ZopeTestCase/zopedoctest/testFunctionalDocTest.py
hannosch: This is admittedly me papering over an issue, so I can see the other errors.

This failed with something like SafeIter has no attribute `__len__` and that's most likely an AccessControl issue.

src/Products/Five/browser/tests/resource.txt
hannosch: Ah, I didn't even know about that one.

PR #131, opened 2017-05-16
Py3 again

hannosch:
This includes a larger refactoring of the HTTP response to get a semblance of order into it. It constantly moves its body back and forth between bytes and text and needs one or the other to guess the content type, modify the body in place or do compression.

The tests currently all pass, and I introduced a new `charset` attribute, which gets set to the zpublisher default encoding and gets modified when someone calls setHeader content-type with a charset value, incl. specifying a charset in the initial header set. A new `text` get/set property allows one to read or write the body according to this charset.

I hope those additions help when fixing other issues. I did delete a workaround for special casing certain characters in latin-1, as that looked like a Netscape 2 / IE 4 type of workaround and was hard to keep without another set of text/bytes roundtripping and charset guessing.

Tests: Total: 1116 tests, 12 failures, 8 errors and 1 skipped in 9.615 seconds.

I'm holding off on a release, as we are sooo close to have all tests passing, and currently I still have to do fixes in AccessControl, DocumentTemplate or zExceptions all the time.

The largest chunk of the test failures are now related to request parsing and processInputs.

### DIFF
--- a/src/OFS/ObjectManager.py
+++ b/src/OFS/ObjectManager.py
@@ -589,17 +589,20 @@ class ObjectManager(CopyContainer,
         suffix = 'zexp'
 
         if download:
-            f = BytesIO()
-            ob._p_jar.exportFile(ob._p_oid, f)
+            with BytesIO() as f:
+                ob._p_jar.exportFile(ob._p_oid, f)
+                result = f.getvalue()
+
             if RESPONSE is not None:
                 RESPONSE.setHeader('Content-type', 'application/data')
                 RESPONSE.setHeader('Content-Disposition',
                                    'inline;filename=%s.%s' % (id, suffix))
-            return f.getvalue()
+            return result
 
         cfg = getConfiguration()
         f = os.path.join(cfg.clienthome, '%s.%s' % (id, suffix))
-        ob._p_jar.exportFile(ob._p_oid, f)
+        with open(f, 'w+b') as fd:
+            ob._p_jar.exportFile(ob._p_oid, fd)
 
         if REQUEST is not None:
             return self.manage_main(
--- a/src/Products/PageTemplates/tests/secure.pt
+++ b/src/Products/PageTemplates/tests/secure.pt
@@ -1,5 +1,6 @@
 <div xmlns="http://www.w3.org/1999/xhtml"
      xmlns:tal="http://xml.zope.org/namespaces/tal">
-  <span tal:define="soup view/tagsoup | options/soup"
-        tal:replace="structure python: modules['cgi'].escape(soup, True)" />
+  <span tal:define="soup view/tagsoup | options/soup;
+                    mod modules/html | modules/cgi"
+        tal:replace="structure python: mod.escape(soup, True)" />
 </div>
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -501,7 +501,7 @@ class HTTPBaseResponse(BaseResponse):
             self.notFoundError(body[1:-1].decode(self.charset))
         else:
             if title:
-                title = str(title)
+                title = text_type(title)
                 if not is_error:
                     self.body = body = self._html(
                         title, body.decode(self.charset)).encode(self.charset)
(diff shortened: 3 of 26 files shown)


src/OFS/ObjectManager.py
```diff
@@ -589,17 +589,20 @@ def manage_exportObject(self, id='', download=None,
...
+            with BytesIO() as f:
+                ob._p_jar.exportFile(ob._p_oid, f)
+                result = f.getvalue()
+
             if RESPONSE is not None:
                 RESPONSE.setHeader('Content-type', 'application/data')
                 RESPONSE.setHeader('Content-Disposition',
                                    'inline;filename=%s.%s' % (id, suffix))
-            return f.getvalue()
+            return result
```
tseaver: Seems like using a tempfile and returning it as the body would be better than reading it all back into RAM.  Is that feasible?

src/Products/PageTemplates/tests/secure.pt
```diff
@@ -1,5 +1,6 @@
 <div xmlns="http://www.w3.org/1999/xhtml"
      xmlns:tal="http://xml.zope.org/namespaces/tal">
-  <span tal:define="soup view/tagsoup | options/soup"
-        tal:replace="structure python: modules['cgi'].escape(soup, True)" />
+  <span tal:define="soup view/tagsoup | options/soup;
+                    mod modules/html | modules/cgi"
+        tal:replace="structure python: mod.escape(soup, True)" />
```
tseaver: Nice!

src/ZPublisher/tests/test_mapply.py
```diff
@@ -51,13 +54,15 @@ def __call__(self, b, c=4):
         v = mapply(cc.compute, (), values)
         self.assertEqual(v, '334')
 
-        class c2:
```
tseaver: Couldn't this be split into a separate testcase, using `unittest.skipUnless(PY2)`.

src/ZPublisher/tests/test_utils.py
```diff
@@ -33,5 +35,8 @@ def test_unicode(self):
         self.assertEqual(self._makeOne(u'foö'), u'foö')
 
     def test_utf_8(self):
-        self.assertEqual(self._makeOne('test\xc2\xae'), u'test\xae')
+        if PY2:
+            self.assertEqual(self._makeOne('test\xc2\xae'), u'test\xae')
+        else:
+            self.assertEqual(self._makeOne('test\xc2\xae'), u'test\xc2\xae')
```
tseaver: WAAAA!

src/ZPublisher/HTTPResponse.py
```diff
@@ -153,31 +165,42 @@ class HTTPBaseResponse(BaseResponse):
...
+            self.setBody(body)
+
+    @property
+    def text(self):
+        return self.body.decode(self.charset)
+
+    @text.setter
+    def text(self, value):
+        self.body = value.encode(self.charset)
+
```
tseaver: Nice!

src/ZPublisher/HTTPResponse.py
```diff
@@ -465,44 +492,38 @@ def setBody(self, body, title='', is_error=0,
             try:
                 body = bytes(body)
             except UnicodeError:
```
tseaver: How is this reachable?  We just handled the case that body is unicode above.

src/ZPublisher/HTTPResponse.py
```diff
@@ -900,7 +895,7 @@ def exception(self, fatal=0, info=None,
                 try:
                     b = str(b)
                 except UnicodeEncodeError:
-                    b = self._encode_unicode(unicode(b))
+                    b = self._encode_unicode(text_type(b))
```
tseaver: Is `b` supposed to be text or bytes?

src/ZPublisher/HTTPResponse.py
hannosch: This takes care of other classes, like a BytesIO instance.

src/ZPublisher/HTTPResponse.py
hannosch: It's supposed to be text, fixing in a new commit.

src/ZPublisher/tests/test_utils.py
hannosch: I'm not sure how else to write this test. The u'' -> u'' test is right above this and the b'' -> u'' test below. This wants to test the conversion of a prefix-less string, which means something different in Py 2/3.

src/OFS/ObjectManager.py
hannosch: Maybe, I only tried for a couple minutes and couldn't get it to work. Either the file was already closed by the time it reached the publisher or it wasn't being closed at all.

src/ZPublisher/tests/test_utils.py
tseaver: Maybe we should rename the test case something like `test_utf8_native_str`?

src/ZPublisher/HTTPResponse.py
tseaver: If it is `BytesIO`, then it won't raise `UnicodeError` (I was asking "is this exception reachable?").

PR #133, opened 2017-05-17
Py3 once more

hannosch:
Some of these fixes feel icky to me. They are a result of treating the field2string (str) type converter / property to always mean native str. I think that makes sense, but results in those version dependent test expectations.

Tests: Total: 1120 tests, 2 failures, 6 errors and 2 skipped in 9.273 seconds.

P.S. Special thanks goes to that metaconfigure.py ismethod fix. Took me ages to find it :(

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1206,9 +1206,8 @@ class HTTPRequest(BaseRequest):
             object = req.traverse(path)
         except Exception as exc:
             rsp.exception()
-            if object is None:
-                req.clear()
-                raise exc.__class__(rsp.errmsg)
+            req.clear()
+            raise exc.__class__(rsp.errmsg)
 
         # The traversal machinery may return a "default object"
         # like an index_html document. This is not appropriate
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -23,7 +23,6 @@ from os import unlink
 from os.path import isfile
 import random
 import re
-import sys
 from tempfile import (
     mkstemp,
     _TemporaryFileWrapper,
@@ -34,6 +33,7 @@ from AccessControl.tainted import TaintedString
 import pkg_resources
 from six import binary_type
 from six import PY3
+from six import string_types
 from six import text_type
 from six.moves.urllib.parse import unquote
 from zope.i18n.interfaces import IUserPreferredLanguages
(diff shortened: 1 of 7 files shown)


src/ZPublisher/HTTPRequest.py
```diff
@@ -1204,11 +1204,11 @@ def resolve_url(self, url):
...
             object = req.traverse(path)
-        except Exception:
+        except Exception as exc:
             rsp.exception()
-        if object is None:
-            req.clear()
-            raise sys.exc_info()[0](rsp.errmsg)
+            if object is None:
+                req.clear()
+                raise exc.__class__(rsp.errmsg)
```
tseaver: This bit shouldn't be under the `except Exception as exc` block, as `object` won't be bound if `req.traverse(path)` raises.

src/ZTUtils/Zope.py
```diff
@@ -297,7 +303,11 @@ def complex_marshal(pairs):
 def simple_marshal(v):
     if isinstance(v, str):
         return ''
+    if isinstance(v, binary_type):
+        # Py 3 only
+        return ':bytes'
```
tseaver: Does that need an explicit `if PY3` check?  Under Python 2, that will be *any* `str` instance.

src/ZPublisher/HTTPRequest.py
hannosch: Object is set to None before the preceding comments

tseaver: If the `req.clear()` and `raise` are supposed to happen inside the `except` block, then the `if object is None:` test is bogus.

src/ZTUtils/Zope.py
hannosch: This relies on the preceding check to return for Python 2 str before this is reached. I can change this to check for PY3 explicitly.

src/ZPublisher/HTTPRequest.py
hannosch: That's true. That check was outside the except block, but assumes an exception was raised. In Python 3 sys.exc_info was (None, None, None) there, so I moved the check inside the except block and use the exc directly. The object is None check can go than.

src/ZTUtils/Zope.py
tseaver: Nope, never mind.  My eye skipped over that.  This is still a bit confusing.

src/ZPublisher/HTTPRequest.py
tseaver: Sounds good.

hannosch:
Merged manually, squashing the last three commits into one.

PR #135, opened 2017-05-22
Keep buildout artifacts in tox env dir.

tseaver:
@hannosch This is the solution I was actually hoping to find in Halle:  I added the `buildout:develop={toxinidir}` line which keeps all the buildouts in the tox environment directories.


tseaver:
I'm going to merge this to make it simpler to push on with testing at PyCon sprints.

hannosch:
A late +1, this does look better.

PR #140, opened 2017-05-28
Add universal wheel marker.

hannosch:
I'd like to start releasing wheels, at least for pure Python projects with universal code (Python 2 / 3 compatibility in one code base).

While buildout can only pick these up via an extension, I think it would be nice to start releasing them, so either those extension users or pip users can get their benefits.

Objections, thoughts?

### DIFF
--- a/docs/maintenance.rst
+++ b/docs/maintenance.rst
@@ -52,7 +52,7 @@ Steps for creating a new Zope release
 
 - Upload the tagged release to PyPI::
 
-    python2.7 setup.py egg_info -RDb '' sdist bdist_wheel register upload
+    python2.7 setup.py egg_info -RDb '' sdist bdist_wheel upload
 
 - Update version information:
 
--- a/docs/maintenance.rst
+++ b/docs/maintenance.rst
@@ -52,7 +52,7 @@ Steps for creating a new Zope release
 
 - Upload the tagged release to PyPI::
 
-    python2.7 setup.py egg_info -RDb '' sdist register upload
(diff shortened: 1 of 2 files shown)


sallner:
In case the release should be done with `zest.releaser` the following entry would be also required.

```
[zest.releaser]
create-wheel = yes
```

docs/maintenance.rst
```diff
@@ -52,7 +52,7 @@ Steps for creating a new Zope release
 
 - Upload the tagged release to PyPI::
 
-    python2.7 setup.py egg_info -RDb '' sdist register upload
+    python2.7 setup.py egg_info -RDb '' sdist bdist_wheel register upload
```
dataflake: FYI, the ``register`` keyword is obsolete. You can do a straight ``upload`` without ``register``.

dataflake:
I can't see any disadvantages.

While on the topic of release files, should we at least encourage the practice of publishing detached GPG signature files for all release packages?

hannosch:
> While on the topic of release files, should we at least encourage the practice of publishing detached GPG signature files for all release packages?

I'd strongly encourage it. It's not that hard to set up, and you just need the extra `--sign` option on the command line. I think I've signed everything for a number of years now.

Speaking of signing, I'd also encourage everyone to sign all git commits and tags and tell github about your key. There's excellent docs for this over at https://help.github.com/articles/signing-commits-with-gpg/

You get these nice "verified" labels next to commits and tags as a result.

PR #141, opened 2017-05-28
Py3 one more time

hannosch:
This deals with one test failure for processInputs.

The test failure for the file upload case changed from "it's not an iterator" to "operation on closed file".

I've been trying to figure out where that underlying BytesIO instance is closed, but couldn't figure it out so far. Commenting out the close call in `cgi.FieldStorage.__del__` seems to help, but I'm not sure why that is called early in the first place.

### DIFF
--- a/src/ZPublisher/Converters.py
+++ b/src/ZPublisher/Converters.py
@@ -34,6 +34,8 @@ def field2string(v):
         return v.read()
     elif six.PY2 and isinstance(v, text_type):
         return v.encode(default_encoding)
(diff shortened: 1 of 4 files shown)


icemac:
I think the reason for calling `cgi.FieldStorage.__del__` is here:

https://github.com/zopefoundation/Zope/pull/141/files#diff-84fef2977b43b4f582ebcb38cd0ad7a1R1686

`aFieldStorage` is no longer referenced afterwards so it gets garbage collected.

PR #142, opened 2017-05-29
Update dependencies, start claiming Py3 compat in setup.py.

hannosch:
Most of the new dependencies in versions.cfg are pulled in because of the new Sphinx version.

I've also synced the ztktests list with the ZTK versions block, so there's a test runner set up to run all of those ZTK tests. They aren't run by default, as there are test isolation problems and they've never passed all at once.


requirements-full.txt
```diff
@@ -16,19 +16,19 @@ Products.MailHost==3.0
...
 ZEO==5.1
-ZODB==5.2.3
+ZODB==5.2.4
 ZServer==4.0a2
 five.globalrequest==1.0
 five.localsitemanager==3.0.1
 funcsigs==1.0.2
 mock==2.0.0
-pbr==3.0.0
+pbr==3.0.1
```
tseaver: I think `pbr` may be a fossil from `subunit`:  at least, nothing in my egg cache requires it now.

versions-prod.cfg
```diff
@@ -17,7 +17,7 @@ mock = 2.0.0
 Missing = 4.0
 MultiMapping = 4.0
 PasteDeploy = 1.5.2
-pbr = 3.0.0
+pbr = 3.0.1
```
tseaver: Ditto here.

requirements-full.txt
hannosch: I tried removing it, but it's also used by `mock`, which in turn is a test dependency of `transaction`. Not sure why that wasn't an issue in the ZTK buildout

PR #143, opened 2017-05-31
Fix final Py3 test failure.

hannosch:
This is a combination of three fixes. The tests had to be adjusted, as the cgi module in Python 3 won't work without a content-length header anymore. This is a side-effect of python/cpython@5c23b8e6ea7cdb0002842d16dbce4b4d716dd35a. Without the content length, the limit / read_bytes calculation is off, and we end up with three fieldstorage instances each with an empty byte, rather than one instance with a content of `test\n`.

The second part is a heavy handed workaround for closing the BytesIO instance underlying the FieldStorage too early. In Python 2.7, there wasn't a `__del__` defined yet. There's a good chance, a proper fix could also be related to http://bugs.python.org/issue23700 and http://bugs.python.org/issue18879, but I couldn't work out if something similar was indeed the issue here.

Finally the TemporaryFileWrapper no longer needs special closing/del logic under Python 3, as the implementation changed. There is no `close_called` attribute anymore, but rather a more complex `_closer` indirection.

I'm not particularly happy with the `__del__` part of this fix, but I couldn't come up with a better way. Any better ideas are more than welcome. I'm at my wits' end :(

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1644,18 +1644,19 @@ class TemporaryFileWrapper(_TemporaryFileWrapper):
     and doesn't complain about such a move either.
     """
 
-    unlink = staticmethod(unlink)
-    isfile = staticmethod(isfile)
+    if PY2:
+        unlink = staticmethod(unlink)
+        isfile = staticmethod(isfile)
 
-    def close(self):
-        if not self.close_called:
-            self.close_called = True
-            self.file.close()
+        def close(self):
+            if not self.close_called:
+                self.close_called = True
+                self.file.close()
 
-    def __del__(self):
-        self.close()
-        if self.isfile(self.name):
-            self.unlink(self.name)
+        def __del__(self):
+            self.close()
+            if self.isfile(self.name):
+                self.unlink(self.name)
 
 
 class ZopeFieldStorage(FieldStorage):
@@ -1663,6 +1664,13 @@ class ZopeFieldStorage(FieldStorage):
     def make_file(self, binary=None):
         handle, name = mkstemp()
         return TemporaryFileWrapper(os.fdopen(handle, 'w+b'), name)
+
+    def __del__(self):
+        # Only call close on file object, cStringIO/BytesIO objects
+        # would be closed too early.
+        if (self.file is not None and
+                isinstance(self.file, TemporaryFileWrapper)):
+            self.file.close()
 
 
 # Original version: zope.publisher.browser.FileUpload
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1668,8 +1668,7 @@ class ZopeFieldStorage(FieldStorage):
     def __del__(self):
         # Only call close on file object, cStringIO/BytesIO objects
         # would be closed too early.
-        if (self.file is not None and
-                isinstance(self.file, TemporaryFileWrapper)):
+        if isinstance(self.file, TemporaryFileWrapper):
             self.file.close()
 
 
--- a/src/ZPublisher/tests/testHTTPRequest.py
+++ b/src/ZPublisher/tests/testHTTPRequest.py
@@ -749,7 +749,9 @@ class HTTPRequestTests(unittest.TestCase, HTTPRequestFactoryMixin):
         s = BytesIO(TEST_FILE_DATA)
         start_count = sys.getrefcount(s)
 
-        req = self._makeOne(stdin=s, environ=TEST_ENVIRON.copy())
+        environ = TEST_ENVIRON.copy()
+        environ['CONTENT_LENGTH'] = str(len(TEST_FILE_DATA))
+        req = self._makeOne(stdin=s, environ=environ)
         req.processInputs()
         self.assertNotEqual(start_count, sys.getrefcount(s))  # Precondition
         req.close()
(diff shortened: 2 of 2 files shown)


src/ZPublisher/HTTPRequest.py
```diff
@@ -1644,18 +1644,19 @@ class TemporaryFileWrapper(_TemporaryFileWrapper):
     and doesn't complain about such a move either.
     """
 
-    unlink = staticmethod(unlink)
-    isfile = staticmethod(isfile)
+    if PY2:
+        unlink = staticmethod(unlink)
+        isfile = staticmethod(isfile)
```
tseaver: Why are these made `staticmethods`?  They could just be called directly (even as `os.unlink` and `os.path.isfile`).

src/ZPublisher/HTTPRequest.py
```diff
@@ -1664,6 +1665,13 @@ def make_file(self, binary=None):
         handle, name = mkstemp()
         return TemporaryFileWrapper(os.fdopen(handle, 'w+b'), name)
 
+    def __del__(self):
+        # Only call close on file object, cStringIO/BytesIO objects
+        # would be closed too early.
+        if (self.file is not None and
+                isinstance(self.file, TemporaryFileWrapper)):
```
tseaver: This could just be `if isinstance(self.file, TemporaryFileWrapper):`.

src/ZPublisher/HTTPRequest.py
```diff
@@ -1644,18 +1644,19 @@ class TemporaryFileWrapper(_TemporaryFileWrapper):
...
+                self.file.close()
 
-    def __del__(self):
-        self.close()
-        if self.isfile(self.name):
-            self.unlink(self.name)
+        def __del__(self):
+            self.close()
+            if self.isfile(self.name):
+                self.unlink(self.name)
```
tseaver: Seems like the `__del__` could be the only bit inside the `if PY2:` test.

src/ZPublisher/tests/testHTTPRequest.py
```diff
@@ -749,7 +749,9 @@ def test_close_removes_stdin_references(self):
         s = BytesIO(TEST_FILE_DATA)
         start_count = sys.getrefcount(s)
 
-        req = self._makeOne(stdin=s, environ=TEST_ENVIRON.copy())
+        environ = TEST_ENVIRON.copy()
+        environ['CONTENT_LENGTH'] = str(len(TEST_FILE_DATA))
```
tseaver: Could we just add the `CONTENT_LENGTH` unconditionally to `TEST_ENVIRON`?

src/ZPublisher/tests/testHTTPRequest.py
```diff
@@ -1144,7 +1150,7 @@ def test_no_traversal_of_view_request_attribute(self):
 
 TEST_FILE_DATA = b'''
 --12345
-Content-Disposition: form-data; name="file"; filename="file"
+Content-Disposition: form-data; name="smallfile"; filename="smallfile"
```
tseaver: :+1:

src/ZPublisher/HTTPRequest.py
hannosch: I think they were made into staticmethods, to avoid becoming bound methods and magically getting `self` as an additional call argument. Since these are simple functions, that'd fail.

src/ZPublisher/HTTPRequest.py
hannosch: I've simply indented the entire class body into the PY2 block. The `close` method looks for `self.close_called`, which raises an AttributeError in Python 3. And the fix in Python 3 was about avoiding a `__del__` in the first place, and rather using a new `_TemporaryFileCloser`. So I think both `close/__del__` should be in the block. The unlink/isfile methods are only used by the `__del__`, so we don't need them either in Py3.

src/ZPublisher/HTTPRequest.py
hannosch: Oh, yes. I've changed that a couple times ... fixing.

src/ZPublisher/tests/testHTTPRequest.py
hannosch: We could, but what would the value be? The tests all use a different body, so the length value needs to be different.

src/ZPublisher/HTTPRequest.py
hannosch: Oh. And why are these methods on the class in the first place ... good question. No idea :)

tseaver: But why do they need to be defined on the class at all?  The `__del__` method could just call them:

```python
        def __del__(self):
            self.close()
            if os.path.isfile(self.name):
                os.unlink(self.name)
```

At first I thought it was so that they could be mocked out more easily, but that doesn't appear to be the case.

hannosch: And the answer likely is found in Python 2.7's tempfile module. From the `_TemporaryFileWrapper` class:
```
# Cache the unlinker so we don't get spurious errors at
# shutdown when the module-level "os" is None'd out.  Note
# that this must be referenced as self.unlink, because the
# name TemporaryFileWrapper may also get None'd out before
# __del__ is called.
unlink = _os.unlink
```

src/ZPublisher/tests/testHTTPRequest.py
tseaver: We could add a helper which constructed the environment dict and stream from a string, and populate `CONTENT_LENGTH` there.

src/ZPublisher/HTTPRequest.py
tseaver: Bloody hell. :(

src/ZPublisher/tests/testHTTPRequest.py
hannosch: Added such a helper in the last commit.

tseaver:
LGTM pending CI.

hannosch:
Travis is acting up. The Py34 job isn't running. Cancelled and restarted it once without any change. Since the PR job ran the exact same tests and passed, I'll ignore the branch job.

gotcha:
Great ! Thanks to you @hannosch and all that made it possible !

PR #144, opened 2017-06-05
Auto-generate the release index on Travis when a tag is made.

tseaver:
This is based on another project, where we (used to) re-build docs on Travis on a `gh-pages` branch after every tag (as well as pushes to`master`).

### DIFF
--- a/update_index.sh
+++ b/update_index.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+# Update tag-based release indexes on the 'gh-pages' branch.
+
+set -ev
+
+# Override locally via e.g., 'GH_HOST=git@github.com: ./update_index.sh'
+GH_HOST="${GH_HOST:-https://${GH_OAUTH_TOKEN}@github.com/}"
+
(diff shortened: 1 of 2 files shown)


icemac:
I like it! ❤️

update_index.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+# Adding GitHub pages branch. `git submodule add` checks it
+# out at HEAD.
+GH_PAGES_DIR="ghpages"
+git submodule add -q -b gh-pages ${GH_REPO} ${GH_PAGES_DIR}
+
+# Update gh-pages with the generated release docs.
+cd ${GH_PAGES_DIR}
+./build_index.sh
```
hannosch: Where is this script? Did you maybe forget to git add it ;)

hannosch: Ah ... I should not comment before drinking my coffee. The script is of course on the gh-pages branch.

tseaver:
The proof of the pudding will be when the next tagged release lands.

Issue #103, opened 2017-02-20
Possibly clean up connection open/close logic


hannosch:
Right now Zope opens a ZODB connection while traversing over the application root object (https://github.com/zopefoundation/Zope/blob/master/src/App/ZApplication.py) via a ZApplicationWrapper. The ZApplicationWrapper is set up during app startup in https://github.com/zopefoundation/Zope/blob/master/src/Zope2/App/startup.py#L126

This wrapper opens the connection and than adds a `Cleanup` instance to the request instance `_held` attribute. This cleanup instance has a handle on the ZODB connection (`_p_jar`). While closing the request, the request's `clear` method is called (https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/BaseRequest.py#L219), which sets `_held` to `None`. This in turn causes the `ZApplication.Cleanup.__del__` to be called, which in turn aborts any still open transaction and closes the connection (`_p_jar.close()`).

I think that at least the connection open/close logic should be moved into the publisher as a context manager or try/finally block. The current logic is awkward to say the least and stems from a time when neither try/finally nor context managers where available.

sallner:
There was a regression in case an error is raised. An attempt to fix this can be found in #168.

hannosch:
Closed for good via #176.

PR #155, opened 2017-07-28
Use a context manager to explicitly open/close the ZODB connection.

stephan-hof:
Hi,

this could be a fix for #103 to explicitly open/close the connection to ZODB/Zeo.

In theory the `__getattr__` and `__bobo_traverse__` methods could be removed from
the `ZApplicationWrapper`, because `WSGIPublisher` does not use them any more.

However there is the (deprecated ?) ZServer, which still relies on them. So I did not dare to remove these two methods.

The more tricky bit was to adjust the unittests to work with the changed behaviour of the publisher. I found no way around the monkey patch in `sandboxy.py`

### DIFF
--- a/src/Testing/ZopeTestCase/sandbox.py
+++ b/src/Testing/ZopeTestCase/sandbox.py
@@ -13,9 +13,10 @@
 """Support for ZODB sandboxes in ZTC
 """
 
+import contextlib
 import transaction
 
-from App.ZApplication import ZApplicationWrapper
+import ZPublisher.WSGIPublisher
 from Testing.makerequest import makerequest
 from Testing.ZopeTestCase import connections
 from Testing.ZopeTestCase import ZopeLite as Zope2
@@ -62,15 +63,17 @@ class AppZapper:
         return self._app
 
 
-def __bobo_traverse__(self, REQUEST=None, name=None):
-    '''Makes ZPublisher.publish() use the current app object.'''
+@contextlib.contextmanager
+def load_app(module_info):
+    """Let the Publisher use the current app object."""
     app = AppZapper().app()
     if app is not None:
-        return app
-    return self.__old_bobo_traverse__(REQUEST, name)
+        yield app, module_info[1], module_info[2]
+    else:
+        with ZPublisher.WSGIPublisher.__old_load_app__(module_info) as ret:
+            yield ret
 
 
-if not hasattr(ZApplicationWrapper, '__old_bobo_traverse__'):
-    ZApplicationWrapper.__old_bobo_traverse__ = (
-        ZApplicationWrapper.__bobo_traverse__)
-    ZApplicationWrapper.__bobo_traverse__ = __bobo_traverse__
+if not hasattr(ZPublisher.WSGIPublisher, '__old_load_app__'):
+    ZPublisher.WSGIPublisher.__old_load_app__ = ZPublisher.WSGIPublisher.load_app
+    ZPublisher.WSGIPublisher.load_app = load_app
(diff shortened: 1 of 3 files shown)


src/ZPublisher/tests/test_pubevents.py
```diff
@@ -159,8 +159,13 @@ def _succeed():
...
 
 
+# Poor man's mock.
 class _Application(object):
-    pass
+    def __getattr__(self, name):
+        return self
+
+    def __call__(self, *args, **kwargs):
+        return self
```
tseaver: What tests depend on mocking these two methods?

src/Testing/ZopeTestCase/sandbox.py
```diff
@@ -62,15 +63,17 @@ def app(self):
...
+            yield ret
 
 
-if not hasattr(ZApplicationWrapper, '__old_bobo_traverse__'):
-    ZApplicationWrapper.__old_bobo_traverse__ = (
-        ZApplicationWrapper.__bobo_traverse__)
-    ZApplicationWrapper.__bobo_traverse__ = __bobo_traverse__
+if not hasattr(ZPublisher.WSGIPublisher, '__old_load_app__'):
+    ZPublisher.WSGIPublisher.__old_load_app__ = ZPublisher.WSGIPublisher.load_app
+    ZPublisher.WSGIPublisher.load_app = load_app
```
tseaver: Very weird to be monkey-patching a function we are just now adding.  Couldn't we arrange for the "real" `load_app` to do what we need instead?  E.g., have `app_wrapper()` be the `AppZapper().app()` instance?

stephan-hof: > Very weird to be monkey-patching a function we are just now adding.

I fully understand your concern. The problem I faced was that `app_wrapper` refers to the global variable `Zope2.bobo_application` retrieved from `ZPublisher.get_module_info()`. So to keep `load_app` untouched one of these have to be monkey patched instead. Lets say I do a monkey patch of `get_module_info`. Then `load_app` closes the database connection too soon and unittests fail. Which means some code needs to be added to mock/hide the call to `app._p_jar.close()` in `load_app()`.

I also tried to patch `Zope2.bobo_application` in `sandbox.Sandboxed._app` and then restore it `sandbox.Sandboxed._close`. With the goal to limit the duration the monkey patch is active.
However some unittests reinitialize `Zope2.bobo_application` after `sandbox.Sandboxed._app` is called and failing again. For example `Products.Five.browser.tests.test_defaultview.test_default_view`.

So my reasoning was: monkey patching `load_app` looks like the lesser evil.

src/ZPublisher/tests/test_pubevents.py
stephan-hof: All the tests in `test_pubevents` using the publisher, which is:
```
testSuccess
testFailureReturn
testFailureException
testFailureConflict
```
These tests call `publish_module`. Now with my change, to explicitly load the application, the app returned via `get_module_info()` must be a wrapper. Thus supporting the `__call__` method. The return value of this `__call__` must have a `_p_jar` member which must have a `close()` method, therefore I added the `__getattr__` method.

All of these tests verify that the events inside the publisher are emitted, but do not care what the application really does. They use a custom request object with its own implementation of `traverse` to achieve that.

icemac:
What has to be done for this PR to get it merged?

icemac:
@stephan-hof, @hannosch  After the merge of this PR I see an exception telling `Cannot close a connection joined to a transaction` in a package tested on top of the Zope master branch. Does this ring a bell or should I dig deeper into the issue first?

PR #156, opened 2017-08-11
Fix handling of multipart requests in functional doctests.

sallner:
[Some tests](https://github.com/zopefoundation/five.formlib/blob/master/src/five/formlib/tests/forms.txt) in `five.formlib` pass a multipart message to the http helper function. When using the default parser in the `email` module, the payload of multipart messages are a list of messages instead of a string for non_multipart messages.   

This PR fixes that and ensures that the encoding is not failing with utf-8 encoded in the request body.


sallner:
Thanks for the review!

PR #168, opened 2017-09-13
Abort the transaction before closing the connection.

sallner:
Before the rework to use a context manager the transaction was aborted directly
on the transaction manager. We still need to do this as an Unauthorized would
not be raised but lead to a `Cannot close a connection joined to a
transaction`.

This fixes a regression introduced by #155.

### DIFF
--- a/src/ZPublisher/WSGIPublisher.py
+++ b/src/ZPublisher/WSGIPublisher.py
@@ -222,6 +222,7 @@ def load_app(module_info):
     try:
         yield (app, realm, debug_mode)
     finally:
+        transaction.abort()
         app._p_jar.close()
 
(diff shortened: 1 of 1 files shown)


stephan-hof:
I'm a bit puzzled why this `Cannot close a connection joined to a transaction` happens.
I was assuming that `transaction_pubevents` deals already with aborting, committing  the transaction.

If there are good reasons to still do it in `load_app`, we should do something like in ZApplication. https://github.com/zopefoundation/Zope/blob/05ec0d6617aeba0f9429aa874e1931313276d3d3/src/App/ZApplication.py#L80
Only abort the transaction if its there.

PR #172, opened 2017-09-14
Use func_code if __code__ is None.

MrTango:
See also https://github.com/zopefoundation/zope.publisher/pull/10

### DIFF
--- a/src/ZPublisher/mapply.py
+++ b/src/ZPublisher/mapply.py
@@ -52,6 +52,10 @@ def mapply(object, positional=(), keyword={},
             raise
         code = f.__code__
         defaults = f.__defaults__
+        if code is None:
+            # RestrictedPython / templates
+            code = f.func_code
+            defaults = f.func_defaults
         names = code.co_varnames[count:code.co_argcount]
 
     nargs = len(names)


hannosch:
Based on the discussion in zope.publisher it seems this was a workaround, which is no longer needed if the old code using `func_code` / `im_func` is updated to use `__code__` / `__func__`.

As far as I know everything in Zope 4.0 (incl. PythonScripts) has been changed, so this workaround shouldn't be needed anymore.

icemac:
@hannosch This PR seems to be another way to fix the same issue I am trying to fix in zopefoundation/Products.PythonScripts#12. It can only happen for PythonScript instances which where created before Zope 4.

@MrTango Does this PR only target Products.PythonScripts or are there  other cases where `func_code` might still exists persistently?

MrTango:
@icemac @hannosch better ask the original authors of the change in https://github.com/zopefoundation/zope.publisher/pull/10

hannosch:
The zope.publisher bug mentions this as workarounds for changes required by Products.CMFCore, which are supposedly fixed on a plone-zope4 branch with zopefoundation/Products.CMFCore@0c9e66ddf31c90c3f0e60bf9c6a174c2462272c4.

But the same changes have independently been done on master as well (zopefoundation/Products.CMFCore@bce846dc9db71eb34f079ceaf80aa7dd75c836a4).

At least for newly created, non-persistent objects, I'd rather like to avoid falling back to any `func_` attributes. Inline migration code for persistent objects is something else. But that should be done in the persistent objects in question, not generally in the publisher code.

icemac:
As this does not seem to be needed any longer I am closing it now.

Issue #166, opened 2017-09-11
Support Unicode URIs


hannosch:
We decided to make all object ids, contributing to the physical path Unicode. Which also means they can show up in URIs as path segments.

We should make sure that you can use full Unicode URIs including internationalized domain names.

In Zope 2.13 object ids and URIs where limited to ascii-only and there were multiple places that checked for this explicitly.

sallner:
Resolved by #181.

PR #181, opened 2017-09-15
Handle unicode object ids.

sallner:
This solves #166.


sallner:
@tseaver There was this test.

https://github.com/zopefoundation/Zope/blob/54c6ea651ca4070ea69b8b7471d8ab32bc034aec/src/OFS/tests/testObjectManager.py#L553-L557

Do you remember why this is bad in the object id? This should be quoted by urllib anyway, shouldn't it?

sallner:
Actually, when using the angle brackets in an object id, it currently fails to do a rename with this patch. I have to look into a that again and produce a failing test first. But it would be still nice to now the reason behind the explicit declaration as illegal. In that case, I would not need to find the reason for the failure.

sallner:
@icemac I added the requested changes and dropped the attempt to allow `<`, `>` and `&`, as they caused an error during a `hasattr` call. The `id` was present as `TaintedString` in this case.

src/OFS/tests/testObjectManager.py
```diff
@@ -528,64 +528,60 @@ def test_empty_string(self):
...
-        # Unicode can only be handled under Python 3, Python 2 needs
-        # bytestrings.
-        if PY2:
-            e = self.assertBadRequest(u'abc☃')
-            self.assertEqual(
-                str(e), "('Empty or invalid id specified', u'abc\\u2603')")
-        else:
-            # Does not raise
-            self._callFUT(self._makeContainer(), u'abc☃')
+    @unittest.skipIf(PY3, 'Python 3 cannot handle bytestrings here.')
```
ale-rt: A minor thing I also have been warned about... it is better to use "not PY2" which is more future proof, and we know we have a long future :)

icemac: +1 maybe `@unittest.skipUnless(PY2 ...`could be used alternatively to prevent to use the `not`.

sallner: I changed this to `skipUnless`. Thanks.

icemac:
@sallner Could you please add a change log entry on master for this PR containing a link to the PR?

sallner:
@icemac Done.

PR #188, opened 2017-09-18
Fix install_product for products with a non-empty `misc_` attribute.

### LINKED ISSUE #187: change to new style classes breaks installing products
The change to new style classes in https://github.com/zopefoundation/Zope/commit/bb7837c667454ba1cf13c6750c6700cb31944df0 by @hannosch breaks installing products (among other things I guess):

```
Running plone.app.testing.layers.PloneFixture tests:
  Tear down Testing.ZopeTestCase.layer.ZopeLite in 0.000 seconds.
  Set up plone.testing.zca.LayerCleanup in 0.000 seconds.
  Set up plone.testing.z2.Startup in 0.058 seconds.
  Set up plone.app.testing.layers.PloneFixture Traceback (most recent call last):
  File "/home/jenkins/.buildout/eggs/zope.testrunner-4.7.0-py2.7.egg/zope/testrunner/runner.py", line 400, in run_layer
    setup_layer(options, layer, setup_layers)
  File "/home/jenkins/.buildout/eggs/zope.testrunner-4.7.0-py2.7.egg/zope/testrunner/runner.py", line 713, in setup_layer
    layer.setUp()
  File "/home/jenkins/workspace/plip-zope4/src/plone.app.testing/plone/app/testing/layers.py", line 107, in setUp
    self.setUpProducts(app)
  File "/home/jenkins/workspace/plip-zope4/src/plone.app.testing/plone/app/testing/layers.py", line 186, in setUpProducts
    z2.installProduct(app, p)
  File "/home/jenkins/workspace/plip-zope4/src/plone.testing/src/plone/testing/z2.py", line 68, in installProduct
    raise_exc=1)
  File "/home/jenkins/workspace/plip-zope4/src/Zope/src/OFS/Application.py", line 396, in install_product
    Application.misc_.__dict__[product_name] = misc_
   - __traceback_info__: ExternalEditor
TypeError: 'dictproxy' object does not support item assignment
```

See http://jenkins.plone.org/view/PLIPs/job/plip-zope4/

hannosch:
This should fix #187.

It only broke for Products, whose `__init__` module contained a non-empty `misc_` variable. I'm not sure anymore what was once stored in that module global variable, but none of the core Products contain one anymore.

`Application.misc_` is a class (and not an instance), so assigning to the class dict fails now. Only the `setattr` change of this commit is strictly required, the test works either way (the dict/mappingproxy support `__contains__`).

### DIFF
--- a/src/OFS/Application.py
+++ b/src/OFS/Application.py
@@ -393,7 +393,7 @@ def install_product(app, product_dir, product_name, meta_types,
     if misc_:
         if isinstance(misc_, dict):
             misc_ = Misc_(product_name, misc_)
-        Application.misc_.__dict__[product_name] = misc_
+        setattr(Application.misc_, product_name, misc_)
 
     productObject = FactoryDispatcher.Product(product_name)
     context = ProductContext(productObject, None, product)
--- a/src/OFS/tests/testAppInitializer.py
+++ b/src/OFS/tests/testAppInitializer.py
@@ -106,4 +106,4 @@ class TestInitialization(unittest.TestCase):
         self.configure(good_cfg)
         i = self.getOne()
         i.install_products()
-        self.assertTrue('__roles__' in Application.misc_.__dict__)
+        self.assertTrue(hasattr(Application.misc_, '__roles__'))


pbauer:
@hannosch Could the following traceback (during statup) be somehow related to the change to new-style classes? I've never encountered it before I used the checkout of Zope today.

```
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/PubCore/ZServerPublisher.py", line 31, in __init__
    response=b)
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/ZPublisher/Publish.py", line 412, in publish_module
    environ, debug, request, response)
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/ZPublisher/Publish.py", line 300, in publish_module_standard
    request.close()
  File "/Users/pbauer/workspace/coredev/src/Zope/src/ZPublisher/BaseRequest.py", line 220, in close
    notify(EndRequestEvent(None, self))
  File "/Users/pbauer/.cache/buildout/eggs/zope.event-4.3.0-py2.7.egg/zope/event/__init__.py", line 32, in notify
    subscriber(event)
  File "/Users/pbauer/.cache/buildout/eggs/zope.component-4.4.0-py2.7.egg/zope/component/event.py", line 27, in dispatch
    component_subscribers(event, None)
  File "/Users/pbauer/.cache/buildout/eggs/zope.component-4.4.0-py2.7.egg/zope/component/_api.py", line 139, in subscribers
    return sitemanager.subscribers(objects, interface)
  File "/Users/pbauer/.cache/buildout/eggs/zope.interface-4.4.2-py2.7-macosx-10.12-x86_64.egg/zope/interface/registry.py", line 442, in subscribers
    if getattr(self.adapters, 'subscribers'):
AttributeError: 'PersistentAdapterRegistry' object has no attribute 'subscribers'
```

hannosch:
That traceback is weird. There is no getattr call on that line in zope.interface 4.4.2 (https://github.com/zopefoundation/zope.interface/blob/4.4.2/src/zope/interface/registry.py#L442).

The PersistentAdapterRegistry comes from zope.component and has some inline-migration code in its getstate method, but that doesn't remove the `self.adapters` part.

Either you have some really really old never migrated persistent object at hand or something is monkey-patching this and is now breaking things.

hannosch:
Eh. I read that wrong. Forget what I said in my 2nd/3rd sentence, I'm way too under-caffeinated :(

pbauer:
@hannosch Sorry, I pasted the wrong traceback. The one above was from a botched attempt to see what happpanes when I ignored the AttributeError.

```
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/PubCore/ZServerPublisher.py", line 31, in __init__
    response=b)
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/ZPublisher/Publish.py", line 412, in publish_module
    environ, debug, request, response)
  File "/Users/pbauer/.cache/buildout/eggs/ZServer-4.0a2-py2.7.egg/ZServer/ZPublisher/Publish.py", line 300, in publish_module_standard
    request.close()
  File "/Users/pbauer/workspace/coredev/src/Zope/src/ZPublisher/BaseRequest.py", line 220, in close
    notify(EndRequestEvent(None, self))
  File "/Users/pbauer/.cache/buildout/eggs/zope.event-4.3.0-py2.7.egg/zope/event/__init__.py", line 32, in notify
    subscriber(event)
  File "/Users/pbauer/.cache/buildout/eggs/zope.component-4.4.0-py2.7.egg/zope/component/event.py", line 27, in dispatch
    component_subscribers(event, None)
  File "/Users/pbauer/.cache/buildout/eggs/zope.component-4.4.0-py2.7.egg/zope/component/_api.py", line 139, in subscribers
    return sitemanager.subscribers(objects, interface)
  File "/Users/pbauer/.cache/buildout/eggs/zope.interface-4.4.2-py2.7-macosx-10.12-x86_64.egg/zope/interface/registry.py", line 442, in subs
    return self.adapters.subscribers(objects, provided)
AttributeError: 'PersistentAdapterRegistry' object has no attribute 'subscribers'
```

hannosch:
Ah :))

One thing you could try is downgrading ZODB from 5.3.x to 5.2.4. It switched from pickle protocol 1 to 2, which changes the way new-style classes are pickled. Those registries try to keep some values out of the persistent state and re-instantiate them in-memory only. It might be that some of them now need a `__getinitargs__` or `__getnewargs__` or some other change to `__get/setstate__`.

The classes themselves all come from zope.interface and component and probably always have been new-style classes.

hannosch:
Hah, there's this nice warning message on the pickle protocol page (https://docs.python.org/2/library/pickle.html):

> Note: For new-style classes, if __getstate__() returns a false value, the __setstate__() method will not be called.

I bet that's the problem and the code in https://github.com/zopefoundation/zope.component/blob/master/src/zope/component/persistentregistry.py is at fault. In the getstate it removes all those memory-only bits (everything in _delegated), but that probably means it returns an empty value at first, as there's nothing in it.

When unpickling things, `__init__` isn't called, so this code relies on the `self._createLookup()` to be called as part of `__setstate__` to put those in-memory objects back. Only this no longer happens thanks to the empty `__getstate__` return value.

I think that PersistentAdapterRegistry needs to implement a `__new__` to populate its invariants, maybe a `__getnewargs__` as well.

hannosch:
Filed under zopefoundation/zope.component#34.

PR #199, opened 2017-10-15
Raise app exceptions without rendering response if wsgi.handleErrors is False

davisagli:
Raise the publish exception immediately rather than trying to return an exception response if the client requested this by setting the `wsgi.handleErrors` environ key to False. This is set by zope.testbrowser when the browser's `handleErrors` attribute is set to False. This restores behavior from Zope 2 that a number of Plone tests depend on in order to be able to inspect the actual exception raised by the app rather than a rendered response.

### DIFF
--- a/src/ZPublisher/WSGIPublisher.py
+++ b/src/ZPublisher/WSGIPublisher.py
@@ -142,6 +142,11 @@ def transaction_pubevents(request, response, tm=transaction.manager):
         # Create new exc_info with the upgraded exception.
         exc_info = (exc_type, exc, sys.exc_info()[2])
 
+        # Raise exception from app if handle-errors is False
+        # (set by zope.testbrowser in some cases)
+        if not request.environ.get('wsgi.handleErrors', True):
+            reraise(*exc_info)
+
         if isinstance(exc, Unauthorized):
             # _unauthorized modifies the response in-place. If this hook
             # is used, an exception view for Unauthorized has to merge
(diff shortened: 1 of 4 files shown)


src/ZPublisher/WSGIPublisher.py
```diff
@@ -142,6 +142,11 @@ def transaction_pubevents(request, response, tm=transaction.manager):
         # Create new exc_info with the upgraded exception.
         exc_info = (exc_type, exc, sys.exc_info()[2])
 
+        # Raise exception from app if handle-errors is False
+        # (set by zope.testbrowser in some cases)
+        if not request.environ.get('wsgi.handleErrors', True):
+            reraise(*exc_info)
+
```
hannosch: Doesn't this have to be inside the `try/finally` block to delete the local `exc` / `exc_info` variables and avoid the reference cycle?

hannosch: I mean this kind of cycle: https://cosmicpercolator.com/2016/01/13/exception-leaks-in-python-2-and-3/

davisagli:
Thanks for catching that; fixed.

pbauer:
I did not want to wait for a merge so I started a build with this branch at http://jenkins.plone.org/view/PLIPs/job/plip-zope4/58

icemac:
@davisagli The [WSGI-Spec](http://wsgi.readthedocs.io/en/latest/specifications/throw_errors.html?highlight=wsgi.handleErrors#specification) insists that the environment variable should be called `x-wsgiorg.throw_errors`. `zope.testbrowser` even sets this name, too. Maybe it could be used here, too.

davisagli:
@icemac interesting. I'm fine with that.

PR #200, opened 2017-10-17
revert to callable() check

davisagli:
Go back to using the `callable` builtin to see if an object is callable. We saw problems with some views that were instances of collections.Callable but did not actually have a `__call__` method. And while `callable` was removed for Python 3.0-3.1, it has returned.


jamadden:
I think the test is flaky. Restarting it fixed the 3.5 issue.

PR #201, opened 2017-10-17
Fix path expressions with uncallable views

davisagli:
This fixes a regression from Zope 2: if a browser view does not implement `__call__`, traversing to it in a path expression should return the view instance itself rather than raising an AttributeError. `callable()` was returning a false positive in this case, probably tricked by the `__call__` property on the view superclass.

(Zope 2 caught the AttributeError and checked to see if its parameter was `__call__`, but I figure catching NotImplementedError matches the intent better.)

### DIFF
--- a/src/Products/PageTemplates/tests/testExpressions.py
+++ b/src/Products/PageTemplates/tests/testExpressions.py
@@ -1,8 +1,7 @@
 # *-* coding: iso-8859-1 -*-
 
-import sys
 import unittest
-from six import text_type, binary_type
+from six import text_type
 
 from zope.component.testing import PlacelessSetup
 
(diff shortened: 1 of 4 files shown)


src/Products/PageTemplates/tests/testExpressions.py
```diff
@@ -90,6 +89,15 @@ def test_evaluate_with_render_simple_callable(self):
         ec = self._makeContext()
         self.assertEqual(ec.evaluate('dummy'), 'dummy')
 
+    def test_evaluate_with_unimplemented_call(self):
+        class Dummy(object):
+            def __call__(self):
+                raise NotImplementedError
```
icemac: I'd suggest to raise an instance of `NotImplementedError` instead of the class so Python 3 can store the traceback on it without side effects.

jamadden: Interestingly, [Python automatically turns](https://github.com/python/cpython/blob/master/Python/ceval.c#L4140) `raise <type>` into `raise <type>()`---that is, if raise is given a class argument, Python creates a fresh instance and raises that. This turns out to be just a tiny bit faster than constructing the instance yourself (it's one bytecode shorter).

icemac: @jamadden Thank you for clarifying this.

PR #206, opened 2017-10-21
Fix 'lines' property to always use 'bytes'

witsch:
The 'lines' property type could not be used with bytes so far, as `str` will include the prefix and quotes:
```
$ python3.6
Python 3.6.3 (default, Oct 21 2017, 11:47:54) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> str(b'foo')
"b'foo'"
>>> 
```

This fixes the type to always store bytes as was already the case with Python 2.

### DIFF
--- a/src/ZPublisher/Converters.py
+++ b/src/ZPublisher/Converters.py
@@ -135,7 +135,7 @@ def field2lines(v):
     if isinstance(v, (list, tuple)):
         result = []
         for item in v:
-            result.append(str(item))
+            result.append(field2bytes(item))
         return result
     return field2text(v).splitlines()
 
--- a/src/ZPublisher/tests/test_Converters.py
+++ b/src/ZPublisher/tests/test_Converters.py
@@ -52,12 +52,14 @@ class ConvertersTests(unittest.TestCase):
     def test_field2lines_with_list(self):
         from ZPublisher.Converters import field2lines
         to_convert = ['one', 'two']
-        self.assertEqual(field2lines(to_convert), to_convert)
+        expected = [b'one', b'two']
+        self.assertEqual(field2lines(to_convert), expected)
 
     def test_field2lines_with_tuple(self):
         from ZPublisher.Converters import field2lines
         to_convert = ('one', 'two')
-        self.assertEqual(field2lines(to_convert), list(to_convert))
+        expected = [b'one', b'two']
+        self.assertEqual(field2lines(to_convert), expected)
 
     def test_field2lines_with_empty_string(self):
         from ZPublisher.Converters import field2lines
--- a/src/ZPublisher/tests/test_Converters.py
+++ b/src/ZPublisher/tests/test_Converters.py
@@ -51,13 +51,13 @@ class ConvertersTests(unittest.TestCase):
 
     def test_field2lines_with_list(self):
         from ZPublisher.Converters import field2lines
-        to_convert = ['one', 'two']
+        to_convert = ['one', b'two']
         expected = [b'one', b'two']
         self.assertEqual(field2lines(to_convert), expected)
 
     def test_field2lines_with_tuple(self):
         from ZPublisher.Converters import field2lines
-        to_convert = ('one', 'two')
+        to_convert = ('one', b'two')
         expected = [b'one', b'two']
         self.assertEqual(field2lines(to_convert), expected)
 


src/ZPublisher/Converters.py
```diff
@@ -135,7 +135,7 @@ def field2lines(v):
     if isinstance(v, (list, tuple)):
         result = []
         for item in v:
-            result.append(str(item))
+            result.append(field2bytes(item))
         return result
     return field2text(v).splitlines()
```
hannosch: Having the last return statement with `field2text` (which calls `field2string`) and the changed one using `field2bytes` looks inconsistent to me. Does the last line also have to be changed to use `field2bytes`?

tseaver:
I'm pretty sure that a `lines` property should be returning lists of text, not lists of bytes.

witsch:
Hi Tres,

David, Godefroid and I had a pretty lengthy discussion about this and we feel people should use ‘ulines' if they need lines of text while ‘lines’ should essentially keep using the same data.  Or in other words, `str` in Python 2 is way closer to bytes in Python 3 than it is to text…  If we use lists of text here, we’ll likely end up with (utf-8) encoded data so making this bytes hopefully makes this more obvious (and saves us some guessing later on).

src/ZPublisher/Converters.py
witsch: I agree, but I guess we should first sort out and agree which direction we want to take, i.e. lists of text vs lists of bytes.  There are no hard reasons for either I suppose, but like I said, we think "bytes" feel more right.

In any case, the `str` is wrong and cannot stay (imho).

hannosch:
Hey. I think we had a similar lengthy discussion at some sprint and went the other way, where we tried to make all the normal converters (without the u prefix) use native strings. That is use the str type under both Python 2 and 3, even though they mean something very different.

The reasoning was that almost nobody ever used the u-prefixed converters and most people used ascii-only content with these converters.

But that means you'd have to use some other converter if you actually wanted binary content.

I don't remember all the details, and am not convinced the way we decided to do this before is the right way. So I'm all open for arguments and changing this all again.

davisagli:
Hmm. I think using native strings makes fine sense for when these converters are used to process request params. But they are also used when values are stored by a PropertyManager, which is the case that @witsch and @gotcha were working on. I'm concerned that if this converter returns native strings, then after loading a database created with Zope 2 into Zope 4 on Python 3, existing values of `lines` properties will be bytes but newly written ones will be stored as text. Or is this an example of something that needs to be dealt with by app-specific migration/db conversion code?

leorochael:
Back when these converters like `lines` were created, there was no Unicode in Python (and no u-converters like `ulines`.

So I agree most people using `lines` indeed intended to get lines of text, and this is what they got, in the form of a list of (byte) strings that just so happened to contain bytes in whatever encoding they were using (most likely iso-8859-1).

These lines of text encoded as bytestrings were 
then likely sent to whatever storage or processing back-end was in use, with the expectation that they would be processed "as is".

So while I like the argument that "lines of text" is what people expect when they use "lines" (hence  we should be giving them a list of Unicode strings), in practice this will probably break existing applications that are migrating to Python 3 (and are already facing a whole set of challenges) throwing Unicode errors far from the point of cause, and with no quick workaround (since there is no `blines` converter).

IMO, `lines` returning a list of Unicode strings only on Python 3 is DWIMy, and goes against at least 3 entries in the Zen of Python (EIBTI, ITFOARTTTG, TSBOOWTDI). I think it should return "list of bytes" in both Python versions.

tseaver:
OK, I'll bow to the consensus.

icemac:
Is there still something to do before this PR can be merged?

gotcha:
@hannosch comment (https://github.com/zopefoundation/Zope/pull/206/files#r146114998) still need to be addressed.

witsch:
I can have a look on monday (I meant to wrap this up anyway :))

witsch:
I finally looked into this again and ended up pushing a [work-in-progress commit](https://github.com/zopefoundation/Zope/pull/206/commits/104def6911812cea7d6dfc70b340816bbe14772f) I had already started back at the sprint.  Actually I meant to amend it in order to fix the two remaining test errors, which (both) look like this:
```python
Error in test test_processInputs_w_simple_marshalling (ZPublisher.tests.testHTTPRequest.HTTPRequestTests)
  File "…/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "…/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "…/ZPublisher/tests/testHTTPRequest.py", line 277, in test_processInputs_w_simple_marshalling
    req = self._processInputs(inputs)
  File "…/ZPublisher/tests/testHTTPRequest.py", line 146, in _processInputs
    req.processInputs()
  File "…/ZPublisher/HTTPRequest.py", line 698, in processInputs
    if '<' in tainted[i]:
TypeError: a bytes-like object is required, not 'str'
```
However, after digging a bit deeper I'm not sure how to best solve this.  If we change `lines` to expect and return lists of bytes, we'll probably also need a `TaintedBytes` class (in [`AccessControl/tainted.py`](https://github.com/zopefoundation/AccessControl/blob/master/src/AccessControl/tainted.py#L32)) or make `TaintedString` "byte-aware" somehow (because the instance holding the "tainted" byte sequence also needs to return bytes again).

But before giving this a try, I'd like to hear your thoughts about it… :)

gotcha:
I explore how much work it was to add TaintedBytes. Not that much actually.
Needs review though !

icemac:
@gotcha Do you think this PR can be merged now?

gotcha:
Thanks to you @icemac for following up !

PR #208, opened 2017-10-23
Fix unpickling of instances those base class changed to a new-style class.

### LINKED ISSUE #205: Migrate Plone site from Zope2 2.13 to Zope2 4.0b3.dev0
python version : 2.7.14

Steps to reproduce : 
- create Plone site on Zope2 2.13 (Plone 5.1) (with [buildout.coredev](https://github.com/plone/buildout.coredev/))
- restart buildout with "-c plip-zope4.cfg" option (use Zope2 4.0b3) 
- start instance

```
2017-10-21 11:34:22 ERROR ZODB.Connection Couldn't load state for OFS.Application.Application 0x01 
  File "/home/bsuttor/.buildout/eggs/ZODB-5.3.0-py2.7.egg/ZODB/Connection.py", line 800, in setstate
    self._reader.setGhostState(obj, p) 
  File "/home/bsuttor/.buildout/eggs/ZODB-5.3.0-py2.7.egg/ZODB/serialize.py", line 622, in setGhostState
    state = self.getState(pickle)
  File "/home/bsuttor/.buildout/eggs/ZODB-5.3.0-py2.7.egg/ZODB/serialize.py", line 615, in getState
    return unpickler.load()         
TypeError: ('__init__() takes exactly 2 arguments (1 given)', <class 'ZPublisher.BeforeTraverse.NameCaller'>, ())  
```

icemac:
If the instance was created before Zope 4.0b2 unpickling broke because in 4.0b2 all classes became new-style classes.

Fixes #205.

The proposes solution seems to work as shown in #205.
Could anyone think of a cleaner solution?

### DIFF
--- a/src/ZPublisher/BeforeTraverse.py
+++ b/src/ZPublisher/BeforeTraverse.py
@@ -86,7 +86,8 @@ class MultiHook(object):
     MultiHook calls the named hook from the class of the container, then
     the prior hook, then all the hooks in its list.
     """
-    def __init__(self, hookname=None, prior=None, defined_in_class=None):
+    def __init__(self, hookname='<undefined hookname>', prior=None,
+                 defined_in_class=False):
         # The default values are needed for unpickling instances of this class
         # which where created before 4.0b2 where this class was still an old
         # style class. For details see
@@ -124,7 +125,7 @@ class NameCaller(object):
     >>> registerBeforeTraverse(folder, NameCaller('preop'), 'XApp')
     """
 
-    def __init__(self, name=None):
+    def __init__(self, name='<undefined name>'):
         # The default value is needed for unpickling instances of this class
         # which where created before 4.0b2 where this class was still an old
         # style class. For details see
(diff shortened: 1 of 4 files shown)


src/ZPublisher/BeforeTraverse.py
```diff
@@ -86,7 +86,11 @@ class MultiHook(object):
     MultiHook calls the named hook from the class of the container, then
     the prior hook, then all the hooks in its list.
     """
-    def __init__(self, hookname, prior, defined_in_class):
+    def __init__(self, hookname=None, prior=None, defined_in_class=None):
```
tseaver: I think an explicit string would be a better default for `hookname` (maybe `'<undefined hookname>'`?) and `False` would be a better default for `defined_in_class`.

src/ZPublisher/BeforeTraverse.py
```diff
@@ -120,7 +124,11 @@ class NameCaller(object):
     >>> registerBeforeTraverse(folder, NameCaller('preop'), 'XApp')
     """
 
-    def __init__(self, name):
+    def __init__(self, name=None):
```
tseaver: Another case for an explicit string, rather than `None`:  maybe `'<undefined name>'`?

icemac:
@hannosch @jamadden Thank you for your review.
@tseaver You are right: Using the default arguments should not lead to an exception. Are you fine with my changes?

icemac:
@tseaver Thank you!

PR #174, opened 2017-09-14
DO NOT MERGE! Plonezope4

MrTango:
Will be merged in single PR's:

- [x] #172 
- [X] #215

PR #215, opened 2017-10-27
Avoid error when calling setBody with lock on a wsgi response.

pbauer:
Same as the unmerged commit https://github.com/zopefoundation/Zope/pull/174/commits/727612db36456ea389a7b612acc1046bb975664b by @mauritsvanrees from https://github.com/zopefoundation/Zope/pull/174

### DIFF
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -1075,7 +1075,7 @@ class WSGIResponse(HTTPBaseResponse):
 
         self.stdout.write(data)
 
-    def setBody(self, body, title='', is_error=False):
+    def setBody(self, body, title='', is_error=False, lock=None):
         if isinstance(body, IOBase):
             body.seek(0, 2)
             length = body.tell()


tseaver:
I wonder if we should actually be emulating the *semantics* of `HTTPResponse.setBody`, rather than just its signature:  if the application is passing `lock=True` to the method, then it will be expecting that future calls to `setBody` will be no-ops.

pbauer:
Not sure about that, in Plone locking is implemented on another level. I think these locks were only used by webdav. 
Anyway, I'd love to see this merged since we really need to get the tests for Plone 5.2 green (see https://github.com/plone/Products.CMFPlone/issues/2195) to be able to merge other work (e.g. we made the first steps towards py3 support in Plone).

tseaver:
I'm OK with a merge-right-now-to-fix-Plone-tests:  can you open an issue to track whether `WSGIResponse.setBody` should mimic the `HTTPResponse.setBody` if `lock` is truthy?

PR #218, opened 2017-11-03
Install a default page for the root view in new installations again, fixes #212

### LINKED ISSUE #212: No default view for root object
Zope 2.13 had this well known, outdated greetings page. Zope 4 has:

z4

This is not very welcoming even for users switching form older Zope versions or trying out Zope 4.
I propose to add a minimal `index_html` back to the default ZODB to get rid of this Waaa effect.

dataflake:


pbauer:
👍 This restores the view that plone uses on the zope-root with https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/admin.py#L41

PR #219, opened 2017-11-05
Removed reference to old action.html

### LINKED ISSUE #211: Security declaration warning at startup
When starting Zope 4.0b2 it issues the following warning:

```
2017-10-24 11:24:46 - WARNI [Init] Class Products.Five.browser.metaconfigure.+ has a security declaration for nonexistent method 'action.html'
```

This might confuse users. What can de done to get rid of this warning?

jan-jockusch:
Fixes #211

This is more of a blind stab at the problem. I couldn't find any other plausible references, so I tentatively removed this line (which doesn't make a lot of sense to me anyway). The warning is now gone. I've seen no obvious problems.

But I am completely ignorant as to potential side effects, so I'm posting this in the hope that someone more knowlegeable will confirm this is OK.

### DIFF
--- a/src/Products/Five/browser/configure.zcml
+++ b/src/Products/Five/browser/configure.zcml
@@ -16,7 +16,6 @@
       >
 
     <browser:page name="index.html"  template="adding.pt" />
-    <browser:page name="action.html" attribute="action" />
 
   </browser:view>
 


hannosch:
This is probably ok.

The `adding.pt` file has the contents `+ screen not yet supported by Five`.

I think this is all part of some unfinished support for Zope Toolkit adding views and the ZTK skin system. We could probably remove more of this code, but not in the beta cycle unless there's a good reason for it.

icemac:
@jan-jockusch Thank you for fixing this. You are heartily invited to sign a contributor agreement: http://docs.zope.org/developer/becoming-a-committer.html if you are planning bigger contributions in the future.

PR #221, opened 2017-11-06
fix regression in traversal of plone.app.folder.base.ReplaceableWrapper

davisagli:
This restores a webdav-related check to how it was done in Zope 2 to fix an edge case causing a Plone test to fail.

### DIFF
--- a/src/ZPublisher/BaseRequest.py
+++ b/src/ZPublisher/BaseRequest.py
@@ -475,8 +475,7 @@ class BaseRequest(object):
                             hasattr(object, 'aq_base') and
                             not hasattr(object, '__bobo_traverse__')):
 
-                        if (object.__parent__ is not
-                                aq_inner(object).__parent__):
+                        if object.__parent__ is not object.aq_inner.__parent__:
                             object = NullResource(parents[-2], object.getId(),
                                                   self).__of__(parents[-2])
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,10 @@ Bugfixes
 
 - Prevent UnicodeDecodeError when publishing image (bytes) responses without content-type
 
+- Fix regression in traversing to WebDAV NullResources if the final object
+  in the path is a proxy that provides access to aq_inner but is not itself
+  an acquisition wrapper.
+
 Changes
 +++++++
 


src/ZPublisher/BaseRequest.py
```diff
@@ -475,8 +475,7 @@ def traverse(self, path, response=None, validated_hook=None):
                             hasattr(object, 'aq_base') and
                             not hasattr(object, '__bobo_traverse__')):
 
-                        if (object.__parent__ is not
-                                aq_inner(object).__parent__):
```
tseaver: Can we factor the check out such that we use `aq_inner(object)` first, and fall back to the proxied `\.aq_inner`?

davisagli: Hmm, I'm not sure how we would tell when to fall back, because I think aq_inner returns the object you pass in both when it's already the inner acquisition wrapper chain and when the object is not an acquisition wrapper.

I'm happy to change the Plone code that depends on this if I can figure out a good alternative. The proxy in question is here -- https://github.com/plone/plone.app.folder/blob/master/src/plone/app/folder/base.py#L20 -- and exists solely to make sure that the return value from `index_html` further down has a particular `__replaceable__` attribute. In the first place it's used we're instantiating a new `NullResource` and could just set `__replaceable__` directly, but we shouldn't do that on the persistent object that's proxied in the second place.

jamadden: Please forgive the observation of someone who, while familiar with acquisition, is not extremely well versed in this application.  The changes in question are a bit unclear. 

The fact that this was changed without repercussions in this test base, but led to issues elsewhere suggests that a proposed change deserves a test case. I couldn’t help but notice no test cases in this PR.

davisagli: I was hoping that I could get by without adding a test here since it's fixing an existing test case in Plone. But I agree that the change is pretty questionable and I'd rather figure out how to make Plone stop depending on the behavior that changed.

tseaver:
@davisagli What does the Plone traceback / failure look like?

davisagli:
I'm going to withdraw this until I can study the use case a bit more closely.

PR #228, opened 2017-11-30
Make Redirect unavailable as url [master]

mauritsvanrees:
Part of PloneHotfix20171128.


mauritsvanrees:
I had a buildout error locally at first, which went away when I added the extra index. But I agree it should not be needed with zc.buildout 2.9.5. Maybe a temporary problem. Thanks for improving the pull request.

PR #229, opened 2017-12-01
Crash-proof the 2.13 branch

dataflake:
The 2.13 branch currently suffers from crash bug zopefoundation/Acquisition#21. I propose a version update to the latest ``Acquisition`` release, which also necessitates upgrading ``ExtensionClass`` and ``Products.BTreeFolder2``. This will also require removing Python 2.6 support, which IMHO is not a worthwhile goal at this point.

The tests run through except for a cosmetic failure in a ``Persistence`` test where an exception string coming out of the updated ``ExtensionClass`` looks slightly different. I did not want to pin a new ``Persistence`` version, though, for fear of a rat's tail of additional upgrades.

I also have anecdotal evidence that these new version pins don't affect Zope stability. A customer site that crashed a dozen times a day due to the ``Acquisition`` crash bug has been running stable for over a week now with the proposed version pins.


hannosch:
Is the Acquisition crash bug the only reason for the crashes? If so, shouldn't we backport it to the Acquisition branch used by 2.13 and use a new minor release of that branch?

I'm hesitant to update so many packages with C extensions to new major versions. They are all breaking backwards compatibility on the C level, as they switch to the new PyCapsule API. So if anyone has custom C extensions extending ExtensionClass or Acquisition, those would completely break with these new versions.

dataflake:
Yes, that makes sense. I will try my hand at a backport today.

dataflake:
I'll close this, the new ``Acquisition`` 2.13.12 should fix it.

PR #232, opened 2017-12-04
Test that `str.format` checks security for accessed keys and items.

mauritsvanrees:
The real fix is in the AccessControl package, which we checkout on a specific branch here. See https://github.com/zopefoundation/AccessControl/pull/50

Part of PloneHotfix20171128. Some of the tests here were created by @icemac and passed to the Plone security team. Thank you!

We should wait till the AccessControl branch is merged, and then we can use a checkout of the normal 2.13 branch again, or even a release.

I can do the same on the Zope master branch, but it seems better to wait for approval on this PR for the 2.13 branch first.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -23,7 +23,6 @@ parts =
     wsgi
 sources-dir = develop
 auto-checkout =
-    AccessControl
 versions = versions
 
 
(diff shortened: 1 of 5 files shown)


mauritsvanrees:
That is fine with me. But the Travis tests at least pass, which is good to know.

icemac:
It seems that this PR has to wait for zopefoundation/AccessControl#50 to be merged and released, right?

dataflake:
@icemac correct, when that is done the changes to ``buildout.cfg`` and ``sources.cfg`` in this PR can (and should) go.

PR #237, opened 2017-12-22
Add older releases which previously were on zope.org.

### LINKED ISSUE #234: Copy older Zope releases on GH pages
As [zope.org](http://zope.org) is currently expired I'd suggest to collect the older Zope releases and put them on https://zopefoundation.github.io/Zope/ aka https://github.com/zopefoundation/Zope/tree/gh-pages to be no longer dependent on the old infrastructure. Even the installation instructions could be moved over there.

Using the GitHub tags page (e. g. https://github.com/zopefoundation/Zope/releases/tag/2.10.13) is not enough as it does not always contain all the necessary code (e. g. 2.10.13 is missing at least `BTrees`).

This change should be promoted on Zope's RTD page, too.

icemac:
Relates to #234


icemac:
These releases I found on my hard disk. Are there more releases we should add?

icemac:
@hannosch I added a section for Zope 2.12. Is it okay for you or did you mean something different?

PR #256, opened 2018-03-21
Use correct stdout stream in publish_module.

icemac:
If `publish_module` is called with a `_response` parameter which is not `None` its stdout is written to but is not read from.

Detected in zopefoundation/Products.ExternalEditor#10.


icemac:
@hannosch I fixed the tests (I did no expect them to exist – shame on me!) but they helped me to write a test which breaks without the change. Are you fine with my changes?

icemac:
@hannosch Ping.

PR #259, opened 2018-03-22
fix regression in doctest functional helper: sort headers

icemac:
I took this commit from the `plonezope4` branch aka #174.
Is it still relevant?

### DIFF
--- a/src/Testing/ZopeTestCase/functional.py
+++ b/src/Testing/ZopeTestCase/functional.py
@@ -153,17 +153,7 @@ class ResponseWrapper(object):
         return self.getOutput()
 
     def __str__(self):
-        out = self.getOutput()
-        if PY2:
-            return out
-        # This is a hack. This method is called to print a response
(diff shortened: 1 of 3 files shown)


tseaver:
Is this a workaround for the bug fixed in #256?

icemac:
@tseaver I do not think so. It seems to solve the problem that the HTTP headers in Zope 4 are not sorted and in random order which even changes between test runs. This makes it difficult to write fix doctests which assert the raw HTTP response. See https://github.com/zopefoundation/Products.ExternalEditor/pull/10/files#diff-cd5c80de291398a1c9b1b8bb98e45e4dR40 what I had to do in Products.ExternalEditor which uses this approach heavily.

icemac:
This PR help a lot to port doctests to Zope 4. (See zopefoundation/Products.ExternalEditor#10 where I could revert all my changes in the doctest regarding the response checks after using this branch.)
I strongly like to get this PR merged to ease the migration to Zope 4.

PR #260, opened 2018-03-22
Fix sorting of a list of dicts of meta_types

pbauer:


src/OFS/ObjectManager.py
```diff
@@ -210,7 +210,7 @@ def __class_init__(self):
                         mt.append(t)
             except Exception:
                 pass
-        mt.sort()
+        mt.sort(key=lambda k: k['name'])
         self.meta_types = tuple(mt)
```
tseaver: Small nit:  I think `sorted` is preferable to an in-place sort, e.g.:

```python
        self.meta_types = tuple(sorted(mt, key=lambda k: k['name']))
```

Also, might be cleaner to use `operator.itemgetter('name')`?

PR #263, opened 2018-04-02
force cookies to be bytes 

pbauer:
to prevent wrapping bytes as text in py3, which ends up as `"b'somevalue'"`

I'm not sure about this because I heard that most browsers can deal with text/unicode cookies now.


pbauer:
O, damn. This leads to cookies like `b'__ac': "xxxx"`. I guess at least the key should be text.

PR #265, opened 2018-04-04
Fix DTML upload for Python 3.

icemac:
Without this change uploading a file to a DTMLMethod via ZMI breaks because `self.munge(file)` tries to use a string regex on a bytes object (the `file` variable). 

What do you think is this patch the way to go?

Open tasks:

- [x] Discuss solution
- [x] Write test (method is currently untested)
- [x] Port fix to DTMLDocument (it duplicates the code)
- [x] Write test for DTMLDocument

See zopefoundation/Products.SiteErrorLog#11 which uses this method in tests and fails with `AttributeError: 'str' object has no attribute 'read'`. in OFS/DTMLMethod.py:272

### DIFF
--- a/src/OFS/DTMLMethod.py
+++ b/src/OFS/DTMLMethod.py
@@ -27,7 +27,7 @@ from AccessControl.tainted import TaintedString
 from Acquisition import Implicit
 from DocumentTemplate.permissions import change_dtml_methods
 from DocumentTemplate.security import RestrictedDTML
-from six import binary_type
+from six import PY3
 from six.moves.urllib.parse import quote
 from zExceptions import Forbidden
 from zExceptions import ResourceLockedError
@@ -266,10 +266,12 @@ class DTMLMethod(RestrictedDTML,
         if self.wl_isLocked():
             raise ResourceLockedError('This DTML Method is locked.')
 
-        if not isinstance(file, binary_type):
+        if not isinstance(file, basestring):
             if REQUEST and not file:
                 raise ValueError('No file specified')
             file = file.read()
+            if PY3:
+                file = file.decode('utf-8')
 
         self.munge(file)
         self.ZCacheable_invalidate()
--- a/src/OFS/DTMLMethod.py
+++ b/src/OFS/DTMLMethod.py
@@ -44,8 +44,6 @@ from OFS.role import RoleManager
 from OFS.SimpleItem import Item_w__name__
 from ZPublisher.Iterators import IStreamIterator
 
-if PY3:
-    basestring = str
 
 _marker = []  # Create a new marker object.
 
--- a/src/OFS/DTMLMethod.py
+++ b/src/OFS/DTMLMethod.py
@@ -13,7 +13,6 @@
 """DTML Method objects.
 """
 import re
-import sys
 
 from AccessControl.class_init import InitializeClass
 from AccessControl.SecurityInfo import ClassSecurityInfo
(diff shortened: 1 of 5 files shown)


src/OFS/DTMLMethod.py
```diff
@@ -266,10 +266,12 @@ def manage_upload(self, file='', REQUEST=None):
         if self.wl_isLocked():
             raise ResourceLockedError('This DTML Method is locked.')
 
-        if not isinstance(file, binary_type):
+        if not isinstance(file, basestring):
```
davisagli: `basestring` is not defined in Python 3

icemac: @davisagli You are right, but someone defined it in the this Python module, see: https://github.com/zopefoundation/Zope/blob/bc180d5f2c2ae6099f888e1f8b4e520dbe6b46f9/src/OFS/DTMLMethod.py#L46

src/OFS/DTMLMethod.py
```diff
@@ -266,10 +266,12 @@ def manage_upload(self, file='', REQUEST=None):
...
             if REQUEST and not file:
                 raise ValueError('No file specified')
             file = file.read()
+            if PY3:
```
rbu: Shouldn't the conversion binary_type->text_type happen for Python 2 as well? I would assume the file type is always binary_type (in which case the method's default should also be changed to `file=b''`).

icemac: @rbu I think you are generally right. But there are databases out in the world which have stored the file contents as bytes (aka Python 2 `str`). Can we provide a way to convert their file contents to text so we do not have to live forever with file contents wich might be `bytes`?

src/OFS/DTMLMethod.py
rbu: That `if sys.version_info` could use the PY3 symbol imported above

src/OFS/DTMLMethod.py
rbu: Ok I see what you're saying. In the old code, Python 2 would always have `str` byte strings and byte file objects. Those were read, resulting in bytes being saved to ZODB. As I understand it, per the WSGI spec, `file` will usually be a bytes file object, so `read()`ing from it needs to be decoded. So the goal of this code is to always save byte strings in PY2 and always save text strings in PY3.

To make that consistent, though, the default method parameter needs to be decoded as well. What do you think of the following. What do you think of this:
```

        if not isinstance(file, basestring):
            if REQUEST and not file:
                raise ValueError('No file specified')
            file = file.read()

        if PY3 and isinstance(file, binary_type):
                file = file.decode('utf-8')
 ```

**Comment by rbu** [comment]
FYI: I've rebased this on top of master

File: src/OFS/DTMLMethod.py (hunk 1 above)
Comment (icemac) [reply src/OFS/DTMLMethod.py]: Although it is not completely the same it is close enough, as we will drop Python 2 support before Python 4 comes out. :)

File: src/OFS/DTMLMethod.py (hunk 2 above)
Comment (icemac) [reply src/OFS/DTMLMethod.py]: You are right if `manage_upload` is called with a `file` parameter which is `bytes` it should be decoded, too.

**Comment by leorochael** [comment]
@icemac it sounds reasonable that both `manage_upload` and `manage_edit` store the same thing.

File: src/OFS/DTMLMethod.py [hunk 3]
```diff
@@ -42,7 +44,7 @@
 from OFS.SimpleItem import Item_w__name__
 from ZPublisher.Iterators import IStreamIterator
 
-if sys.version_info >= (3, ):
+if PY3:
     basestring = str
```
rbu: In the current state, you don't need the ```basestring``` symbol anymore.

PR #267, opened 2018-04-15
OFS.Traversable: use globalrequest

thet:
In OFS.Traversable, use ``zope.globalrequest.getRequest`` throughout instead of acquiring the request.

I had a problem with plone tiles which use plone.subrequest - I cache relationfield result items on the parent request, but those result items lose their ``URL`` info from the ``REQUEST`` attribute (which seems to be stored on the portal root where we have to ``aq_acquire`` to).

Because we no not need to traverse to the portal root via attribute acquisition, the getRequest method is faster:

Speed comparison for a context with a path like: ``/Plone/folder/folder/folder``

    >>> # Acquiring the REQUEST
    >>> from datetime import datetime; dt1 = datetime.now(); [aq_acquire(self, 'REQUEST') for cnt in range(100)]; (datetime.now() - dt1).total_seconds()

min: 0.001813
max: 0.003554

    >>> # Using zope.globalrequest
    >>> from zope.globalrequest import getRequest
    >>> from datetime import datetime; dt1 = datetime.now(); [getRequest() for cnt in range(100)]; (datetime.now() - dt1).total_seconds()

min: 0.00051
max: 0.001266

The timing difference might not be very significant, but at least zope.globalrequest is more efficient than traversing up the stack.

Before I go further with this I wait for comments.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.0b4 (unreleased)
 ------------------
 
+- In OFS.Traversable, use ``zope.globalrequest.getRequest`` throughout instead of acquiring the request.
+
 - Fix an edge case where the data which was set using ``response.write()`` was
   not returned by ``publish_module`` (#256).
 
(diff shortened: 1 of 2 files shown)


dataflake:
@thet Is this still active? Please let us know. If not I will close this in a week.

dataflake:
Closing due to lack of activity

Issue #217, opened 2017-11-01
manage_pasteObjects breaks when pasting many objects


icemac:
[_cb_decode](https://github.com/zopefoundation/Zope/blob/2.13/src/OFS/CopySupport.py#L686) only allows 8 kB of compressed copy data.
This was introduced in https://github.com/zopefoundation/Zope/commit/6ca369193cb3292aca0824216dbe1775c65ff8be to prevent a DoS attack.

The curent implementation also prevents pasting many objects in file system code be calling e.g.:

```python 

dest.manage_pasteObjects(source.manage_copyObjects(ids))
```

I propose a new parameter to `manage_pasteObjects` to override the `maxsize` value of `_cb_decode` to support this use case again or to use a different value if `cb_copy_data` is not `None` when calling `manage_pasteObjects`.

What do you think of my proposal?

This issue exists on Zope 2.13 and Zope 4.

See also https://stackoverflow.com/questions/17581962/copyerror-ofs-copysupport-manage-pasteobjects-limited-to-160-objects

tseaver:
ISTM that allowing TTW code to override that count is a Bad Thing (TM).

Maybe split the methods into the TTW-callable `manage_{copy,paste}Objects` and an underlying `_{copy,paste}Objects` which takes that flag?

jan-jockusch:
In our use cases, using the TTW interface to move massive amounts of code via copy/paste is a normal thing in Zope. I'd have a hard time telling my users that they're attempting DoS in version 4 with an action that was totally OK in version 2.13. 
If this is a serious concern, can we make the feature at least configurable, please? I can attempt a patch (since I'll have to make sure this works for my installations anyway).

hannosch:
We could also make the default size a zope.conf option and look it up from there.

I think gunicorn's `limit-request-field-size` (http://docs.gunicorn.org/en/stable/settings.html#limit-request-field-size) and waitress' `max_request_header_size` (https://docs.pylonsproject.org/projects/waitress/en/latest/arguments.html) serve a similar purpose here. Those apply length constraints on the size of all headers or on each header. A per-header upper length limit would already limit the size of the `Set-Cookie` header, so maybe this would be enough here?

Otherwise we could make it an additional config option, specific to zlib decompression.

icemac:
[...]

PR #269, opened 2018-04-18
Add ability to paste objects without limits.

icemac:
6ca3691 introduces code to prevent a DoS attack.
This is PR mitigates the consequences for file system code by allowing to paste without limits as discussed in #217.

This PR is for `master` if it is accepted I'm going to create one for `2.13`, too.


icemac:
@tseaver Thank you!

PR #270, opened 2018-04-19
Add ability to paste objects without limits. [2.13]

icemac:
6ca3691 introduces code to prevent a DoS attack.
This is PR mitigates the consequences for file system code by allowing to paste without limits as discussed in #217 for Zope 2.13.

### DIFF
--- a/src/OFS/tests/testCopySupport.py
+++ b/src/OFS/tests/testCopySupport.py
@@ -307,6 +307,11 @@ class TestCopySupport( CopySupportTestBase ):
             self.folder1.manage_pasteObjects(make_data(300))
         self.assertTrue('Item Not Found' in str(err.exception))
 
+        # _pasteObjects allows to paste without restriction:
+        with self.assertRaises(CopyError) as err:
+            self.folder1._pasteObjects(make_data(3500))
+        self.assertTrue('Item Not Found' in str(err.exception))
+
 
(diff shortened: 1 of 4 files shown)


src/OFS/tests/testCopySupport.py
```diff
@@ -285,32 +285,38 @@ def testPasteMultiSameID( self ):
...
         from OFS.CopySupport import CopyError
-        with self.assertRaises(CopyError):
-            self.folder1.manage_pasteObjects()
+        try:
+            callable(*args)
+        except CopyError as err:
+            if error_text:
+                self.assertTrue(error_text in str(err))
+        else:
+            self.fail('CopyError not raised.')
```
tseaver: Ugh. Python 2.6 has `assertRaises` as a method -- it just doesn't allow it as a context manager.  ISTM that this would be cleaner as `self.assertRaises(CopyError, callable, *args)` and just drop the assertion about the error text (I don't care for those assertions anyway).

icemac: @tseaver I know, this is not the nicest code but unfortunately the whole test is about the error text: All three cases result in a `CopyError` but only the error text `Clipboard Error` (Line 310) signals that the the clipboard contents were too big.

An alternative could be to add a dependency to `unittest2` to get more modern unittest assertions but I do not really like it either.

PR #273, opened 2018-05-09
Make sure we only compare ints in ZCacheable_getModTime

davisagli:
This fixes TypeErrors in Python 3 due to comparing an integer to None.


tox.ini
```diff
@@ -9,7 +9,7 @@ commands =
 skip_install = true
 deps =
     coverage
-    setuptools==33.1.1
+    setuptools==39.0.1
```
davisagli: Drive-by fix: this needs to match versions.cfg

src/OFS/Cache.py
```diff
@@ -226,10 +226,12 @@ def ZCacheable_getModTime(self, mtime_func=None):
             # Allow mtime_func to influence the mod time.
             mtime = mtime_func()
         base = aq_base(self)
-        mtime = max(getattr(base, '_p_mtime', mtime), mtime)
+        mtime = max(getattr(base, '_p_mtime', mtime) or 0, mtime)
         klass = getattr(base, '__class__', None)
         if klass:
-            mtime = max(getattr(klass, '_p_mtime', mtime), mtime)
+            klass_mtime = getattr(klass, '_p_mtime', mtime)
+            if isinstance(klass_mtime, int):
```
icemac: Is there a specific reason why you implemented it differently here (`isinstance`) and in line 229 (`or 0`)?

tox.ini
icemac: Good catch.

src/OFS/Cache.py
davisagli: I first tried doing it the same as line 229 here, but in the case that was being tested in Plone the getattr returned a property descriptor which is truthy but not an int, so I needed the isinstance check.

Honestly I don't quite understand what this check on the class's `_p_mtime` is for. Is this maybe vestigial code from the days of persistent classes that we should just remove?

davisagli: No tests in Zope break if I remove it. But yeah, the only test for the Cacheable class is the one I added in this PR...

PR #277, opened 2018-05-17
Add app to product context

rbu:
Products receive the ProductContext as the first parameter
to their initialize() method when they are loaded. This context used to
contain a reference to the app until it was removed during Zope 4
preparations. The Application was, however, never exposed through an
API.

Access to the API is necessary for certain products that automatically
add objects to the Application when they are loaded, such as the
Products.Sessions and Products.TemporaryFolder. This used to happen in
OFS.Application in Zope 2, but was moved out when the products became
indepentent.


src/OFS/tests/applicationproduct/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+    Producs.Sessions and Products.TemporaryFolder are doing.
+"""
+
+def initialize(context):
+    from OFS.Folder import Folder
+    import transaction
+
+    app = context.getApplication()
+    folder = Folder('some_older')
+    app._setObject('some_folder', folder)
```
icemac: Is there a reason why the names are different? (`some_older` vs. `some_folder`)

rbu:
Fixed the typo and added a changelog entry

rbu:
@icemac ping

PR #278, opened 2018-05-17
Fix DeprecationWarning in commit message text

rbu:
Zope/src/OFS/Application.py:213: DeprecationWarning: Expected text
  transaction.get().note(note)


dwt:
LGTM.

PR #280, opened 2018-05-17
WIP: Raise generic errors (catch all InternalError) when using WSGI

tschorr:
Imho it makes sense to propagate errors to the WSGI server in order to make use of the configured logging facilities.


tschorr:
@icemac thanks for clarifying the `HTTPExceptionHandler` will only ever be used for WSGI. This further simplifies the code and should yield higher coverage.

PR #281, opened 2018-05-17
Use native strings for ZODB mount points in Python 2

rbu:
ZODB requires paths and identifiers to be native strings (unicode on
Py3, bytes on Py2). Since the ZConfig parser decodes the config file,
keys and valus in the will consistently be unicode. As a result, the
ZopeDatabase object needs to encode the value back to str/bytes on
Python 2.


rbu:
@davisagli, I assume this got broken for the Python 2 case by this commit in ZConfig/loader.py: https://github.com/zopefoundation/ZConfig/commit/010cb2b2d4d2cdc8cd327f9f46a987024cac5b67

Before this commit, Python 2 and 3 would load all of the config as native strings (because they had different code paths). With this commit, both Pythons load the config as unicode `StringIO`.

PR #290, opened 2018-06-26
Re-add support for XML-RPC

davisagli:
This restores support for XML-RPC when running in Python 3 with the WSGI publisher instead of ZServer. It turns out that we need this to run Plone's browser-based functional tests, because robotframework uses XML-RPC to get the available test keywords.

Mostly this involves moving the `xmlrpc` module back from ZServer.ZPublisher to ZPublisher, along with the corresponding tests. The XML-RPC implementation doesn't depend on using the ZServer, so I think it still belongs in Zope itself.

I fixed a couple Python 3 compatibility issues in `xmlrpc.py`:
- Import from xmlrpc.client instead of xmlrpclib
- Make sure that our override of the code for dumping an arbitrary object instance takes effect in both Python 2 and 3

I also had to work around a Python bug (https://bugs.python.org/issue27777) that was causing an error when processing a request with a text/xml body. The bug was causing the file to hold the body to be opened as text rather than binary, and the workaround is to add a content-disposition header to the request so that cgi.FieldStorage decides to open it as binary. I added a test for this case.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,8 +11,16 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.0b6 (unreleased)
 ------------------
 
+New features
+++++++++++++
+
+- Restore support for XML-RPC when using the WSGI publisher.
+
 - Add a minimum ``buildout.cfg`` suggestion in the docs for creating ``wsgi``
   instances.
+
+Bugfixes
+++++++++
 
 - Fix ZMI upload of `DTMLMethod` and `DTMLDocument` to store the DTML as a
   native ``str`` on both Python versions.
@@ -21,6 +29,10 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 - Inlcude the ``zmi.styles`` repository in this package to break a circular
   dependency.
   (`#307 <https://github.com/zopefoundation/Zope/pull/307>`_)
+
+- Work around Python bug (https://bugs.python.org/issue27777)
+  when reading request bodies not encoded as application/x-www-form-urlencoded
+  or multipart/form-data.
 
 
 4.0b5 (2018-05-18)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 New features
 ++++++++++++
 
-- Restore support for XML-RPC when using the WSGI publisher.
+- Restore support for XML-RPC when using the WSGI publisher - dropped in 4.0a2.
 
 - Add a minimum ``buildout.cfg`` suggestion in the docs for creating ``wsgi``
   instances.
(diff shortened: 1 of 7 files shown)


src/ZPublisher/tests/testHTTPRequest.py
```diff
@@ -766,6 +766,18 @@ def test_processInputs_w_cookie_parsing(self):
...
+        TEST_METHOD_CALL = (
+            b'<?xml version="1.0"?>'
+            b'<methodCall><methodName>test</methodName></methodCall>'
+        )
+        environ = self._makePostEnviron(body=TEST_METHOD_CALL)
+        environ['CONTENT_TYPE'] = 'text/xml'
+        req = self._makeOne(stdin=BytesIO(TEST_METHOD_CALL), environ=environ)
+        req.processInputs()
+        self.assertEqual(req.PATH_INFO, '/test')
+        self.assertEqual(req.args, ())
```
tseaver: Exercise the `qs_on_post` bit?

src/ZPublisher/xmlrpc.py
```diff
@@ -10,15 +10,209 @@
...
+            f = Fault(-3, "Unknown Zope fault type")
+
+        # Do the damage.
+        self.setBody(f)
+        self._real.setStatus(200)
+
+        return tb
+
+
+response = Response
```
tseaver: Is this module a wholesale clone of `ZServer.ZPublisher.xmlrpc`?  Are there any changes?  What should we do about that module?

davisagli: The only changes are the two Python 3 compatibility fixes I mentioned above.

I was also wondering what to do about the copy in ZServer. With these changes as is, it is no longer used. Maybe replace it with a deprecated imports from `ZPublisher.xmlrpc`? But that only works if both ZServer and Zope are updated, and it's hard to enforce that since neither one can have a hard dependency (i.e. with a version spec) on the other. Maybe just replace it with a module that raises an explanation explaining what happened?

src/ZPublisher/tests/testHTTPRequest.py
davisagli: Thanks for asking about this -- helped me realize that I needed a completely different approach (and additional test) to avoid losing the query string.

icemac:
@davisagli What is the status of this PR? Do you think it is ready to merge? Will there be a PR on ZServer to remove the `xmlrpc` package there?

tseaver:
@icemac In addition to the trivial conflict resolution, we still need to address what to do with the almost-duplicate copy left in `ZServer.ZPublisher.xmlrpc`.

icemac:
I think we can remove the code from `ZServer` leaving BBB imports and maybe deprecation warnings (as they were in `Zope` before this PR) and require in `ZServer` a `Zope` version which contains `xmlrpc` again. Maybe I do not understand the problem enough because this seems to easy.

davisagli:
I rebased this and opened a PR to ZServer with deprecated imports.

CHANGES.rst
```diff
@@ -11,9 +11,17 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.0b6 (unreleased)
 ------------------
 
+New features
+++++++++++++
+
+- Restore support for XML-RPC when using the WSGI publisher.
```
icemac: It would be nice to mention the version here which dropped the XML-RPC support.

PR #294, opened 2018-07-19
WIP: get rid of x-wsgi.throw errors

tschorr:
The [specification](https://wsgi.readthedocs.io/en/latest/specifications/throw_errors.html) for this was created in 2006 and is still in "proposed" state.

Most comments on the web-sig mailing list (I've scanned November 2006-February 2007) seem to be against the proposal (I think it's first mentionned [here](https://mail.python.org/pipermail/web-sig/2006-November/002376.html)).

Not sure wether `x-wsgi.throw_errrors` is used anywhere outside of Zope at all?


tschorr:
see also comments in https://github.com/plone/plone.rest/pull/75 and https://github.com/zopefoundation/Zope/pull/280, where `x-wsgi.throw_errors` was removed already.

davisagli:
@tschorr I really don't understand your proposed changes here. It looks like you've started to remove support for rendering exception views altogether, which is an important feature used by Plone, and which is a different change from what you described above.

As for checking x-wsgi.throw_errors... Since Plone renders a nice page for exceptions under normal conditions, it's pretty handy to have a way for tests to tell it not to do that, for cases where we want the test to be able to check for a particular exception. Using x-wsgi.throw_errors is the way we decided to pass the value of the `handleErrors` flag from the test browser. I'm fine if you want to propose a different way to do that, but I am -1 on getting rid of this option entirely.

tschorr:
> I really don't understand your proposed changes here. It looks like you've started to remove support for rendering exception views altogether, which is an important feature used by Plone

That is definitely not what I want to do (code is really WIP).

>  Using x-wsgi.throw_errors is the way we decided to pass the value of the handleErrors flag from the test browser.

What's confusing me is that a 12 year old proposal that was never accepted is used as a justification for introducing into production conditional code that is only relevant for testing ...

icemac:
@tschorr Any news on this PR? (I realised you made a force push to its branch.)

tschorr:
@icemac I finally resumed working on this at Alpine City Sprint. The general idea is to subscribe to IPubFailure and let the subscriber reraise the exception (done in test_WSGIPublisher.py). Next step will be to find a suitable location for the subscriber so that other packages (i.e. zope.testbrowser) can use it. Finally zope.testbrowser (and maybe other packages that currently use the x-wsgi.throw_errors) will need to use the subscriber.

dataflake:
@davisagli @tschorr in #563 I am adding a configuration flag to disable exception view handling. My goal is to allow exception propagation into the WSGI stack to e.g. use WSGI middleware for convenient application debugging.

If you look at the change in `WSGIPublisher.transaction_pubevents`, my change does not shortcut exception handling at the very top of the function like the check for `x-wsgi.throw_errors` does so that conflicts and Unauthorized still get handled correctly by Zope.

Maybe this new flag is sufficient for the other use cases? Then you could just remove the `x-wsgi.throw_errors` check at the top, once my PR is merged, and set the new flag in unit tests that need it.

dataflake:
@davisagli @tschorr #563 is now in master. You can now either call the module-level function `WSGIPublisher.set_default_debug_exceptions` with a true flag or set the flag `debug_exceptions` to a true value on the response object to activate it.

dataflake:
It is obvious that no one really cares, otherwise there would have been at least some response or progress. Anyone who wants to work on this is free to re-open.

PR #295, opened 2018-07-19
add missing single quotation mark

jugmac00:


tseaver:

PR #296, opened 2018-07-19
complete abrupt ending sentence

jugmac00:
The sentence "On the other hand, cataloging a complex object when it changes" ends abrupt and is missing the message.

Imho cataloging a complex object could be too time consuming - which is exactly what I added to the text.

Also, I wrapped the paragraphs at 80 characters, which is was Tres told me a while ago, see
https://github.com/zopefoundation/Zope/pull/261#pullrequestreview-106874625


tseaver:

PR #298, opened 2018-07-19
fix spelling error

jugmac00:
change "Animal type" to "animal type"


tseaver:

PR #302, opened 2018-07-19
fix link to download location of FieldedTextIndex

jugmac00:


tseaver:

PR #303, opened 2018-07-19
fix spelling error

jugmac00:
change "Schlicting" to "Schlichting"


tseaver:

PR #304, opened 2018-07-20
create exception view before handling Unauthorized

tschorr:
as proposed in https://github.com/plone/plone.rest/pull/75#issuecomment-404971502.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,7 @@ Bugfixes
 ++++++++
 
 - Call exception view before triggering _unauthorized.
-  (_`#304 <https://github.com/zopefoundation/Zope/pull/304>`_)
+  (`#304 <https://github.com/zopefoundation/Zope/pull/304>`_)
 
 - Fix ZMI upload of `DTMLMethod` and `DTMLDocument` to store the DTML as a
   native ``str`` on both Python versions.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ New features
 Bugfixes
 ++++++++
 
+- Call exception view before triggering _unauthorized.
+  (_`#304 <https://github.com/zopefoundation/Zope/pull/304>`_)
+
 - Fix ZMI upload of `DTMLMethod` and `DTMLDocument` to store the DTML as a
   native ``str`` on both Python versions.
   (`#265 <https://github.com/zopefoundation/Zope/pull/265>`_)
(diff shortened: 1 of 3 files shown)


lukasgraf:
Just to clarify, I'm not necessarily *proposing* that the order in `WSGIPublisher` needs to be changed - I just noticed that this is the underlying cause of these failures in `plone.rest`. I'm simply not familar enough with the `WSGIPublisher` to say whether changing the order is the right course of action or not.

But I did notice that [this comment](https://github.com/zopefoundation/Zope/blob/cd1b606db271a2fd9a45189be065428e5118b5b9/src/ZPublisher/WSGIPublisher.py#L153-L155)

```python
                 # _unauthorized modifies the response in-place. If this hook
                 # is used, an exception view for Unauthorized has to merge
                 # the state of the response and the exception instance.
```
basically becomes void if you switch the order of triggering the `_unauthorized()` hook and calling the exception view.

This comment and the surrounding logic was introduced in
[#112 (Restore a `_unauthorized` hook on the response object)](https://github.com/zopefoundation/Zope/pull/112) and later updated in
[#121 (Restore exception views for unauthorized)](https://github.com/zopefoundation/Zope/pull/121) by @hannosch - maybe he has some insight into whether this change here would be warranted to accomodate `plone.rest`?

---

I also tried to apply the recommendation

> an exception view for Unauthorized has to merge the state of the response and the exception instance

from that comment to `plone.rest`'s exception view:

The way I interpret it, this would mean that in `plone.rest`'s exception view, we basically would need to undo the possible "damage" done to the response by an `IChallengePlugin`.

For a stock Plone, this sort of works by undoing the [redirect triggered by the `CookieAuthHelper`](https://github.com/zopefoundation/Products.PluggableAuthService/blob/a04e9bc29091b1bbd96360ecb206ce232a35c681/Products/PluggableAuthService/plugins/CookieAuthHelper.py#L186-L231): resetting the (locked) status code of `30x` to that of the exception (Unauthorized, `401`), and removing the `Location` header from the response. But this feels messy, and I can't see how this could be generalized for other possible `IChallengePlugin`s that `plone.rest` can't know about, if they can just modify the response in place.

pbauer:
@davisagli when rebasing with master after https://github.com/zopefoundation/Zope/pull/290 was merged I found the fix for https://bugs.python.org/issue27777 was causing the conflicts but I didi not know which fix was the right one....  Can you do the rebase?

davisagli:
I rebased it @pbauer

pbauer:
Thanks!

PR #307, opened 2018-08-22
Include zmi.styles

icemac:
Currently `zmi.styles` and `Zope` have a circular dependency.
This PR solves this by transplanting the whole `zmi.styles` repository including its history here.


rbu:
@icemac :tada: :balloon: :birthday:

PR #312, opened 2018-09-04
Fix manage_edit of File

### LINKED ISSUE #240: Python 3: File contents not rendered in ZMI
When uploading a `text/plain` file to a `File` object in the ZMI Python 2 and Python 3 behave differently:

* Python 2: The file contents are rendered in the text area and can be edited
* Python 3: The text area renders something like `<File at my-file>` and prevents saving the form with `TypeError('Data can only be bytes or file-like. Unicode objects are expressly forbidden.',)` even when trying to change some other fields.

I think the Python 2 behaviour is way more useful.

icemac:
Fixes #240.

Follow-up: #313.


icemac:
Ready for code review.

PR #314, opened 2018-09-06
SiteAccess: Make VirtualHostMonster support IPv6

lukenowak:
from https://bugs.launchpad.net/zope2/+bug/699865 :

VirtualHostMonster does not work with IPv6 named hosts.

In case of such rewrite configuration:

RewriteRule (.*) http://10.0.243.129:9280/VirtualHostBase/https/[%{SERVER_ADDR}]:4080$1 [L,P]

When SERVER_ADDR is fd00::74ba VirtualHostMonster dies with:

  File "/eggs/Zope2-2.12.14-py2.6-linux-x86_64.egg/ZPublisher/BeforeTraverse.py", line 145, in __call__
    meth(*(container, request, None)[:args])
  File "/eggs/Zope2-2.12.14-py2.6-linux-x86_64.egg/Products/SiteAccess/VirtualHostMonster.py", line 154, in __call__
    host, port = host.split(':')
ValueError: too many values to unpack

This is because IPv6 addresses contain ":" in them.

### DIFF
--- a/src/Products/SiteAccess/VirtualHostMonster.py
+++ b/src/Products/SiteAccess/VirtualHostMonster.py
@@ -14,6 +14,7 @@ from ZPublisher.BeforeTraverse import queryBeforeTraverse
 from ZPublisher.BeforeTraverse import registerBeforeTraverse
 from ZPublisher.BeforeTraverse import unregisterBeforeTraverse
 from ZPublisher.BaseRequest import quote
+from ZPublisher.HTTPRequest import splitport
 from zExceptions import BadRequest
 
 
@@ -147,16 +148,7 @@ class VirtualHostMonster(Persistent, Item, Implicit):
                 stack.pop()
                 protocol = stack.pop()
                 host = stack.pop()
-                if ':' in host:
-                    if host.startswith('['):
-                        # IPv6 address passed
-                        host, port = host.rsplit(':', 1)
-                    else:
-                        # Name or IPv4 address passed
-                        host, port = host.split(':')
-                    request.setServerURL(protocol, host, port)
-                else:
-                    request.setServerURL(protocol, host)
+                request.setServerURL(protocol, *splitport(host))
                 path = list(stack)
 
             # Find and convert VirtualHostRoot directive
--- a/src/Products/SiteAccess/VirtualHostMonster.py
+++ b/src/Products/SiteAccess/VirtualHostMonster.py
@@ -148,7 +148,12 @@ class VirtualHostMonster(Persistent, Item, Implicit):
                 protocol = stack.pop()
                 host = stack.pop()
                 if ':' in host:
-                    host, port = host.split(':')
+                    if host.startswith('['):
+                        # IPv6 address passed
+                        host, port = host.rsplit(':', 1)
+                    else:
+                        # Name or IPv4 address passed
+                        host, port = host.split(':')
                     request.setServerURL(protocol, host, port)
                 else:
                     request.setServerURL(protocol, host)
--- a/src/Products/SiteAccess/tests/testVirtualHostMonster.py
+++ b/src/Products/SiteAccess/tests/testVirtualHostMonster.py
@@ -169,6 +169,12 @@ class VHMPort(unittest.TestCase):
       self.assertEqual(self.app.REQUEST['ACTUAL_URL'],
                            'http://www.mysite.com/folder/')
 
+    def testIPv4Noport(self):
+      ob = self.traverse('/VirtualHostBase/http/www.mysite.com'
+                           '/folder/')
+      self.assertEqual(self.app.REQUEST['ACTUAL_URL'],
+                           'http://www.mysite.com/folder/')
+
     def testPassedPortIPv4(self):
       ob = self.traverse('/VirtualHostBase/http/www.mysite.com:81'
                            '/folder/')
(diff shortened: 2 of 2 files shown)


jmuchemb:
Thanks. But for us, we want to have it in Zope 2.13, right ?

lukenowak:
@jmuchemb shall I use different destination branch?

leorochael:
Hi Luke,

Thanks for reporting the issue and providing the patch.

It seems to me that the code might fail when an IPv6 host is passed, but not a port. Eg:

 - `/VirtualHostBase/http/[::1]/folder/`

The `host.rsplit(':', 1)` looks like it's going to cut inside the IPv6 address in this case.

My take is that the code should do `host.rsplit(':', 1)` regardless of whether an IPv6 address is passed or not.

If we want to make the port mandatory in the VHM URL, then the `if ':' in host` becomes redundant, but then this change can't be made into the 2.13 branch as it's breaking compatibility.

On the other hand, if we keep the port optional, then we must somehow check for the presence of the `:` only in the part of the string after the IPv6 URL. I.e., `if host.startswith('[')` then locate the next ']` and look for the `:` after that.

src/Products/SiteAccess/tests/testVirtualHostMonster.py
```diff
@@ -139,6 +139,48 @@ def test(self, vaddr=vaddr, vr=vr, _vh=_vh, p=p, ubase=ubase):
...
+        self.app.folder.manage_addDTMLMethod('doc', '')
+        self.app.REQUEST.set('PARENTS', [self.app])
+        self.traverse = self.app.REQUEST.traverse
+
+    def tearDown(self):
+        import transaction
+        transaction.abort()
+        self.app._p_jar.close()
+
+    def testPassedPort(self):
```
tseaver: Can you add `IPV4` to the name here, for parity / explicitness?

src/Products/SiteAccess/tests/testVirtualHostMonster.py
```diff
@@ -139,6 +139,48 @@ def test(self, vaddr=vaddr, vr=vr, _vh=_vh, p=p, ubase=ubase):
...
+      ob = self.traverse('/VirtualHostBase/http/www.mysite.com:81'
+                           '/folder/')
+      self.assertEqual(self.app.REQUEST['ACTUAL_URL'],
+                           'http://www.mysite.com:81/folder/')
```
tseaver: We also need a test for IPV4 without a port.

tseaver:
@jmuchemb Doing the work on `master` first makes sense:  we can then backport to `2.13`.

src/Products/SiteAccess/tests/testVirtualHostMonster.py
lukenowak: Done.

src/Products/SiteAccess/tests/testVirtualHostMonster.py
lukenowak: Done.

lukenowak:
@leorochael in 71f0c96b79962670ce96f362b7fa4bd328567534 I hope I covered your comments and those tests are passing. Is that it, or I missed the problem?

src/Products/SiteAccess/tests/testVirtualHostMonster.py
```diff
@@ -139,6 +139,66 @@ def test(self, vaddr=vaddr, vr=vr, _vh=_vh, p=p, ubase=ubase):
...
+
+    def testIPv6(self):
+        ob = self.traverse('/VirtualHostBase/http/[::1]:80'
+                           '/folder/')
+        self.assertEqual(self.app.REQUEST['ACTUAL_URL'],
+                           'http://[::1]/folder/')
+
+    def testIPv6Noport(self):
```
leorochael: nitpick: capitalyze `Port`, as in `testIPv6NoPort()`

src/Products/SiteAccess/VirtualHostMonster.py
```diff
@@ -148,7 +148,12 @@ def __call__(self, client, request, response=None):
                 protocol = stack.pop()
                 host = stack.pop()
                 if ':' in host:
-                    host, port = host.split(':')
+                    if host.startswith('['):
+                        # IPv6 address passed
+                        host, port = host.rsplit(':', 1)
```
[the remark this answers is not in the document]
jmuchemb: You're right about the IPv6-no-port scenario. It's too ugly to keep code that does not split correctly, even if the current implementation of `setServerURL` rectifies things.

However, I find overkill to use a regex.

I suugest:
```py
                host = stack.pop()
                if host.endswith(']'):
                    # IPv6 without port
                    request.setServerURL(protocol, host)
                else:
                    request.setServerURL(protocol, *host.rsplit(':', 1))
```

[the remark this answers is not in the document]
leorochael: Indeed the suggested regex is a little more complex. The intention is to always parse all parts in a single `match()` call, instead of needing one or two extra conditionals.

I don't mind reusing the deprecated `splitport()` if imported via `ZPublisher.HTTPRequest` as sugested by @jerome-nexedi.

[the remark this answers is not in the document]
leorochael: I'm also OK with @jmuchemb's suggestion, BTW.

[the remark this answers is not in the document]
lukenowak: With `splitport` the implementation simplifies a lot, see https://github.com/zopefoundation/Zope/pull/314/commits/e6b0e76b1d276b108c82ba9cfbaadd86774801f5.

jmuchemb:
> @jmuchemb Doing the work on `master` first makes sense: we can then backport to `2.13`.

@icemac Now that's merged, we'd like a new release of 2.13 with this change. However, I'm not used to the procedure for this repository. I anyway don't have rights for pypi.

tseaver:
@jmuchemb We cannot make a 2.13 release from the `master` branch:  it has moved on to 4.x.  We need a PR which backports the changes to the `2.13` branch.

jmuchemb:
> We cannot make a 2.13 release from the master branch

Sure.

But a PR for a `cherry-pick -x` without conflict looks too much for me.

icemac:
@jmuchemb See #314 for the Port to Zope 2.13.

PR #331, opened 2018-10-01
mkwsgiinstance: silence confusing error message

tlotze:
When probing for Zope2 path by using an external interpreter, suppress stderr.

This avoids confusion since previously, an Exception message was printed while
the mkwsgiinstance script actually succeeded.


sallner:
Maybe it would be helpful to include a final message indicating that the generation of the configuration files was successful even in case of an error.

tlotze:
I don't really see the point – the error isn't actually an error after all, it's just the result of a probe which may or may not succeed. As a user, I wouldn't expect the script to start talking about some errors on the way if it did complete the job in a well-defined way.

PR #334, opened 2018-10-02
Support for installing zope via pipenv #272

### LINKED ISSUE #272: Dokument how to use Zope with pipenv
`pipenv` is now officially recommended for application dependency management, see https://packaging.python.org/guides/tool-recommendations/

We already document how to install Zope in a `virtualenv`: http://zope.readthedocs.io/en/latest/INSTALL-virtualenv.html

It would be nice to have documentation how to install it using `pipenv`.

To write this documentation maybe a helpful guide could be https://realpython.com/pipenv-guide/

dhavlik:
This needs a change which requires an additional step if you install via pip and want to use ZServer. We did explicitly not document installation via pipenv with python 2.

This fixes #272.

### DIFF
--- a/docs/INSTALL-virtualenv.rst
+++ b/docs/INSTALL-virtualenv.rst
@@ -67,7 +67,7 @@ Creating a Zope instance
 
   The following steps describe how to install a WSGI based Zope instance.
   If you want/have to use ZServer instead of WSGI (Python 2 only!) follow
-  the documentation `Creating a Zope instance for Zope 2.13`_, but 
+  the documentation `Creating a Zope instance for Zope 2.13`_.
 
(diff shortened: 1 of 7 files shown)


docs/INSTALL-virtualenv.rst
```diff
@@ -58,8 +67,7 @@ Creating a Zope instance
 
   The following steps describe how to install a WSGI based Zope instance.
   If you want/have to use ZServer instead of WSGI (Python 2 only!) follow
-  the documentation `Creating a Zope instance for Zope 2.13`_, as it has not
-  changed since that version.
+  the documentation `Creating a Zope instance for Zope 2.13`_, but
```
icemac: but what?

PR #335, opened 2018-10-02
Add a configuration to run zope including some optional Products.

icemac:


jugmac00:
What is the background/reason for this file? Testing?  I'd like to see a longer comment at the beginning of the file.

icemac:
@jugmac00 I added documentation in fa96f844e83a892e6ac1e82337ef6d52af830407. Does it match your intensions?

jugmac00:
Thank you very much for the explanation. Now it makes sense to me.

Also, there was at least one time during the Saltlabs sprint I wished I had something like that!

Thanks!

PR #336, opened 2018-10-02
Fix handling of broken objects in ZMI

### LINKED ISSUE #213: Broken objects are rendered as Folders in ZMI
If I migrate a Zope 2.13 `Data.fs` to Zope 4 and not all Products are installed the resulting broken objects are rendered as `Folder` in the `Type` column of the ZMI. This way they cannot be distinguished from actual folders.

Is there a way to render with the type `Broken object` to be able to see what's broken? (If I remember correctly previously the icon was showing that the object was broken.)

### LINKED ISSUE #311: broken objects are rendered with the ID `broken` in ZMI
This prevents from deleting them via ZMI because the ID and thus the URLs are wrong.

This seems to be the case since the merge of #249.

See screenshot:
bildschirmfoto 2018-09-03 um 10 53 20

viktordick:
Instead of obtaining the list of objects from `objectValues` and assuming that these provide all necessary attributes (like `getId()`), a list of dictionaries is created on the basis of `objectItems`, which has access to the object's ID as it is known to the parent. This is used for sorting as well as interaction with the specific items, so it allows to show broken objects and also delete them.

fixes #311 
fixes #213

### DIFF
--- a/src/OFS/tests/testSorting.py
+++ b/src/OFS/tests/testSorting.py
@@ -2,7 +2,8 @@
 import codecs
 import Testing.ZopeTestCase
 import Testing.testbrowser
-import Zope2.App 
+import Zope2.App
+
 
 class SortingTests(Testing.ZopeTestCase.FunctionalTestCase):
     """Browser testing ..Image.File"""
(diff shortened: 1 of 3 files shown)


src/OFS/zpt/main.zpt
```diff
@@ -6,12 +6,22 @@
...
+        [('id', item[0]), ('obj', item[1]),]
+        + [
+          (key, getattr(item[1],key)) 
+          for key in ['title', 'meta_type']
+          if hasattr(item[1],key)
+          ]
+        for item in here.objectItems()
+        ];
+      obs python: [dict(item) for item in obs];
+			obs python: sequence.sort(obs, ((skey,sort_func,rkey),), mapping=1 );
```
icemac: PEP-8, please.

src/OFS/tests/testSorting.py
```diff
@@ -0,0 +1,47 @@
...
+        base_url = 'http://localhost/sortingTest/manage_main?skey=%s&rkey=%s'
+        
+        def do_assert(one_before_two):
+            one_before_two_found = (
+                    self.browser.contents.find('File2') > self.browser.contents.find('File1')
+                    )
+            self.assertTrue(one_before_two == one_before_two_found)
+
+        self.browser.open(base_url % ('id', 'asc'))
+        do_assert(one_before_two = True)
```
icemac: PEP-8 does not allow spaces around `=` in function calls.

src/OFS/tests/testSorting.py
```diff
@@ -0,0 +1,47 @@
...
+                b'manager:manager_pass', 'base64').decode()))
+
+    def test_sortby(self):
+        base_url = 'http://localhost/sortingTest/manage_main?skey=%s&rkey=%s'
+        
+        def do_assert(one_before_two):
+            one_before_two_found = (
+                    self.browser.contents.find('File2') > self.browser.contents.find('File1')
+                    )
+            self.assertTrue(one_before_two == one_before_two_found)
```
icemac: What about `assertEqual`?

PR #338, opened 2018-10-02
Use system locale if no locale is set in config

viktordick:
Also, some documentation of possible config settings.

### DIFF
--- a/src/Zope2/Startup/starter.py
+++ b/src/Zope2/Startup/starter.py
@@ -14,6 +14,8 @@
 
 import logging
 import sys
+import locale
+import codecs
 
 from six import PY2
 from ZConfig import ConfigurationError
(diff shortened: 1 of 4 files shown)


src/Zope2/Startup/starter.py
```diff
@@ -51,32 +53,22 @@ def setupInterpreter(self):
             sys.setcheckinterval(self.cfg.python_check_interval)
 
     def setupLocale(self):
-        # set a locale if one has been specified in the config
-        if not self.cfg.locale:
-            return
-
+        '''
```
icemac: I think this method intentionally had no docstring to make it private.

src/Zope2/Startup/wsgischema.xml
```diff
@@ -133,12 +133,9 @@
 
   <key name="locale" datatype="locale" handler="locale">
     <description>
-     Enable locale (internationalization) support by supplying a locale
-     name to be used. See your operating system documentation for locale
-     information specific to your system. If your Python module does not
-     support the locale module, or if the requested locale is not
-     supported by your system, an error will be raised and Zope will not
-     start.
+     locale name to be used. See your operating system documentation for locale
```
icemac: Nitpick: I'd like the sentence to start with an upper letter.

PR #340, opened 2018-10-02
Fix HTTPResponse.setBody when the publisher returns a tuple

davisagli:
When passed a tuple, `bytes()` tries to interpret it as integers representing each byte. That raises a TypeError which we catch here to set the body to what it was in Zope 2.


davisagli:
Oops! I fixed it.

PR #347, opened 2018-10-03
symetric test setup on all devices and appveyor

loechel:


loechel:
@icemac for now, we need a merge of this work, to get the appveyor.yml in master to get recogniced by the tool and checked. I will than make a new pull request if necessary

PR #352, opened 2018-10-04
Update GettingStarted for Zope 4

jugmac00:
- Made it work with Zope 4
- Made it a bit more readable

### DIFF
--- a/docs/zdgbook/GettingStarted.rst
+++ b/docs/zdgbook/GettingStarted.rst
@@ -2,24 +2,35 @@
 Getting Started
 ###############
 
-.. include:: includes/zope2_notice.rst
-
 Introduction
 ============
 
-This chapter cover installation and getting started with development
-of a simple application.  This guide use a build system called
-`Buildout <http://www.buildout.org>`_ to build the application.  And
-the Python packages developed as part of the application can be
-distributed as `Python eggs
-<http://peak.telecommunity.com/DevCenter/setuptools>`_.
+This chapter covers the installation of Zope and getting started with
+the development of a simple application.  This guide uses a build
+system called `Buildout <http://www.buildout.org>`_ to build the
+application.
 
+Prerequisites
+=============
 
-Directory Structure
+Make sure you have Python installed. Version 3.6 or higher is
+recommended.
+
+Creating a `VirtualEnv <https://pypi.org/project/virtualenv/>`_ is
+recommended.
+
+In order to use buildout, you have to install the ``zc.buildout``
+package.
+
+::
+
+  $ pip install zc.buildout
+
+Directory structure
 ===================
 
-To begin the application development, create a directory structure to
-place Python packages and build related files.
+To begin application development, create a directory structure for
+the Python packages and build related files.
 
 ::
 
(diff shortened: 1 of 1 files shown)


docs/zdgbook/GettingStarted.rst
```diff
@@ -2,142 +2,141 @@
...
+After updating the buildout configuration, you can run the `buildout`
+command to build the system.
 
 ::
 
-  [buildout]
-  parts = zope2
-          instance
-  extends = http://download.zope.org/Zope2/index/2.13.20/versions.cfg
+  $ buildout
```
icemac: IMHO this should be `bin/buildout`.

docs/zdgbook/GettingStarted.rst
```diff
@@ -2,142 +2,141 @@
...
-You can see a list of installable applications in the drop-down box.
-Also you can see it in "Control_Panel" -> "Products".
+Use the user name and password you set earlier.
 
 ::
 
-  http://localhost:8080/Control_Panel/Products/manage_main
+  http://localhost:8080/manage
+
+You can see a list of installable applications in the drop-down box
```
icemac: I think this are not applications but objects which can be created in the ZODB.

docs/zdgbook/GettingStarted.rst
```diff
@@ -149,19 +148,18 @@ inside `src` directory.
   $ touch src/poll/main/__init__.py
   $ touch src/poll/main/configure.zcml
 
-The last file we created is a configuration file called Zope
-Configuration Markup Language (ZCML). Soon we will add some boiler
-plate code inside ZCML file.
+The last file is a configuration file. The file ending stands for
+``Zope Configuration Markup Language``.
```
icemac: Maybe keep the abbreviation `ZCML`.

docs/zdgbook/GettingStarted.rst
```diff
@@ -2,142 +2,141 @@
...
+
+Creating a `VirtualEnv <https://pypi.org/project/virtualenv/>`_ is
+recommended.
+
+In order to use buildout, you have to install the ``zc.buildout``
+package.
 
+::
+
+  $ pip install zc.buildout
```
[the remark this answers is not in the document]
jugmac00: I know, there are two ways to work with virtualenvs, either activating them or explicitly name the path. To keep the documentation concise and the paths short, I went the way to activate the virtualenv. Now, I explicitly ask the reader to activate it.

docs/zdgbook/GettingStarted.rst
jugmac00: With the recommendation to activate the virtualenv, buildout is added to the path and so it is fine to just use "buildout" without the bin or possibly venv/bin prefix.

docs/zdgbook/GettingStarted.rst
jugmac00: I think it is always good to know a little bit of the background, so I kept explaining what the abbr stands for.

docs/zdgbook/GettingStarted.rst
jugmac00: Thank you!  "applications" was the wording of the previous author, and I was not entirely sure about the naming convention. Changed it to "objects".

PR #359, opened 2018-10-04
Update Interface/Component documentation for Python 3

jugmac00:


docs/zdgbook/ComponentsAndInterfaces.rst
```diff
@@ -258,8 +252,8 @@ ask an interface the names of all the various interface items it
 describes.  From the Python interpreter, for example, you can walk
 right up to an interface and ask it for its *names*::
 
-  >>> IUser.names()
-  ['getUserName', 'getFavoriteColor', 'getPassword']
+  >>> IHello.names()
+  dict_keys(['hello'])
```
jamadden: That will vary on different versions of Python, won't it? Python 2 still returns a list? Which would result in doctest failures if this file gets run there.

docs/zdgbook/ComponentsAndInterfaces.rst
```diff
@@ -268,39 +262,44 @@ method.
...
+  >>> IHello.namesAndDescriptions()
+dict_items([('hello', <zope.interface.interface.Method object at 0x7fc6875110f0>)])
 
-As you can see, the "description" of the Interface's three items in
-these cases are all `Method` objects.  Description objects can be
-either 'Attribute' or `Method` objects.  Attributes, methods, and
-interface objects implement the following interface::
+.. note::
+  You cannot access the `Method` object by index, as
+  ``namesAndDescriptions`` returns a dict_view.
```
jamadden: But only on Python 3. I wonder if there's an argument to be made that returning lists is actually the most useful behaviour on all versions of Python. Returning views may save memory, but these are small objects; returning views may stay in sync, but these objects don't change.

docs/zdgbook/ComponentsAndInterfaces.rst
```diff
@@ -309,27 +308,24 @@ method parameters.
 Checking Implementation
 =======================
 
-You can ask an interface if a certain class or instance that you hand
+You can ask an interface if a certain class that you hand
```
jamadden: Instances can also implement interfaces, not just provide them, though I think this is rare.

docs/zdgbook/ComponentsAndInterfaces.rst
jugmac00: Thank you for the feedback. I have to admit, I totally forgot about Python 2 compatibility when revising the document. Coincidentially, I talked with @icemac about bringing documentation under doctest, which is not the case in the moment. And I guess it won't in near future.

As Python 3 is the future, I do not think it is absolutely necessary to bloat the document to be fully Python 2 and 3 compatible.

Unfortunately, it is also no good idea to reference the previous version (from 9 years ago), as the code examples had several bugs.

I think it would be a good idea to add a notice at the top of the document, that is was written for Python 3 only.

Also, one could add something like "If you want to use Python 2, please visit https://zopeinterface.readthedocs.io/en/latest/ " or any specific url.

What's your take on it?

docs/zdgbook/ComponentsAndInterfaces.rst
jugmac00: Actually, I removed "or instance" as the document falsely advised to use `IHello.implementedBy(my_hello_instance)`, which resulted in some trouble, cf.
https://bugs.launchpad.net/zope2/+bug/1197508
and
http://blog.fourdigits.nl/how-to-break-your-plone-site-with-implementedby

Generally, I think it is ok to keep this document on a basic level.

Should we reference to https://zopeinterface.readthedocs.io/en/latest/ ?

jugmac00:
@icemac  or @jamadden - could you have a look at the amended pull request? I think, now it is good to go.

Issue #268, opened 2018-04-18
Fix deprecation warnings shown at test run


icemac:
There are many deprecation warnings shown when running the tests, some even in packages Zope depends on. They should be fixed.

- [x] Products.ZCatalog: zopefoundation/Products.ZCatalog/issues/37
- [x] Products.BTreeFolder2: zopefoundation/Products.BTreeFolder2/pull/5
- [x] five/localsitemanager/localsitemanager.txt:1: `DeprecationWarning: zope.site.hooks has moved to zope.component.hooks. Import of zope.site.hooks will become unsupported in Version 5.0`
- [X] Acquisition: https://github.com/zopefoundation/Acquisition/commit/2cb2efaa0006a92ce7df7f256c6476b2e17c4a98
- [x] AccessControl/ZopeGuards.py:46: `DeprecationWarning: the sets module is deprecated`
- [x] RestrictedPython/compile.py:202 `SyntaxWarning: Line 183: Print statement is deprecated and not avaliable anymore in Python 3.` in `AccessControl.tests.testZopeGuards.TestActualPython.testPython`, `AccessControl.tests.testZopeGuards.TestActualPython.test_dict_access` zopefoundation/RestrictedPython#155
- [x] Zope/src/OFS/Application.py:213: `DeprecationWarning: Expected text`  `transaction.get().note(note)`
- [x] ExtensionClass/tests.py:1: `ImportWarning: Not importing directory '[..]/ExtensionClass-4.3.0-py2.7-linux-x86_64.egg/ExtensionClass/pickle': missing __init__.py`
- [x] zopefoundation/five.localsitemanager#9

andbag:
PR created on https://github.com/zopefoundation/Products.ZCatalog/pull/38

dataflake:
I believe errors like the following are unfixable unless you rename the folder inside the affected package so the name clash with the import is fixed. They only occur on Python 2.7. There's several other places, like a folder named `xml` in some packages that import `xml.sax...`.

`ImportWarning: Not importing directory '[..]/ExtensionClass-4.3.0-py2.7-linux-x86_64.egg/ExtensionClass/pickle': missing __init__.py`

icemac:
@dataflake That's fine. So we'll get finally rid of it in Zope 5. That's early enough.

icemac:
There is only the issue in zopefoundation/RestrictedPython#155 remaining, so I am closing this ticket here. Thank you all for working on getting rid of the warnings.

PR #362, opened 2018-10-05
Fix deprecation and resource warnings

tschorr:
s. https://github.com/zopefoundation/Zope/issues/268


src/ZPublisher/tests/test_xmlrpc.py
```diff
@@ -61,7 +61,7 @@ def test_nil(self):
         response = self._makeOne(faux)
         response.setBody(body)
         data, method = xmlrpclib.loads(faux._body)
-        self.assert_(data[0]['public'] is None)
+        self.assertTrue(data[0]['public'] is None)
```
icemac: I'd prefer `self.assertIs()`

PR #364, opened 2018-10-05
Remove bootstrap.py

icemac:
It leads to update loops when running `bin/buildout -n` for the first time.
When omitting `-n` the setuptools version might be too old.


icemac:
@dataflake I tried to install Zope together with @vernans on his new laptop using the installation procedure described in the documentation. To get it working we had to:

* run `python3.6 bootstrap.py` (It installs `setuptools 33.1.1`.)
* run `bin/buildout` but this fails because the recipe `collective.recipe.sphinxbuilder` complains about the wrong `setuptools` version because we pin a newer one in out versions.
* run `bin/buildout -n` to get the current `setuptools` but this call runs into a loop because it tries again and again to update `setuptools`, so we had to stop this buildout process
* run `bin/buildout` again which now succeeds.

This installation story is very unpleasant and even embarrassing to show it to a beginner.
The reason seems to be a bug (?) in recent `zc.buildout` versions which is no longer able to cleanly upgrade `setuptools`.

PR #370, opened 2018-10-11
Install the `ipaddress` package only on Python 2.7.

### LINKED ISSUE #368: Package ipaddress only needed in Python 2
``ipaddress`` is part of Python 3.3+ standard library.
https://github.com/zopefoundation/Zope/blob/6806c482ce1ab87e8828a423ae4bb37616b84c9a/setup.py#L73

It works as-is, but one get then the following at buildout time.

```
/home/workspacejensens/cdev/plone5.2/py3/lib/python3.6/site-packages/zc/buildout/easy_install.py:591: UserWarning: Module ipaddress was already imported from /usr/lib/python3.6/ipaddress.py, but /home/jensens/.buildout/shared-eggs/ipaddress-1.0.22-py3.6.egg is being added to sys.path
```

icemac:
It is part of the stdlib in Python 3.

Fixes #368.


setup.py
```diff
@@ -117,6 +116,11 @@ def _read_file(filename):
         'zope.traversing',
         'zope.viewlet',
     ],
+    extras_require={
+        ':python_version=="2.7"': [
```
sallner: Does it work that way with the colon `:` in front? The [documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies) seems a bit unclear to me.

icemac: It tried it with `buildout` and it worked: It was installed on Python 2.7 but not on Python 3.

icemac: I have taken the approach from https://github.com/inveniosoftware/troubleshooting/issues/1

jamadden: It's also possible to do this directly on a requirement in `install_requires`:

```python
install_requires=[
    'ipaddress ; python_version == "2.7"',
]
```

(That may require a newer setuptools.)

icemac:
Do we have to create separate wheels for Python 2.7 and Python 3 after this PR gets merged?

jensens:
depends on the setuptools version, as found here https://hynek.me/articles/conditional-python-dependencies/

jensens:
Reading more carefully it seems this way of doing it is deprecated and we should use an list entry in `install_requires` likes so: `ipadress ; python_version=="2.7"`.
We should also ensure `setuptools >=36.2` - well, Python packaging... Things are getting better, but its far from a stable environment.

PR #373, opened 2018-10-11
Ensure 'ZOPE2PATH' is text, not bytes

### LINKED ISSUE #372: mkwsgiinstance:  Fails under Py3k on Unix filesystem
In a fresh Zope 4.0b6 install:

```bash
$ bin/mkwsgiinstance -d .
Please choose a username and password for the initial user.
These will be the credentials you use to initially manage
your new Zope instance.

Username: xxx
Password: 
Verify password: 
  File "bin/mkwsgiinstance", line 11, in <module>
    sys.exit(main())
  File "/tmp/zope4/lib/python3.6/site-packages/Zope2/utilities/mkwsgiinstance.py", line 142, in main
    copyzopeskel.copyskel(skelsrc, skeltarget, None, None, **kw)
  File "/tmp/zope4/lib/python3.6/site-packages/Zope2/utilities/copyzopeskel.py", line 182, in copyskel
    copydir(root, dirs, files, targetdir, replacements, uid, gid)
  File "/tmp/zope4/lib/python3.6/site-packages/Zope2/utilities/copyzopeskel.py", line 206, in copydir
    gid)
  File "/tmp/zope4/lib/python3.6/site-packages/Zope2/utilities/copyzopeskel.py", line 225, in copyin
    text = text.replace("<<%s>>" % k, replacements[k])
TypeError: replace() argument 2 must be str, not bytes
```

Debugging, the issue is that `ZOPE2PATH` is bytes.

tseaver:
Closes #372.

### DIFF
--- a/src/Zope2/utilities/mkwsgiinstance.py
+++ b/src/Zope2/utilities/mkwsgiinstance.py
@@ -218,7 +218,7 @@ def get_zope2path(python):
         output = subprocess.check_output(
             [python, '-c', 'import Zope2; print(Zope2.__file__)'],
             stderr=subprocess.PIPE)
-        zope2file = output.strip()
+        zope2file = output.strip().decode('utf-8')
     except subprocess.CalledProcessError:
         # fall back to current Python interpreter
         import Zope2


src/Zope2/utilities/mkwsgiinstance.py
```diff
@@ -218,7 +218,7 @@ def get_zope2path(python):
         output = subprocess.check_output(
             [python, '-c', 'import Zope2; print(Zope2.__file__)'],
             stderr=subprocess.PIPE)
-        zope2file = output.strip()
+        zope2file = output.strip().decode('utf-8')
```
davisagli: I think it should be decoded for Python 3 only; the return value from this function ends up getting used in a str.replace on the content of files opened in text mode.

tseaver: `str.decode` returns `unicode` on Python2.x, which is what we want, I believe.

davisagli: I don't think so. The line that broke in Python 3 in #372 was:

    text = text.replace("<<%s>>" % k, replacements[k])

where `replacements[k]` comes from the `zope2file` value.

`text` is a `str` so we want `zope2file` to also be a `str`, which means not decoding it in Python 2.

icemac:
Superseded by #371.

PR #374, opened 2018-10-12
Fix handling of globals and locals in zconsole scripts.

sallner:
In case `exec` is called without arguments, the current scope is used. This
results in a overfull `globals` missing the `app` which is only in `locals`.

### DIFF
--- a/src/Zope2/utilities/tests/test_zconsole.py
+++ b/src/Zope2/utilities/tests/test_zconsole.py
@@ -21,12 +21,19 @@ instancehome $INSTANCE
 
 test_script = """
 import sys
+import Testing
 
+
+
+def print_info():
+    # This tests the availability of global variables and imports.
+    print(sys.argv[1:])
+    print(Testing.__doc__[:7])  # `Set up`
 
 if __name__ == '__main__':
     app.foo = '42'
     print(app.foo)
-    print(sys.argv[1:])
+    print_info()
 """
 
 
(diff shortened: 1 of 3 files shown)


src/Zope2/utilities/tests/test_zconsole.py
```diff
@@ -21,12 +21,19 @@
...
 test_script = """
 import sys
+import Testing
 
 
+
+def print_info():
+    # This tests the availability of global variables and imports.
+    print(sys.argv[1:])
+    print(Testing.__doc__[:7])  # `Set up`
```
tschorr: what about changing this into something less prone to change, like e.g. `math.pi`

src/Zope2/utilities/zconsole.py
```diff
@@ -15,11 +15,12 @@ def runscript(zopeconf, script_name, *extra_args):
     app.REQUEST['PARENTS'] = [app]
     setRequest(app.REQUEST)
     newSecurityManager(None, user)
-    scriptglobals = globals()
+    scriptglobals = {}
```
tschorr: the next 3 lines could be coded in a single statement

src/Zope2/utilities/tests/test_zconsole.py
icemac: Maybe there is even something inside Zope which is more stable, e. g. the class name of `OFS.PropertyManager.PropertyManager`.

src/Zope2/utilities/zconsole.py
icemac: +1

src/Zope2/utilities/tests/test_zconsole.py
sallner: I actually wanted to import something more complex with namespace etc. instead of standard library as the import of `sys` accidentally worked before, because it was in the scope of the exec call.

src/Zope2/utilities/zconsole.py
sallner: okay, will do that.

PR #392, opened 2018-10-27
Set _p_repr to __repr__ to override Persistent's __repr__

pbauer:
Persistent's `__repr__` tries to delegate to a `_p_repr` if it exists. Setting `_p_repr` works around the issue that in some MROs Persistent's `__repr__` came before the one in `OFS.SimpleItem`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,8 +45,7 @@ Bugfixes
 Other changes
 +++++++++++++
 
-- Restore old ``__repr__`` via ``OFS.SimpleItem.PathReprProvider``. Use this
-  as first base class for your custom classes, to restore the old behaviour.
+- Restore old ``__repr__`` via ``OFS.SimpleItem.PathReprProvider``.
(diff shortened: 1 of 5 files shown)


pbauer:
I had a talk with @icemac and we decied to drop this for now since we seem not to be able to get this to work without seemingly loosing acquisition, resulting in a `__repr__` with a incorrect path. 
See for more details and failing tests https://github.com/plone/Products.CMFPlone/issues/2590#issuecomment-433746872
If someone (@jamadden ?) finds a way to make the delegation to `_p_repr` work correctly we will re-add it later.

jamadden:
Yeah, it's absolutely not clear to me why `_p_repr` wouldn't be working but `__repr__` does. I haven't looked at the ExtensionClass code though. I will do that when I get a chance. It may be as simple as having `Persistent` look up `_p_repr` on the instance instead of the class. Since we're already handling exceptions getting it, that's probably ok.

PR #396, opened 2018-11-04
Enable history page for OFS.Image.File

perrinjerome:
Use the standard history page, enabled using the same condition than the one
that enables the text area on edit (ie. the file must be text or javascript and
reasonably sized)

### DIFF
--- a/src/OFS/tests/testFileAndImage.py
+++ b/src/OFS/tests/testFileAndImage.py
@@ -302,6 +302,16 @@ class FileTests(unittest.TestCase):
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0][1], self.file)
 
+    def testHistoryCompare(self):
+        self.file.manage_edit('a', 'text/plain',
+                              filedata='a')
+        getattr(self.app, self.factory)('b',
+                file='b', content_type='text/plain')
+        self.file.manage_historyCompare(
+            self.file,
+            self.app.b,
+            self.app.REQUEST)
+
     def test_interfaces(self):
         from zope.interface.verify import verifyClass
         from OFS.Image import File
--- a/src/OFS/tests/testFileAndImage.py
+++ b/src/OFS/tests/testFileAndImage.py
@@ -304,13 +304,15 @@ class FileTests(unittest.TestCase):
 
     def testHistoryCompare(self):
         self.file.manage_edit('a', 'text/plain',
-                              filedata='a')
+                              filedata='content_of_a')
         getattr(self.app, self.factory)('b',
-                file='b', content_type='text/plain')
-        self.file.manage_historyCompare(
+                file='content_of_b', content_type='text/plain')
+        page = self.file.manage_historyCompare(
             self.file,
             self.app.b,
             self.app.REQUEST)
+        self.assertTrue('content_of_a' in page)
+        self.assertTrue('content_of_b' in page)
 
     def test_interfaces(self):
         from zope.interface.verify import verifyClass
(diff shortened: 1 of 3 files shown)


perrinjerome:
I remembered I also had this old patch. I understand history support is removed from master ( #346 ), but maybe this sill makes sense for 2.13 branch. If not, then don't hesitate to close this pull request.

There's the same problem that tests are broken for 2.13 because of zc.recipe.egg version conflict. and there's no fix in this branch. Tests for this branch plus an extra commit to fix version conflicts are OK 
https://travis-ci.org/perrinjerome/Zope/jobs/450638496

cc @icemac @leorochael

icemac:
@perrinjerome Thank you for your pull request. Please rebase it to the current version of the 2.13 branch as I fixed TravisCI there in #393.

Additionally please sign a Zope Committer Agreement, see https://github.com/zopefoundation/developer_docs/blob/master/source/Zope_Foundation_Committer_Agreement-20181008.pdf

perrinjerome:
Thanks, I just submitted a signed contributor agreement.

src/OFS/tests/testFileAndImage.py
```diff
@@ -302,6 +302,16 @@ def testFindFile(self):
...
 
+    def testHistoryCompare(self):
+        self.file.manage_edit('a', 'text/plain',
+                              filedata='a')
+        getattr(self.app, self.factory)('b',
+                file='b', content_type='text/plain')
+        self.file.manage_historyCompare(
+            self.file,
+            self.app.b,
+            self.app.REQUEST)
```
icemac: The test should `assert` something in the return value of the method.

src/OFS/tests/testFileAndImage.py
```diff
@@ -302,6 +302,18 @@ def testFindFile(self):
...
+    def testHistoryCompare(self):
+        self.file.manage_edit('a', 'text/plain',
+                              filedata='content_of_a')
+        getattr(self.app, self.factory)('b',
+                file='content_of_b', content_type='text/plain')
+        page = self.file.manage_historyCompare(
+            self.file,
+            self.app.b,
+            self.app.REQUEST)
+        self.assertTrue('content_of_a' in page)
```
icemac:
```suggestion
        self.assertIn('content_of_a', page)
```

This will result in a nicer error message if the assertion fails.

src/OFS/tests/testFileAndImage.py
```diff
@@ -302,6 +302,18 @@ def testFindFile(self):
...
+        self.assertTrue('content_of_b' in page)
```
icemac:
```suggestion
        self.assertIn('content_of_b', page)
```

src/OFS/tests/testFileAndImage.py
perrinjerome: Thanks, this is also what I did first, but travis-ci also runs on python 2.6, where `assertIn` was not yet available. ( For the records, error on my branch when I tried `assertIn` was https://travis-ci.org/perrinjerome/Zope/jobs/456256346 )

I see that on [setup.py](https://github.com/zopefoundation/Zope/blob/2.13/setup.py#L43) python 2.6 is listed, so if I understand correctly it's not only travis-ci and we have to stick this old style assertion.

So we leave it like this ?

icemac: Sorry, you are absolutely right. I always forget that Zope 2.13 still supports Python 2.6. So please keep your code as it is.

perrinjerome:
Thanks for feedback,I'm merging this !

perrinjerome:
This change was only on 2.13 and not on Zope 5, I pushed a slightly updated version as https://github.com/zopefoundation/Zope/pull/1245

PR #398, opened 2018-11-14
Add security declarations to SimpleItem's FTP methods.

icemac:
This no longer requires child classes to do so.
See zopefoundation/Products.CMFCore#53.

### DIFF
--- a/src/OFS/Image.py
+++ b/src/OFS/Image.py
@@ -707,9 +707,6 @@ class File(
             RESPONSE.setStatus(204)
             return RESPONSE
 
-        security.declareProtected(ftp_access, 'manage_FTPstat')  # NOQA: D001
-        security.declareProtected(ftp_access, 'manage_FTPlist')  # NOQA: D001
-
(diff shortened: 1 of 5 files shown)


src/OFS/Image.py
```diff
@@ -878,12 +875,6 @@ class Image(File):
     security.declareProtected(View, 'get_size')  # NOQA: D001
     security.declareProtected(View, 'getContentType')  # NOQA: D001
 
-    if bbb.HAS_ZSERVER:
-        security.declareProtected(change_images_and_files, 'PUT')  # NOQA: D001
```
davisagli: This one should maybe remain?

icemac: It is already declared in the base class `File`. See https://github.com/zopefoundation/Zope/blob/06e2b82d7ee6421baea95853158708a2f6ed910a/src/OFS/Image.py#L692-L694

icemac:
Thank you.

PR #400, opened 2018-11-15
Add changelog entry for #314

perrinjerome:
We did not add a changelog entry on #314

Also, it's not clear to me if this qualifies as Bug fix or new feature.


icemac:
I think it is a feature because the VirtualHostMonster is older than IPv6. The change log entry should include a link to #314.

perrinjerome:
Thanks, you're right it's a new feature.

I updated to include a link to #314 and also changed the wording a bit. Let me know if that's OK.

PR #402, opened 2018-11-20
Fix str/bytes confusion on add DTMLDocument/DTMLMethod

rbu:
DTML documents need to be stored as native strings ("str" in both Py2 and Py3),
which was implemented in #265 for the manage_upload case (create a document,
then upload the content).

This commit introduces the conditional type conversion for the "add" case
as well (create a document with initial uploaded content).

### DIFF
--- a/src/OFS/DTMLDocument.py
+++ b/src/OFS/DTMLDocument.py
@@ -136,6 +136,10 @@ def addDTMLDocument(self, id, title='', file='', REQUEST=None, submit=None):
     """
     if hasattr(file, 'read'):
         file = file.read()
+    if PY3 and isinstance(file, binary_type):
+        file = file.decode('utf-8')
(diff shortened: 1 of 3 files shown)


rbu:
@icemac @dwt I was thinking the whole
```py
    if hasattr(file, 'read'):
        file = file.read()
    if PY3 and isinstance(file, binary_type):
        file = file.decode('utf-8')
    if PY2 and isinstance(file, text_type):
        file = file.encode('utf-8')
```
triad could go into a helper function like `def _ensure_native_string(file_or_text_or_bytes)`, as it's used three times now in DTMLMethod/DTMLDocument. Thoughts?

dataflake:
I just pulled it into one function that's now used by ``addDTMLDocument``, ``addDTMLMethod`` and ``manage_upload``.

PR #405, opened 2018-11-22
Update expire date on expireCookie function

gabrielgisoldo:
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie , the date used on expires should be in the same format as Date on HTTP Header.
As of Firefox 60.3.0esr, passing the date in any other format doesn't work and is ignored by the browser.

### DIFF
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -331,7 +331,7 @@ class HTTPBaseResponse(BaseResponse):
         if 'value' in d:
             d.pop('value')
         d['max_age'] = 0
-        d['expires'] = 'Wed, 31-Dec-97 23:59:59 GMT'
+        d['expires'] = 'Wed, 31 Dec 1997 23:59:59 GMT'
 
         self.setCookie(name, value='deleted', **d)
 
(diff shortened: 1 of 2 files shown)


dataflake:
Hi Gabriel, did you sign a committer agreement (see https://docs.zope.org/developer/becoming-a-committer.html)? Without it, we cannot accept any contribution.

Secondly, the checking breaks all current tests for that date format. Fixing those is a required part of a contribution like this.

gabrielgisoldo:
Hi Jens, I just sent the committer agreement and corrected the tests to use the new date format.
Can you take a look, please?

dataflake:
I'm trying to verify the contributor agreement with those who handle them. I am not involved in that so I have to wait to hear back from them.

dataflake:
@gabrielgisoldo I just heard from Jim Fulton who gets the agreements at the official address contributor-admin@zope.org, and he says he has not received anything yet.

gabrielgisoldo:
@dataflake, I sent it at 12.:41 utc-2... 
Anyway, just sent it again. 

screenshot_20181123-204645_gmail

dataflake:
@tseaver confirmed that the contributor agreement has been approved.

I checked several documentation resources about that date format for the Expires header. While I couldn't find any information about when it changed or why, I did notice that old documentation (pre-2010 roughly) still shows the old format we use in Zope, but anything after that uses the space-separated format. So with the tests passing this seems safe to merge.

dataflake:
P.S.: I've done the merge

PR #407, opened 2018-11-29
SECURITY.rst mistake

dek4nice:


docs/SECURITY.rst
```diff
@@ -1,4 +1,4 @@
-Filesytem Permissions
+Filesystem Permissions
 =====================
```
ale-rt: Thanks a lot!
You should also add a "=" in line 2 to match the length of the sentence in line 1 :)

ale-rt:
Thanks again, @icemac I do not know if @dek4nice has signed the contributor agreement and if it is required for such a small change.

docs/SECURITY.rst
dek4nice: two commits for two bytes.. overkill :(

icemac: No problem, we'll squash merge them into one. 😆

icemac:
@dek4nice Thank you for your pull request. I do not think that it is necessary to sign a contributor agreement for changing those two bytes.

tseaver:
Typo fixes definitely do not meet the bar for "copyrightability."

PR #412, opened 2018-12-10
Add back getZopeVersion()

### LINKED ISSUE #411: Reliable detecting the Zope version
Add-ons interacting with the ZMI need to provide different templates for Zope 2 (old style ZMI) and Zope 4 (bootstrap based ZMI). Therefore, there is a need to reliable distinguish those versions -- potentially via a mechanism analogous to Python's `sys.version_info`.

d-maurer:
Fixes #411.


src/App/version_txt.py
```diff
@@ -10,22 +10,54 @@
...
     if _version_string is None:
         v = sys.version_info
         pyver = "python %d.%d.%d, %s" % (v[0], v[1], v[2], sys.platform)
         dist = pkg_resources.get_distribution('Zope')
         _version_string = "%s, %s" % (dist.version, pyver)
 
+        expr = re.compile(
+            r'(?P<major>[0-9]+)\.(?P<minor>[0-9]+)(\.(?P<micro>[0-9]+))?'
+            '(?P<status>[A-Za-z]+)?(?P<release>[0-9]+)?')
+        dict = expr.match(dist.version).groupdict()
```
ale-rt: Can we avoid the use of the name dict for a variable?

d-maurer:
Alessandro Pisa wrote at 2018-12-14 00:13 -0800:
>ale-rt requested changes on this pull request.
>
>
>
>>      if _version_string is None:
>         v = sys.version_info
>         pyver = "python %d.%d.%d, %s" % (v[0], v[1], v[2], sys.platform)
>         dist = pkg_resources.get_distribution('Zope')
>         _version_string = "%s, %s" % (dist.version, pyver)
>
>+        expr = re.compile(
>+            r'(?P<major>[0-9]+)\.(?P<minor>[0-9]+)(\.(?P<micro>[0-9]+))?'
>+            '(?P<status>[A-Za-z]+)?(?P<release>[0-9]+)?')
>+        dict = expr.match(dist.version).groupdict()
>
>Can we avoid the use of the name dict for a variable?

Would "dict --> vd" (for "version_dict") please you?

icemac:
@d-maurer wrote:
> What should such a test test?

It should at least assure that the code does not break a bit like the way you tested it manually. Additionally it could assert that at least the major version is correct.

> Side note: "libraries.io" seems to withhold the "sem ver" (= "semantic versioning") point if there are distribution versions not following the format "<major>.<minor>.<micro>..." Maybe, the final release should have a "micro"?

I personally do not like `micro` version numbers which are `0`, so I like to omit them.
I looked at https://libraries.io/pypi/Zope but I did not see any problems there. What kind of problems did you encounter there?

d-maurer:
Michael Howitz wrote at 2018-12-16 23:15 -0800:
>@d-maurer wrote:
> ...
>> Side note: "libraries.io" seems to withhold the "sem ver" (= "semantic versioning") point if there are distribution versions not following the format "<major>.<minor>.<micro>..." Maybe, the final release should have a "micro"?
>
>I personally do not like `micro` version numbers which are `0`, so I like to omit them.
>I looked at https://libraries.io/pypi/Zope but I did not see any problems there. What kind of problems did you encounter there?

Follow the link on the number for `SourceRank`
(this leads you to "https://libraries.io/pypi/Zope/sourcerank").
It explains why `Zope` has only source rank 6.

One point is associated with "Follows SemVer" where
"SemVer" stands for "SEMantic VERsioning" and likely means
that version numbers have the form "major", "minor", "micro"
with "major" indicating major incompatibilities; "minor" indicating
significant but backward compatible changes; and "micro" indicating
minor changes such as bug fixes.

I have the impression that the "pypi" integration into
"libraries.io" is bad (or at least poorly documented).
E.g., it apparently does not see the
source repository nor the contributors for `Zope`.
The `Zope` community might therefore decide to ignore the ranking
on "libraries.io".


--
Dieter

src/App/version_txt.py
d-maurer: `dict` renamed to `version_dict`.
Added `getZopeVersion` related entry to `CHANGES.rst`.
Added tests for `getZopeVersion`: Note that 2 unrelated tests from `test_ApplicationManager` failed with `AttributeError: module 'Testing' has no attribute 'testbrowser'`(maybe due to an outdated `Testing` module).

CHANGES.rst
```diff
@@ -11,7 +11,11 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.0b9 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Fixes
++++++
+
+- Recreate ``App.version_txt.getZopeVersion``
+  (`#411 <https://github.com/zopefoundation/Zope/issues/411>` _)
```
```suggestion
  (`#411 <https://github.com/zopefoundation/Zope/issues/411>`_)
```

I believe this is needed to fix the ReST syntax.

src/App/tests/test_getZopeVersion.py
```diff
@@ -0,0 +1,37 @@
...
+            getZopeVersion().major,
+            int(get_distribution("Zope").version.split(".")[0])
+            )
+
+    def test_types(self):
+        zv = getZopeVersion()
+        for i in (0, 1, 2, 4):
+            self.assert_(isinstance(zv[i], int), str(i))
+        self.assert_(isinstance(zv[3], str), '3')
+
```
icemac: I think there is no need for these empty lines.

src/App/tests/test_getZopeVersion.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_major(self):
+        self.assertEqual(
+            getZopeVersion().major,
+            int(get_distribution("Zope").version.split(".")[0])
+            )
+
+    def test_types(self):
+        zv = getZopeVersion()
+        for i in (0, 1, 2, 4):
+            self.assert_(isinstance(zv[i], int), str(i))
```
icemac: `self.assert_` got deprecated in Python 3.2. I'd suggest to use `self.assertIsInstance` for this assertion.

src/App/version_txt.py
```diff
@@ -10,22 +10,54 @@
...
+import collections
+import re
 import sys
 
 import pkg_resources
 
 _version_string = None
+_zope_version = None
+
+ZopeVersion = collections.namedtuple(
```
icemac: I like the usage of `namedtuple`.

d-maurer:
I fear I need help: I have addressed all change requests and resolved all conversations - but the pull requests remains blocked because some requested change seems still not addressed.

Issue #413, opened 2018-12-10
Retry on ConflictError is broken


icemac:
Retrying of requests in case of an `ConflictError` is broken on multiple levels:

- [x] [HTTPRequest.retry_max_count = 0](https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/HTTPRequest.py#L177) meaning no retries by default. This default is not not changed when setting `max_conflict_retries` in zope.conf See [1] below.
- [ ] If an exception view for `Exception` is reqistered it has to reraise `ConflictException` otherwise it will not be handled by the retry mechanism. See [2] below.
- [x] If a request is actually retried the newly created event gets closed before processing it. See [3] below and #414.
- [x]  `ConflictError` is a subclass of `TransientError`, the code using both can be simplified. See [4] below.
- [x] ~~[HTTPRequest.supports_retry()](https://github.com/zopefoundation/Zope/blob/4b58ee3ba5e048aa20893e7f245086038e5f25bb/src/ZPublisher/HTTPRequest.py#L179) sleeps when called. See [5] below.~~ Moved to #474.

[1] The default used to be `3` in Zope 2.13 but was changed in 740eb601fb9660246e84fe6a58c37a18202e3ef2 to `0`. In 40e5f996ee1dced9bfc821dd7064767411c2bf18 the ability to change the default was added but it sets the value on `ZPublisher.HTTPRequest` (the module) instead of `ZPublisher.HTTPRequest.HTTPRequest` (the class), so it does not have the expected effect.

[2] This seems unexpected to me. Maybe ConflictError should be handled separately like it is already done for `Unauthorized`.

[3] The `request.close()` in the finally block (see [WSGIPublisher.publish_module](https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/WSGIPublisher.py#L279)) closes the newly created request, so it is empty during the retry.

[4] Since https://github.com/zopefoundation/ZODB/commit/364b2a21a1d77e539773001c8a44e6c1df6672d5 a ZODB `ConflictError` is a subclass of transaction's `TransientError`. The commit is part of ZODB 3.10+.

[5] I'd not expect that a method called `supports_retry` does the actual sleeping. It is called three times when a `ConflictError` happens.

dataflake:
ba72a30cf6332fc060e38e3e1849be407180b75a fixes the configuration setting, 40e5f99 was clearly faulty.

dataflake:
[...]

PR #414, opened 2018-12-10
Retry: fix closing newly created request before processing

icemac:
See #413, item [3].

Because we do not have a special handling for `ConflictError` in the exception view code, the exception view for `IException` has to be unregistered as it otherwise prevents the `ConflictError` to propagate to the except block which does the retry handling.


icemac:
This code is running in production since the day after the PR was created so I am going to merge it to include it in the next Zope release.

PR #439, opened 2018-12-23
Fix typo.

jugmac00:


jugmac00:
Section has been revised completely now with pull request #440 - no need for this one any more.

PR #440, opened 2018-12-23
[On hold] revise Zope developers guide -  object publishing

jugmac00:
Over the next couple of days I will review and heavily edit the above mentioned chapter of the ZDG.

I plan to:
- check for Python 3 compatibility
- Zope 4 related changes
- remove references to the now deprecated ZSserver
- ...

As this is a very huge document, I will edit one section at a time.

I am looking forward to receiving feedback early!


jugmac00:
This pull request gets huge - I think it's no fun to review it once it's finished.

What do you think about splitting it into one pull request for each section?

Also, I noticed, the section and sub-section splitting in chapter 4 is pretty messy. Ie "testing" is certainly no sub-section of "environment variables" section.

That would mean roughly 30 sections and thus 30 pull requests. What do you think?

jugmac00:
Almost half way through, I came to the conclusion that it is not a good idea to revise the whole chapter within one pull request - as that one would be massive. The documentation has not been touched much in the last decade, which means there is a lot to change
- whitespace issues
- typos
- fix code examples
- out of date information

That's why I decided to close pull request, and split it into several smaller ones.

PR #458, opened 2019-01-21
Fix confusing and broken button to show/hide the left-hand tree pane

### LINKED ISSUE #457: Please remove toplevel item "Show hide tree menu..."
At the top of the new ZMI there's a symbol for showing/hiding the left hand tree navigation pane. This would be useful if it did what it promised, but that only works in the Zope root. In other places you either lose the left-hand tree pane and cannot get it back at all (non-folderish objects) or in a subfolder you get back a tree pane rooted at that folder, and you can never get back to the full tree view.

This button is not just failing to do what it promises, it basically ruins the UI in unexpected ways. Please get rid of it.

dataflake:
fixes #457

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,11 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.0b9 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Fixes
++++++
+
+- Remove confusing and broken button to show/hide the left-hand tree pane
+  (`#457 <https://github.com/zopefoundation/Zope/issues/457>`_)
 
 
 4.0b8 (2018-12-14)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 Fixes
 +++++
 
-- Remove confusing and broken button to show/hide the left-hand tree pane
+- Improved showing/hiding of the left-hand tree pane
   (`#457 <https://github.com/zopefoundation/Zope/issues/457>`_)
 
 - Don't error out when showing permissions for a non-existent user
--- a/src/App/dtml/manage.dtml
+++ b/src/App/dtml/manage.dtml
@@ -1,24 +1,33 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Frameset//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-frameset.dtd">
-<html>
+<!DOCTYPE html>
 <head>
-<title>Zope on &dtml-BASE0;</title>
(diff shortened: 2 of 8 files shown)


drfho:
Hi, without that button the user does not get aware that there is a (sometimes very helpful) sitemap. The button behaves like URL-toggeling betwenn _manage/manage_main_ in Zope2 .  If you remove the button the user has to enter the URL-Extension manually into the browser - actually with the same result. @dataflake: what behaviour does the button promise?

drfho:
goal here was to replace the URL entering by a button click.  Your proposal now is to reconstruct always the last state of the sitemap ("tree"). The tree-tag might allow that by using the parameter tree-state (cookie tree-s) and a new cookie for the last shown root container folder (e.g. tree-f) . According to
https://docs.zope.org/zope2/zope2book/AppendixA.html#tree-inserts-a-tree-widget
the tree might be assembled by the two parameters dericed form cookies and set here:
https://github.com/zopefoundation/Zope/blob/master/src/App/dtml/menu.dtml#L50
In this case the user will always have the same root in the sitemap - and this will be the one initially requested. Actually this might be more consistent, but I think the former behaviour is useful, too; so this should be executed if the manage-URL is directly requested (an not via sitemap-button). So we should revise the template in that sense. 
If you consider the UX of the "confusing" button worse than without a button for showing the sitemap, please hide it until we deliver the better solution.

dataflake:
@drfho I'm wondering why the goal is to replace the entire URL. Why not just show/hide the navigation tree or set that frame's with to zero using Javascript? That way you wouldn't have the unnecessary page reload and you wouldn't have to worry about the tree state since that just stays the way it is, just hidden.

drfho:
@dataflake: okay, zero-frame could be done like this
https://stackoverflow.com/questions/24030228/completely-hide-frame-in-framesets-using-javascript-or-jquery 
But this alone would guarantee us more wrong URLs in our customer support tickets, because after login the users URL will not change anymore. To handle this 1. we could add another button "Copy URL" or 2. we try to synchronize the main URL like `window.parent.history.replaceState('','Main','/folder1/folder2')`. Actually I would prefer the implicit option 2.

dataflake:
@drfho Ah, now I see where you're coming from and why changing the URL is a goal. I was wondering about that.

I personally have no issue with a changing URL, all I want is when I hide the tree and then show it again that the tree looks just like before. How that can be accomplished isn't important to me.

drfho:
@dataflake My code proposal now toggles the left frame via col width, can remember the last (resized) frame-width per localstorage and shows the main-frame URL in the browser. This replaces former simple show/hide function:
https://github.com/zopefoundation/Zope/pull/458/commits/5e15cf60c0be3b93b11683380e61a3e9317a15c3

dataflake:
@icemac Let me know what you think and merge if you like

icemac:
Thank you for fixing this issue.

PR #460, opened 2019-01-22
Removed support for management_page_charset

### LINKED ISSUE #313: Remove support for `management_page_charset`
It does not seem to be a good idea to define different values for `management_page_charset` on different folders in a zope application. This is especially true on Python 3 where [processInputs()](https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/HTTPRequest.py#L508) converts the inputs to text. The default encoding used is `UTF-8`. There is no way to access the `management_page_charset` when `processInputs()` is called because traversal is done afterwards.

The usage of `management_page_charset` should be replaced by `default-zpublisher-encoding` set in the config file.

Any opinions?

dataflake:
fixes #313

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,8 +20,8 @@ Fixes
 - Don't error out when showing permissions for a non-existent user
   (`#437 <https://github.com/zopefoundation/Zope/issues/437>`_)
 
-Other changes
-+++++++++++++
+Breaking changes
+++++++++++++++++
 - Removed support for ``management_page_charset``
  (`#313 <https://github.com/zopefoundation/Zope/issues/313>`_)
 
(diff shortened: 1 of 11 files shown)


CHANGES.rst
```diff
@@ -20,6 +20,11 @@ Fixes
 - Don't error out when showing permissions for a non-existent user
   (`#437 <https://github.com/zopefoundation/Zope/issues/437>`_)
 
+Other changes
```
icemac: I think this is a breaking change as we remove a feature.

dataflake: Re-characterized as "Breaking change" in the change log now

dataflake: And changed WHATSNEW.rst to reflect the removal.

icemac:
GitHub does not seem to like if the last commit of a PR is skipped from the CI run. It insists that the CI tests have to be run before it can be merged.
I am going to use my administrator rights to merge this PR.

PR #461, opened 2019-01-22
- Document filesystem caching for Chameleon page templates

### LINKED ISSUE #291: Activate Chameleon caching on hard disk by default
Chameleon needs an environment variable pointing to a directory, see https://chameleon.readthedocs.io/en/latest/configuration.html#general-usage

Caching on hard disk speeds up using Chameleon PageTemplate a lot, even in tests.

~~At least it should be documented what to do when using Zope in an application.~~ Done in #461.

dataflake:
fixes #291

### DIFF
--- a/docs/zope2book/AdvZPT.rst
+++ b/docs/zope2book/AdvZPT.rst
@@ -1202,6 +1202,27 @@ caching for performance-critical applications.
 For more information on caching in the context of Zope, see the
 chapter entitled `Zope Services <ZopeServices.html>`_.
 
+Filesystem caching for Chameleon-based templates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Zope 4 introduced the `Chameleon HTML/XML template engine
+<https://chameleon.readthedocs.io/>`_ as new backend for Zope Page
+Templates. The Chameleon templating engine can compile templates and
+cache them on the file system for faster startup and execution.
+
+File system caching is activated by setting an environment variable
+named ``CHAMELEON_CACHE`` to the path of a folder on the filesystem
+where Chameleon can write its compiled template representation.
+
+Look for or add a section named ``environment`` in ``etc/wsgi.conf``
+and add a suitable filesystem path, for example::
+
+  <environment>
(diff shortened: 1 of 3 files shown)


docs/zope2book/AdvZPT.rst
```diff
@@ -1202,6 +1202,27 @@ caching for performance-critical applications.
...
+cache them on the file system for faster startup and execution.
+
+File system caching is activated by setting an environment variable
+named ``CHAMELEON_CACHE`` to the path of a folder on the filesystem
+where Chameleon can write its compiled template representation.
+
+Look for or add a section named ``environment`` in ``etc/wsgi.conf``
+and add a suitable filesystem path, for example::
+
+  <environment>
```
icemac: Cool, in our projects we always put this variable directly into `os.environ`, because at least in Zope 2 `environment` was loaded to late to have any effect for the chameleon configuration.

docs/zope2book/AdvZPT.rst
```diff
@@ -1202,6 +1202,27 @@ caching for performance-critical applications.
...
+    CHAMELEON_CACHE $INSTANCE/var
```
icemac: Hm, this will pollute `/var` with many, maybe hundreds of files. I'd prefer to suggest to use a separate directory. I know `chameleon` requires this directory to exist beforehand, maybe this could be documented, too.

docs/zope2book/AdvZPT.rst
dataflake: I just did some light testing and saw those compiled files show up when using the `<environment>` section, so I can tell it works.

docs/zope2book/AdvZPT.rst
dataflake: @icemac I have changed the path to `$INSTANCE/var/cache` and added a note about creating that folder before starting Zope to the Zope Book chapter. Let me know what you think.

docs/zope2book/AdvZPT.rst
icemac: I also did that kind of testing to convince myself that it works now.

docs/zope2book/AdvZPT.rst
icemac: I like it.

PR #462, opened 2019-01-25
Remove support for Bobo Call Interface headers

Rotonen:
I'm testing the waters here. This should probably not get merged before some reasonable time to object has passed so people have time to speak up.

As the headers disclose unnecessary information and my search engine fu on what would still rely on them leads to only the implementation and a discussion from 2003 on whether to drop them or not, I'm thus betting BCI to be dead and unused at this point.

I'm also open to making this configurable, if this is still actually used for something.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,9 @@ Other changes
 Breaking changes
 ++++++++++++++++
 
+- Removed support for Bobo Call Interface
+ (`#462 <https://github.com/zopefoundation/Zope/pull/462>`_)
+
 - Removed support for ``management_page_charset``
  (`#313 <https://github.com/zopefoundation/Zope/issues/313>`_)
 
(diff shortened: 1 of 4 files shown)


Rotonen:
And for reference, I've pinged some of the original participants on the Plone community forums:
https://community.plone.org/t/bobo-exceptions/7876

And here is the thread:
https://zope-dev.zope.narkive.com/EFReCUD7/bobo-exception-can-produce-rfc-non-compliant-headers-in-response

ale-rt:
Did not even know this thing existed.
After reading the 2003 thread I am +1 for merging but I will leave to more experienced devs the honor to review.

CHANGES.rst
```diff
@@ -35,6 +35,8 @@ Other changes
 Breaking changes
 ++++++++++++++++
 
+- Removed support for Bobo Call Interface
+
```
icemac: I would like to have a reference to the PR here.

```suggestion
(`#462 <https://github.com/zopefoundation/Zope/pull/462>`_)

```

Rotonen:
Amended the changelog and rebased onto master.

PR #466, opened 2019-02-03
Fix display of ZMI breadcrumbs with non-ASCII path elements

### LINKED ISSUE #381: DTML Document breaks if non-ASCII chars in URL
If the `id` of a `DTML Document` contains a non-ASCII char (allowed since #181) it breaks while rendering in the ZMI.
Only Python 2 is affected.

### LINKED ISSUE #401: Python 2: Encoding problem in breadcrumbs
When using non-ASCII characters in the id of an object the id is not rendered correctly in the breadcrumbs in the ZMI. This only happens on Python 2.7, Python 3 behaves correctly.

Example:
bildschirmfoto 2018-11-19 um 08 18 57

dataflake:
Fixes #401 
Fixes #381

Not elegant because it requires special-casing Python 2, but it works...


dataflake:
It have a suspicion that those two failing tests in `OFS/tests/testFileAndImage.py` are bad, because when I try to understand what they are testing by manually entering non-ASCII text into a File object's text area it does not work correctly under Python 2:

screenshot 1

I'll try to follow that lead first by changing `OFS.Image.File` code until the ZMI shows correct values.

dataflake:
I'm banging my head against this. Visiting the ZMI with my proposed changes fixes the breadcrumbs and they are fine on both Python 2 and Python 3. At the same time, manual testing of what the file edit tests introduced in 3c56da927fdb84193856a9492023f973fd96237a try to test (edit a `OFS.File` that has a text content type and enter a non-ASCII value in the textarea) show the same results with my breadcrumbs fix applied and without it. Yet these unit tests for the file edit fail with my breadcrumbs fix applied.

There's something bogus with these tests. Clearly they are not testing correctly, anyway, as the visual representation in the ZMI proves. Under Python 2, with or without the breadcrumbs fix applied. you get double-encoding, so there's a bug these tests are not catching.

dataflake:
Looking at one of the tests something strikes me as bad. The test file is marked with `# -*- coding: utf-8 -*-` at the top and then contains text representations that seem to rely on the editor to do the right thing.`'hällo'` is assumed to be a binary string encoded in UTF-8. But what is `u'hällo'` then? There's a bit too much reliance on the text edit in there. It seems a lot safer to me being very explicit and using something like `u'h\xe4llo'` as the starting point and then applying encoding to that explicit unicode string.

``` python
    def test_Image__manage_main__1(self):
        """It shows the content of text files as text."""
        self.app.file.update_data(u'hällo'.encode('utf-8'))
        self.browser.open('http://localhost/file/manage_main')
        text = self.browser.getControl(name='filedata:text').value
        self.assertEqual(text, 'hällo')
```

dataflake:
I changed tack after feeling I had fallen into a rathole where every small change just seemed to break stuff in other places.

This much simpler fix came about with simple trial and error. I just started deleting code from the DTML that is part of (or pulled into) the file edit form until I found what broke visual rendering for all of it: 

`App/dtml/manage_navbar.dtml` pulled in user information with `<dtml-var AUTHENTICATED_USER>`. Turns out the rendered value of it is straight unicode. Pulling it into the enclosing DTML method messed up all elements that render to UTF-8 output by themselves. It seems a unicode value in one place of a greater DTML document will trigger a forced encoding of the complete rendered output, which then led to the double encoding in both the breadcrumbs, the file edit field and other subsequent places where the non-ascii object ID was shown.

The current fix passes all unit tests (although I still think some of them are questionable) and fixes both the breadcrumbs and object ID rendering in all other standard Zope objects I looked at.

One realization I am having out of all of this: Python 2 and Zope 4 is very very brittle. The faster we get Zope 5 out the better. I would explicitly advertise Zope 4 as a stepping stone that helps people port existing applications to Python 3, but not a long-term release. Users should immediately upgrade to Zope 5 to receive long-term support when they are ready for Python 3.

Please take a look.

icemac:
@dataflake Thank you for tackling this issue!

Issue #445, opened 2019-01-01
Z_DEBUG_MODE seems to be unused


jugmac00:
cf https://zope.readthedocs.io/en/latest/zdgbook/ObjectPublishing.html#environment-variables

When grepping the source code, Z_DEBUG_MODE only gets set twice in 

https://github.com/zopefoundation/Zope/blob/5d315a6d68ee31dbc2656a034c6026a7110651d3/src/Zope2/Startup/handlers.py#L49

and

https://github.com/zopefoundation/ZServer/blob/b690f57872dfff4792abba7c5763300c88dec0a1/src/ZServer/Zope2/Startup/handlers.py#L30

but gets nowhere read.

Also, afaik debug mode gets set via buildout config.

Is the variable obsolete? Anybody got insight on this?

I'd like to delete the references to this env variable in the docs.

Thank you!

dataflake:
I cannot find any place where it is actually read, either, so the documentation is definitely wrong when it claims that the variable sets the debug mode. I suggest removing the documentation references.

icemac:
Even the function setting `Z_DEBUG_MODE` seems to be unused.
I found some other places where `Z_DEBUG_MODE` is referenced in the documentation. It seem that it was true some years ago but currently there is another way to enable the debug mode.

jugmac00:
closed by #467

PR #467, opened 2019-02-03
Remove obsolete environment var

jugmac00:
This environment variable is no longer valid. Still, it has been
referenced, both in code by a setter, and several times in the
documentation.

The focus of this commit was to remove the references, not to fix
the documenation about how to debug Zope, as the whole documenation
will be amended, anyway - step by step.

modified:   docs/zdgbook/TestingAndDebugging.rst
modified:   docs/zope2book/MaintainingZope.rst
modified:   src/Zope2/Startup/handlers.py


jugmac00:
closes #445

jugmac00:
Hm, rebase is not enabled for this repository - squah it is.

PR #471, opened 2019-02-05
Prevent overlapping description in medium sized mobile views

pbauer:
This fixes several issues:

* On a medium sized mobile view the description was overlapping with the columns 'Size' and 'Last Modified' makeing that hard to read
* When using Products.ExternalEditor a pencil-link to edit a object is added after the title but the description was rendered over that link maing it unuseable.
* With this change the descriptions is rendered in the next line when it is so long that it would overlap with anything.

Before:
before

After:
after


drfho:
The table styling has to provide a good compromise between clear layout / quick orientation and completeness of the shown data. This is archived by a monotonous table grid and avoiding any jumping of the row heights. The additional pencil icon should be rendered grid conformant (i.e. as additional table column). Actually there are some situations near the medium-breakpoint where an unpleasant text overlap by long titles can occur. Maybe someone has an idea, how to avoid that by more sophisticated CSS or by JS?

icemac:
@pbauer Could you please check against the commit of @drfho?

drfho:
@pbauer :  ... and for simulating a grid conformant implanting of the ExternalEditor icon here a tentatively code snnippet in JavaScript you can  evaluate per browser webdev-tool:

```
 $('table.objectItems').find('tr').each(function(){
		$(this).find('th').eq(1).after('<th>&nbsp;</th>');
		var random_boolean = Math.random() >= 0.5;
		if (random_boolean) {
			$(this).find('td').eq(1).after('<td><i title="External Editor" class="far fa-edit"></i></td>');
		} else {
			$(this).find('td').eq(1).after('<td>&nbsp;</i></td>');
		}
        
   });
```
zmi_grid

icemac:
@drfho This looks nice. It requires changes in ExternalEditor with overwrites the templates to add its link. Maybe it should be switched to do it using JavaScript as it would to drop the existing monkey patch mechanism. It could be included using the mechanism described in https://zope.readthedocs.io/en/latest/ZMI.html#use-custom-icons-and-resources.
Someone volunteering to create a PR in `Products.ExternalEditor`?

icemac:
Created an issue, see zopefoundation/Products.ExternalEditor#13.

PR #473, opened 2019-02-08
- Make sure conflicts are always retried and not masked by exception views

### LINKED ISSUE #413: Retry on ConflictError is broken

dataflake:
Fixes #413 where an exception view registered for `ConflictError` or any of its subclasses would prevent transactions from being automatically retried up to the configured maximum number of retries.

That test is really nasty, but it was a lot of back and forth to create an environment that actually showed the problem.


src/ZPublisher/tests/test_WSGIPublisher.py
```diff
@@ -469,6 +469,48 @@ def _publish(request, module_info):
         finally:
             HTTPRequest.retry_max_count = original_retry_max_count
 
+    def test_retry_ConflictError(self):
+        # Make sure requests are retried as often as configured
+        from zope.interface import directlyProvides
+        from zope.publisher.browser import IDefaultBrowserLayer
+        registerExceptionView(ConflictError)
```
icemac: I think the change you made ist not specific for `ConflictError` it should work for every exception class (besides `Unauthorized`), so I'd prefer using another exception class here.

dataflake: I have changed the registration to just `Exception`. I have also renamed the test method with a more descriptive name and added a better comment at the top.

PR #476, opened 2019-02-08
Replace deprecated splitport and splittype.

icemac:
Python 3.8 deprecates `splitport` and `splittype`, see
https://github.com/python/cpython/blob/master/Lib/urllib/parse.py#L1039-L1042
resp.
https://github.com/python/cpython/blob/master/Lib/urllib/parse.py#L969-L972

VirtualHostMonster.py imports `splitport` from `HTTPRequest`.
The new function is modelled after the requirements of its tests.

The default ports changed from str to int because `urlparse` returns int port
numbers.


src/ZPublisher/HTTPRequest.py
```diff
@@ -46,12 +46,10 @@
...
 if PY3:
     from html import escape
-    from urllib.parse import splitport
-    from urllib.parse import splittype
+    from urllib.parse import urlparse
 else:
     from cgi import escape
-    from urllib import splitport
-    from urllib import splittype
+    from urlparse import urlparse
```
dataflake: You could also use `from six.moves.urllib.parse import urlparse` to get around the try/except, which `isort` ignores for import sorting.

PR #477, opened 2019-02-08
Add preliminary support for Python 3.8.

icemac:


icemac:
This should be merged after RestrictedPython is released or the branch there is at least merged.

PR #479, opened 2019-02-08
Fix import sort order.

icemac:
This commit is mainly the result of `tox -eisort-apply`.
It makes the first part of the linting environments happy.

What do you think of it?
Is it worth the hassle?
Should the isort config be adapted?

### DIFF
--- a/src/App/ApplicationManager.py
+++ b/src/App/ApplicationManager.py
@@ -11,14 +11,9 @@
 #
 ##############################################################################
 
-import os
-import sys
-
 from AccessControl.class_init import InitializeClass
 from AccessControl.requestmethod import requestmethod
 from Acquisition import Implicit
-from six.moves.urllib import parse
-
 from App.config import getConfiguration
 from App.Management import Tabs
 from App.special_dtml import DTMLFile
@@ -27,6 +22,10 @@ from App.version_txt import version_txt
 from OFS.Traversable import Traversable
 from Persistence import Persistent
 from Products.PageTemplates.PageTemplateFile import PageTemplateFile
+from six.moves.urllib import parse
+
+import os
+import sys
 
 
 class FakeConnection(object):
--- a/src/App/Common.py
+++ b/src/App/Common.py
@@ -12,14 +12,17 @@
 ##############################################################################
 """Commonly used utility functions."""
 
+from Acquisition import aq_base
+from Acquisition import aq_parent
+# BBB
+from os.path import realpath  # NOQA
+
 import os
 import sys
 import time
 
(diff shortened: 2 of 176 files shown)


dataflake:
Can the isort configuration be adapted to our (informal) convention? Should we continue using such a convention?

```
import builtin1
from builtin2 import foo

import six
import other_thirdparty

from Acquisition import aq_base
from zope.library1 import bar

import localmodule
```

dataflake:
For a customer project I have been using this configuration in `setup.cfg`:

```
[isort]
force_single_line = True
combine_as_imports = True
sections = FUTURE,STDLIB,THIRDPARTY,ZOPE,FIRSTPARTY,LOCALFOLDER
known_third_party = <e.g. six goes here>
default_section = ZOPE
line_length = 79
lines_after_imports = 2
not_skip =
    __init__.py
```

dataflake:
What really annoys me is that there's no configuration to bunch "from..." and "import..." together. I'd really prefer something like this:

```
import foo
from foo.bar import baz
import zzz
```

icemac:
I prefer having a single block of import statements sorted alphabetically case sensitive. This way I have to think least where an import has to be place: add it to the block sort the block using a keyboard shortcut and done.

I know there are many other habits so we have to decide on one or drop forcing the sort order of imports from the Zope project.

The one in https://github.com/zopefoundation/Zope/blob/master/setup.cfg#L13 is the one used in Plone projects, I think. (@loechel can you confirm this?)

icemac:
@dataflake wrote:

> For a customer project I have been using this configuration in `setup.cfg`:
> 
> ```
> [isort]
> force_single_line = True
> combine_as_imports = True
> sections = FUTURE,STDLIB,THIRDPARTY,ZOPE,FIRSTPARTY,LOCALFOLDER
> known_third_party = <e.g. six goes here>
> default_section = ZOPE
> line_length = 79
> lines_after_imports = 2
> not_skip =
>     __init__.py
> ```

Looks interesting, although I'd prefer to merge `THIRDPARTY` and `ZOPE` as there are at least some packages used by Zope which are hard to decide and third party packages have to be listed.

Maybe you could try it out in a separate PR. There is a tox environment `isort-apply` so the isort config could be applied using `tox -eisort-apply`.

dataflake:
With the configuration I pasted everything that is not specifically excluded will end up as ``ZOPE``. The only packages I would explicitly separate using ``known_thirdparty`` are those that aren't under the zopefoundation umbrella.

dataflake:
I have created a second PR #529 off of current master with a configuration I talked about in here.

Issue #438, opened 2018-12-22
Z_REALM seems to be unused


jugmac00:
I am planning to rework https://zope.readthedocs.io/en/latest/zdgbook/ObjectPublishing.html for the final Zope 4 release.

In chapter 4.11 I stumbled upon the "Z_REALM" environment variable.

Grepping the source only revealed one hit, where the value gets set by a function called "http_realm" in handlers.py.
https://github.com/zopefoundation/Zope/blob/5d315a6d68ee31dbc2656a034c6026a7110651d3/src/Zope2/Startup/handlers.py#L70

- this function seems to be never called
- Z_REALM seems to be never read
- function is not under test
- no docstring to bring light

Function seems unused.

Also, at least for wsgi config - there is http-realm directive.

Odd or not, the very same function appears in handler.py of the ZServer package.
https://github.com/zopefoundation/ZServer/blob/b690f57872dfff4792abba7c5763300c88dec0a1/src/ZServer/Zope2/Startup/handlers.py#L78

Anybody can bring light into darkness? Maybe @icemac or @hannosch ?

dataflake:
I have a feeling that's a leftover from when `ZServer` was extricated from Zope itself.

I'd leave the code but remove the documentation.

icemac:
I think even the function can be removed, the HTTP realm is now set like this:

https://github.com/zopefoundation/Zope/blob/47c71380812912b77639d32a7eb5fd7958a96c5a/src/Zope2/Startup/starter.py#L76-L77

and ZServer has its own (unused) version of `http_realm()`:

https://github.com/zopefoundation/ZServer/blob/110bdf04e08dbb01e868534e950e557e0d749a41/src/ZServer/Zope2/Startup/handlers.py#L73-L75

PR #486, opened 2019-02-10
Remove references to obsolete Z_REALM

jugmac00:
The Z_REALM environment variable is no longer used to control the
realm name of a basic auth dialog.


jugmac00:
closes #438 - not 483..

PR #494, opened 2019-02-12
Resurrected copyright and license page

### LINKED ISSUE #482: ZMI: Broken copyright link
There is a link called "© Zope Foundation" in the left panel, it points to http://localhost:8080/manage_copyright which does not (no longer?) exist:

broken link

Where should this link point to? 
To http://www.zope.org/en/latest/foundation.html?

dataflake:
Fixes #482 

I have started with the old manage_copyright and removed everything that was outdated.

Please note that I have deliberately avoided a `manage_` prefix this time. This is expressly supposed to be a public page and should not be hidden behind rules that forbid public access to pages ending in `manage_XXX`.


drfho:
Hello @dataflake , I added ZMI styles:
https://github.com/zopefoundation/Zope/pull/494/commits/5c1f5dcd654de5abd0ffb198011c5a8e3cf4d15f
Discussion: Instead of using <dtml-var manage_page_header|footer> I used static html to make sure that this html-page works without any context. Okay or not?

zope_copyrightpage

dataflake:
That's great, thanks! I had not added any styling because I was not sure if any of these resources would break when visited by anonymous users outside the ZMI. Looks good.

PR #501, opened 2019-02-14
ZMI-add does not recognize absolute urls #490

### LINKED ISSUE #490: ZMI-add does not recognize absolute urls
The ZMI add modal dialogue does not recognize absolute urls in the form action and puts the base url before them.

I observe this behaviour with Plone 5.2a1 (which embeds Zope 4.0b8): creating a Plone site opens the dialogue via `http://localhost:8180/manage_addProduct/CMFPlone/zmi_constructor?zmi_dialog=modal`; this redirects to `http://localhost:8180/@@plone-addsite?site_id=Plone`; this returns a form with action `http://localhost:8180/@@plone-addsite?site_id=Plone`; some JS-magic transforms this into `http://localhost:8180/manage_addProduct/CMFPlone/http://localhost:8180/@@plone-addsite` rather then letting the action as it is.

drfho:
Fixes #490 

@dataflake according to your suggestion "Could you add a check in the `addItem` function in `zmi_base.js` that checks if the `url_action` variable contains a fully-formed URL?" I revised the JS code:
https://github.com/zopefoundation/Zope/compare/issue_490?expand=1#diff-a2c5c94dd7e2821eda76f068fc4d4157
Actually I could not test Plone, but my webdev tests with complete urls as select-option name have been successful (modal dialog will be avoided on absolute urls).

HINT: If adding a Zope product induces a chain of redirects the modal dialog might lead to complications. Or if its GUI is very complex the modal dialog might be inappropriate. For these cases a JS array  'no_modal_dialog'-allows exceptional GUI behaviour by adding the initial method name given as 'value'in the addItemSelector element. Example:

```
// \Zope\src\zmi\styles\resources\zmi_base.js, LINE 26
// List of Object Types Inserting Without Modal Dialog
var no_modal_dialog = [
	'manage_addRegistry',
	'manage_addUserFolder',
	'manage_addErrorLog',
	'manage_addVirtualHostMonster',
	'manage_addzmsform',
	'addPluggableAuthService',
+	'zmi_constructor',
];
```


dataflake:
Thanks @drfho, I'm waiting for feedback from Dieter since I don't have any sandbox with Plone or other products with that behavior.

dataflake:
I hate to say it, but this whole idea of blacklisting specific products or dialog names will not work. We can't cut new Zope releases just to make a third party product work that needs to be added to a blacklist.

There's two issues at play:

* Products like Plone that hand you a fully formed URL for the form action and don't want any interference.
* Products like the Site Error Log that do not collect any configuration option and don't need the dialog at all.

For the first case, just "get out of the way" and redirect the browser frame to that URL. No fuss, no magic, just give up.

To cover the second case, I would look for a way that "compatible" products, meaning those that need a configuration dialog, to flag that. The modal code looks for the flag and only comes into play if it is present.

With this strategy you don't need any blacklist or whitelist.

PR #504, opened 2019-02-15
Make sure the WSGI Response object respects lock semantics

### LINKED ISSUE #216: Should WSGIResponse.setBody mimic the HTTPResponse.setBody if lock is truthy?
`WSGIResponse.setBody` mimics the signature of `HTTPResponse.setBody` but not the semantics. If the application is passing `lock=True` to the method, then it will be expecting that future calls to `setBody` will be no-ops.
See https://github.com/zopefoundation/Zope/pull/215.

dataflake:
Fixes #216 
Expands #215 

I'm in favor of following the `HTTPResponse` behavior because IMHO there is no good reason for just discarding the lock semantics.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,11 @@ Fixes
 - Fix error when not selecting a file for upload in Files and Images
   (`#492 <https://github.com/zopefoundation/Zope/issues/492>`_)
 
+Other changes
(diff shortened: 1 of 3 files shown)


pbauer:
I tested this against Plone and had no fresh test-failures.

PR #509, opened 2019-02-17
Fix FindSupport binary value handling

### LINKED ISSUE #406: FindSupport is not able to handle bytes data
If `PrincipiaSearchSource` is called on a `File` or an `Image` it returns `bytes` but the search string is `text` which leads to an to `TypeError: a bytes-like object is required, not 'str'` in 

https://github.com/zopefoundation/Zope/blob/478319637795508082823eeabe907e34322b1fe2/src/OFS/FindSupport.py#L130

The same exception might happen two lines later on in 

https://github.com/zopefoundation/Zope/blob/478319637795508082823eeabe907e34322b1fe2/src/OFS/FindSupport.py#L132

But I did not try to provoke it.

I am not sure what a good solution can be:

* Should `PrincipiaSearchSource` always return text? (If the object contains binary data it should return an empty text?)
* Should the call in `FindSupport` only made if `PrincipiaSearchSource` returned `text`?
* Same for `SearchableText` – but this might be a bit easier to answer as the method contains `text` in its name.

dataflake:
Fixes #406 

This PR was made under the assumption that this is purely a Python 3 issue. Please correct me if I am wrong.


dataflake:
@icemac The workflow when a PR is ready for merge with a suggested change is very clunky. I guess you'll have to visit again to give the green light.

icemac:
@dataflake Don't you have the ability in the PR: "As an administrator, you may still merge this pull request."?
I habe disabled the option "Dismiss stale pull request approvals when new commits are pushed", so the review does not get invalidated by a new commit on the branch.

dataflake:
@icemac Yes I can still squash and merge, but my impression was that you frown upon that.

icemac:
@dataflake I think it is okay to use admin permissions from time to time. I'm doing so, too. 😀

Issue #511, opened 2019-02-19
Double quoting in `ZPublisher.HTTPRequest.search_type`: image input fields no longer correctly handled


d-maurer:
In Zope 2, `ZPublisher.HTTPRequest.search_type` was defined as `re.compile('(:[a-zA-Z][-a-zA-Z0-9_]+|\\.[xy])$').search`. For Zope [4], the re was turned into a raw string but the `\\.` not replaced by `\.`. As a consequence, the `.x` and `.y` suffixes for `image` input fields are no longer correctly recognized.

dataflake:
Fixed via eec22a9c19c7d3ea57a0f9c60203e511b833b280

PR #512, opened 2019-02-19
Issue #511: remove double quoting in `ZPublisher.HTTPRequest.search_t…

d-maurer:
…ype`


dataflake:
Could you add a test or at least provide me with enough information to write one? I'm not familar with the `processInputs` code.

d-maurer:
Jens Vagelpohl wrote at 2019-2-19 15:24 -0800:
>Could you add a test or at least provide me with enough information to write one? I'm not familar with the `processInputs` code.

As always with tests, it is important to decide what should be tested.
In this particular case, whether you want to test the
`search_type` definition (maybe even restricted to the case fixed
by this pull request) or whether you want to test `processInputs` itself.
The first is simple, the latter more or less complex (depending
to the depth `processInputs` should be tested).


I detected the error in Plone - with its `acl_users/plugins/manage_plugins`
ZMI page. There, an `<input type=image name="manage_delPlugins:method" ...>`
control is used. When such a control is clicked, the browser
sends two parameters "<name>.x" and "<namge>.y" with the coordinates
of the click. In the particular case, this
gives `manage_delPlugins:method.x` and `manage_delPlugins:method.y`.

The task of `ZPublisher.HTTPRequest.search_type` is to detect
the rightmost type indication in a control name. For
`manage_delPlugins:method.x` this would be the `.x` (as it indicates
the click into a control of type `image`), for `manage_delPlugins:method`
`:method` and for `manage_delPlugins` nothing (in the former cases, a
match object for the respective string would be returned).

`processInputs`'s task is complex: it turns the information from
the request's `environ` and `stdin` into the request's `form` dict
and (for `:method` parameters) extends the request's `PATH_INFO`.
There are many dozens of different cases handled by `processInputs`.


While `search_type` can easily be tested with simple unit tests (outlined
above),
`processInputs` is likely best tested with functional tests -
with input similar to that created by real browsers.
I do not know whether the functional test infrastructure
has support for controls of type `image`. In the test
examples I know, the "click" does not incorporate coordinates
as it would be necessary to test such controls.


Finally, do we really need a test for this fix?
In Zope 2, we have had a working
`search_type=re.compile('(:[a-zA-Z][-a-zA-Z0-9_]+|\\.[xy])$').search`,
in Zope [4], this became
`search_type = re.compile(r'(:[a-zA-Z][-a-zA-Z0-9_]+|\\.[xy])$').search`
(not working for controls of type `image`);
the fix is
`search_type = re.compile(r'(:[a-zA-Z][-a-zA-Z0-9_]+|\.[xy])$').search`
which obviously compensates for the (new) use of a raw string in the re
definition.
An alternate fix would be to revert the "raw"ification.

icemac:
If we have had a test beforehand, it would not have been possible to introduce this bug in Zope 4. So I'd like to have a test checking that the regex does what it promises as regex code is not so easy to read. I'd treat it as code which should be tested. If it would be totally clear what it does, the bug would not have been introduced, too.

Just my 2 cents.

dataflake:
A test for the regex itself is fine. Just a list of potential input strings that show which one it is matching and what the results should be.

d-maurer:
Performing the `ZPublisher` tests locally surprisingly resulted in errors:
```
xmlsec3: ztest -s ZPublisher
/home/dieter/tmp/Zope/src /home/dieter/tmp
...
  File "/home/dieter/tmp/Zope/src/ZPublisher/tests/testPostTraversal.py", line 8, in <module>
    Zope2.startup_wsgi()
  File "/home/dieter/tmp/Zope/src/Zope2/__init__.py", line 49, in startup_wsgi
    _startup()
  File "/home/dieter/tmp/Zope/src/Zope2/App/startup.py", line 116, in startup
    DB = dbtab.getDatabase('/', is_root=1)
AttributeError: 'NoneType' object has no attribute 'getDatabase'
...
  Ran 341 tests with 0 failures, 1 errors and 1 skipped in 0.712 seconds.
  Set up Testing.ZopeTestCase.layer.ZopeLite Traceback (most recent call last):
  File "/home/dieter/tmp/xmlsec3/python/lib/python3.6/site-packages/zope/testrunner/runner.py", line 451, in run_layer
...
  File "/home/dieter/tmp/Zope/src/Testing/ZopeTestCase/ZopeLite.py", line 157, in _startup
    _theApp = Zope2.app()
  File "/home/dieter/tmp/Zope/src/Zope2/__init__.py", line 55, in app
    return bobo_application(*args, **kw)
TypeError: 'NoneType' object is not callable
...
Test-modules with import problems:
  ZPublisher.tests.testPostTraversal

Tests with errors:
   Layer: Testing.ZopeTestCase.layer.ZopeLite
Total: 341 tests, 0 failures, 2 errors and 1 skipped in 0.879 seconds.
```
Looks as if the tests were not properly isolated.

dataflake:
I have seen test isolation issues in several places. Errors that went away if I ran the whole battery of tests.

d-maurer:
> Performing the `ZPublisher` tests locally surprisingly resulted in errors:
> ...
> Looks as if the tests were not properly isolated.

Turned out that the `ZPublisher` tests require the inclusion of a Zope configuration file. With a new version of `ztest` (supporting the new option `--config-wsgi`), `ztest --config-wsgi etc/wsgi.conf -s ZPublisher` runs the tests without problem.

icemac:
@d-maurer What is `ztest`? I do not see it listed in `bin/` of a Zope instance.

d-maurer:
@icemac 
> @d-maurer What is `ztest`? I do not see it listed in `bin/` of a Zope instance.

`ztest` is [`dm.zopepatches.ztest`](https://pypi.org/project/dm-zopepatches-ztest/) - what formerly has bin "zopectl test" (of which the support has been dropped - to my dismay). It is a tiny wrapper of `zope.testrunner` adding automatic detection of `package_path` information and loading a Zope configuration file.

PR #514, opened 2019-02-19
Improve wording.

jugmac00:
No skipped ci as otherwise only admins can merge.

### DIFF
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -146,14 +146,14 @@ Publishable Object Requirements
 
 Zope has few restrictions on publishable objects. The basic rule is
 that the object must have a doc string. This requirement goes for
-method objects too.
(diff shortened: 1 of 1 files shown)


jugmac00:
Hi @ale-rt - do you think the method objects from your links, which seems to be a C API documentation, has anything to do with Zope's object publishing? If so, I'll certainly revert my changes. P.S.: Thanks for the links. I was well aware that in Python almost everything is a object, but haven't dug into the C-API, yet.

ale-rt:
@jugmac00 I am completely ignorant about that topic (and many others) :)
I approved with pleasure your PR because IMHO it is an improvement (and thanks for all the ones you are doing!).
I was just pointing out that **maybe** "method objects" might cover a wider class of **things** that Zope can (could?) publish.
That sentence is there since at least 10 years and maybe it was referring to **things** which now are not even part of Python.

PR #527, opened 2019-03-12
Prevent `ValueError: I/O operation on closed file.`

icemac:
when using `ZPublisher.HTTPRequest.FileUpload` in Python 3.

This happened before because Python 3 seems to garbage collect objects earlier
than Python 2 and FieldStorage was no longer referenced after storing its file
on the `FileUpload` instance. But `FieldStorage` closed its file in its
`__del__` method.

Also see #141.

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1684,6 +1684,9 @@ class FileUpload(object):
         self.headers = aFieldStorage.headers
         self.filename = aFieldStorage.filename
         self.name = aFieldStorage.name
+        # Keep a reference to the field storage to prevent it from closing
+        # its file during garbage collection:
+        self.__field_storage = aFieldStorage
 
         # Add an assertion to the rfc822.Message object that implements
         # self.headers so that managed code can access them.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,12 @@ Features
   which are deprecated in Python 3.8.
   (`#476 <https://github.com/zopefoundation/Zope/pull/476>`_)
 
+Fixes
++++++
+
+- Prevent ``ValueError: I/O operation on closed file.`` when using
+  ``ZPublisher.HTTPRequest.FileUpload`` in Python 3.
+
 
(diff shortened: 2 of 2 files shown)


dataflake:
I'm out of my depth on this and need to remove myself as reviewer.

src/ZPublisher/HTTPRequest.py
```diff
@@ -1684,6 +1684,9 @@ def __init__(self, aFieldStorage):
         self.headers = aFieldStorage.headers
         self.filename = aFieldStorage.filename
         self.name = aFieldStorage.name
+        # Keep a reference to the field storage to prevent it from closing
+        # its file during garbage collection:
+        self.__field_storage = aFieldStorage
```
tseaver: Have you traced through to make sure that we aren't leaking anything here?

d-maurer:
Tres Seaver wrote at 2019-3-13 18:40 +0000:
>> @@ -1684,6 +1684,9 @@ def __init__(self, aFieldStorage):
>         self.headers = aFieldStorage.headers
>         self.filename = aFieldStorage.filename
>         self.name = aFieldStorage.name
>+        # Keep a reference to the field storage to prevent it from closing
>+        # its file during garbage collection:
>+        self.__field_storage = aFieldStorage
>
>Have you traced through to make sure that we aren't leaking anything here?

In principle, "ZPublisher.HTTPRequest.HTTPRequest.processInputs"
already has:

```
        # Keep a reference to the FieldStorage. Otherwise it's
        # __del__ method is called too early and closing FieldStorage.file.
        self._hold(fs)

```

Why does it not prevent the observed problem?
Maybe, a different "fs"...

If a different "fs" is causing the problem, then the code
above shows how to pretect it -- without big risks with respect
to leaks.

icemac:
@d-maurer Thank you for your explanation. I am using `FileUpload` in a unittest where `ZPublisher.HTTPRequest.HTTPRequest.processInputs` is not called.
As there is already a solution and I am not sure if I am leaking anything here I am closing this PR. It should be easier to change my test code.

PR #529, opened 2019-03-19
import sort attempt

dataflake:
closes #479


icemac:
Even though this PR changes more files and more lines than #479 it seems to be more natural to the code base. According to [TravisCI](https://travis-ci.org/zopefoundation/Zope/jobs/508305645#L507-L549) there are still about 40 places where the import sort order does not match the requirements?

@dataflake Did you change the sort order by hand or did you use `tox -eisort-apply`? If the latter is true we should think again about the config because I think it is not a good idea if the correct sort order cannot be adjusted automatically. (In contrast #479 does not have any sorting issues left on [TravisCI](https://travis-ci.org/zopefoundation/Zope/jobs/490568082#L508l).)

I personally do not like the differentiation between `THIRDPARTY` and `ZOPE` in the import sections because we have to curate the list of packages belonging to `THIRDPARTY` by hand. I'd like to collapse them into one section as `default_section` for all imports which are not in one of the other sections.

dataflake:
I did not use `tox`, but `isort` directly from the command line to apply the changes.

I am against mixing unrelated items like `six` with Zope library packages.

dataflake:
I don't know why, but before I had it working so that sorting was case-insensitive, but now all I can get from `isort` is case-sensitive sorting, so the latest checkin uses that consistently.

I also noted that the version of isort pulled for the tox `isort-apply` configuration is the latest, but `lint-py36` uses an older version. That may have caused inconsistencies as well.

With this latest checking both `tox -eisort-apply` an `tox -elint-py36` agree.

dataflake:
P.S.: About the `ZOPE` versus `THIRDPARTY` section and the need to curate them, please keep in mind how often you'd really be pulling new third party dependencies into the code. That will be a rare occurrence. The effort to keep the isort configuration up to date appears minimal to me.

dataflake:
@icemac FYI, after merging the PEP-8 fixes from `master` into this branch and with a little cleanup we now have both `tox` and Travis CI tests GREEN from top to bottom.

icemac:
@dataflake Thank you very much to get this solved!

PR #536, opened 2019-04-05
Make showing the ZMI modal add dialog configurable per product

### LINKED ISSUE #535: ZMI: Determining if an object doesn't need/want a modal Add dialog
Right now there is a hardcoded list of constructor and product names inside `zmi/styles/resources/zmi_base.js` for items that don't want to show the modal dialog when adding them to a container.

There are many Zope products out there that worked fine before just attempting to redirect to their container without showing an add form, which now pop up the modal dialog that just renders the container contents. And if you dismiss the modal dialog the underlying page shows the old container contents, as if nothing was added.

I think we can agree that the solution is sub-optimal. Third-party authors should be able to set a flag themselves that will prevent the modal from popping up. This could be anything from a new class attribute (they already have to add `zmi_icon` so that's no big deal) or maybe just an added query string item that gets passed with the product's own redirect URL. I'm open for pretty much any solution that allows the product author to decide without asking to have Zope code changed.

dataflake:
fixes #535 

With this change a product author can prevent the "Add" modal by setting a class-level variable `zmi_modal` to any non-true value (empty string, `None`, `False`, etc.). The default is `True`, set in `OFS.SimpleItem.SimpleItem`. Classes that derive from something else and don't have `zmi_modal` set at all will also default to show the modal.

@drfho: Don't be confused by the massive changes to `manage_navbar.dtml`, that's all white space and reformatting. I simply cannot stand tabs anywhere. The only "real" change in there is adding the `data-dialog` attribute on each option in the select list.

The current harcoded list inside `zmi_base.js` can/should stick around until the products in there are updated to use the new convention.


docs/zope4/migration/zmi.rst
```diff
@@ -24,6 +24,13 @@ Example to use the info icon (i in a circle)::
 
     zmi_icon = 'fas fa-info-circle'
 
+If your product does not need to show a form when adding an object in the ZMI
+using the `Select type to add` dropdown, which only applies to a small
```
icemac: I think the part between the commas ("which only applies to a small …") make the sentence harder to read. I am not sure if it provides additional value to the sentence. Maybe it could be removed.

docs/zope4/migration/zmi.rst
```diff
@@ -24,6 +24,13 @@ Example to use the info icon (i in a circle)::
...
+percentage of available Zope products, you need to add another class variable
+to prevent the modal dialog from appearing::
+
+    zmi_modal = False
```
icemac: As this attribute will be rarely used I'd like to have a name which tells ab bit more about its purpose. Maybe `zmi_show_modal_add_dialog` or `zmi_render_modal_dialog_on_add`.

src/OFS/ObjectManager.py
```diff
@@ -274,6 +274,12 @@ def all_meta_types(self, interfaces=None):
                 if container_filter(self):
                     meta_types.append(entry)
 
+        # Synthesize a new key into each item that decides whether to show
+        # the modal add dialog in the Zope 4 ZMI
+        for mt in meta_types:
+            want_modal = getattr(mt.get('instance', None), 'zmi_modal', True)
+            mt['zmi_modal'] = (want_modal and 'modal') or ''
```
icemac:
```suggestion
            mt['zmi_modal'] = 'modal' if want_modal else ''
```

src/zmi/styles/resources/zmi_base.js
```diff
@@ -31,7 +32,6 @@ function addItem( elm, base_url, zmi_dialog='modal') {
 			'manage_addRegistry',
 			'manage_addUserFolder',
 			'manage_addErrorLog',
-			'manage_addVirtualHostMonster',
```
icemac: Are you planning to create a ticket to shorten this list after the PR is merged?

docs/zope4/migration/zmi.rst
dataflake: I have changed the wording to be clearer now. I want to make sure people realize this only affects very few products.

docs/zope4/migration/zmi.rst
dataflake: I have renamed the variable to `zmi_show_add_dialog`

src/OFS/ObjectManager.py
dataflake: Change incorporated. Two different style habits colliding ;-)

src/zmi/styles/resources/zmi_base.js
dataflake: Not "a ticket" but a ticket in each of the affected products' issue tracker that we control. Some we don't.

PR #538, opened 2019-04-07
Always keep form buttons visible on the content list for large folders

### LINKED ISSUE #537: Question: Button positioning in main folder list view
I'm dealing with many ZODB folders that have a large number of items in them. How hard would it be to make the buttons ("Rename", "Cut", "Copy", etc) always visible at the bottom of the page, no matter where you scroll to?

dataflake:
fixes #537 

Usability improvement: I dont want to have to scroll all the way to the bottom to reach the form buttons when a folder contains a lot of items. This is very helpful if you have to deal with large folders on a daily basis.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@ Fixes
 - make the ZMI `Find` tab work for searching HTML tags
   by adding support for `Tainted` strings in ``ZopeFind``
 
+- prevent ``mkwsgiinstance`` from blowing up parsing ``buildout.cfg``
+
 Features
 ++++++++
 
@@ -29,6 +31,9 @@ Other changes
 
 - Always keep action buttons visible on the content list for large folders
   (`#537 <https://github.com/zopefoundation/Zope/issues/537>`_)
+  
+- Make showing the ZMI modal add dialog configurable per product
+  (`#535 <https://github.com/zopefoundation/Zope/issues/535>`_)
 
 - Added a few Zope 4 ZMI screenshots to the documentation
   (`#378 <https://github.com/zopefoundation/Zope/issues/378>`_)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ Features
 Other changes
 +++++++++++++
 
+- Always keep action buttons visible on the content list for large folders
+  (`#537 <https://github.com/zopefoundation/Zope/issues/537>`_)
+
 - Added a few Zope 4 ZMI screenshots to the documentation
   (`#378 <https://github.com/zopefoundation/Zope/issues/378>`_)
 
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -93,7 +93,7 @@ and extra-steps that this list glosses over.
 
 
 URL Traversal
-=============
+-------------
 
 Traversal is the process the publisher uses to locate the published
 object. Typically the publisher locates the published object by
@@ -142,7 +142,7 @@ Now let's take a closer look at traversal.
 
 
 Publishable Object Requirements
-===============================
+-------------------------------
 
 Zope has few restrictions on publishable objects. The basic rule is
 that the object must have a doc string. This requirement goes for
@@ -156,7 +156,7 @@ Finally, published objects cannot be Python modules.
 
 
 Traversal Methods
-=================
+-----------------
 
 During traversal, *ZPublisher* cuts the URL into path elements
 delimited by slashes, and uses each path element to traverse from the
@@ -212,7 +212,7 @@ will try in order to find the next object:
   
 
 Publishing Methods
-==================
+------------------
 
 Once the published object is located with traversal, Zope *publishes*
 it in one of three possible ways:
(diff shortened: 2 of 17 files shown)


drfho:
@dataflake 
in respect to your proposal to work with full opacity I changed the concept:
https://github.com/zopefoundation/Zope/pull/538/commits/1c5591e2933f2ef9bc2f0d15b5e21605adc7b2bb

objectitems_select

When you load the page, on the first sight it looks traditional. If you do a selection now and the controls are out of the viewport they will jump to the window's bottom and become visible. Advantage: if you do not select anything (most cases)  the viewport is not reduced by the fixed control bar.

dataflake:
From a UX standpoint I would say this crosses into "behavior that is not apparent, unexpected and needs to be explained to users". Please don't do this. I personally don't mind that the viewport is reduced by the height of that row of buttons.

drfho:
@dataflake 
okay, but it's just the classical case of a context depended menu. Usually this should not leave any questions.  So, personally I am not sure whether the quoted UX rule perfectly fits here. But who knows?
Please, consider that we came from a different problem. In that context, I am more afraid of a consecutive UX problem: now _emphasize_ permanently buttons which are rarely needed and only when the lists are long.
Maybe we should ask others?

If majority likes the steady control bar you can quickly deactivate it's jumping by commenting it like this:
Zope\src\OFS\zpt\main.zpt LINE 174
```
  function select_objectitem(ob) {
    ob.parent().parent().toggleClass('checked');
   //  if ( !zmicontrols_visible() ) {
   //  $('form#objectItems').addClass('selected');
   // }
  };
```
and shortending the css
Zope\src\zmi\styles\resources\zmi_base.css LINE 242
```
- .zmi form#objectItems.selected {
+ .zmi form#objectItems {
	margin-bottom: 60px;
}
- .zmi form#objectItems.selected .form-group.zmi-controls {
+ .zmi form#objectItems .form-group.zmi-controls {
	position: fixed;
...
```

drfho:
@dataflake 
FYI: to make it simple I just excluded "orderd folder" controls and small screens. For both we had to go deeper into the facelift - maybe for a later version.

dataflake:
@drfho My main concern is the way the buttons are "activated". Clicking the checkbox for that is the weirdness I have a hard time with. That's the unexpected and confusing part. At that point it may be better to have some other distinct element that's visible when the buttons are below the viewport. Something that suggests "clicking here provides some options or interactions".

dataflake:
@drfho I'll need to rephrase. You're right in that the buttons are not needed often. And they are only needed in conjunction with selecting one of the list items. I wonder if there's a way based on your checkbox click solution that makes that interaction (and its resulting button appearance) less surprising. Some cue on or near the checkbox? I'm just throwing out ideas.

By the way, the buttons only appear for me on Firefox, not on Chromium or Safari.

dataflake:
Correction: It doesn't work on Safari. Chromium works now.

Another idea: The buttons appear when a selection is made. Can you make them disappear when the checkbox is unchecked (or, to be more precise, as soon as no checkbox is selected anymore)?

drfho:
many thanks for the input. i'll check that checking behavior (tomorrow)

drfho:
@dataflake 
hi, following your proposal the control bar now toggels it's appearance with checkbox selecting. I had to redesign this control bar so that it can handle the "Folder-(Ordered)"-object. Now that bar has a better responsive behaviour and looks  nicer on big screens, because now it can run in-line.
zmi_controls

_Help wanted:_
As a CSS class-name for the html-body element we generally need CSS-syntax-conformant meta-names. But "Folder-(Ordered)" is not a valid because it contains brackets. So I applied a very bad patch  as a workaround to Zope\src\App\dtml\manage_page_header.dtml:
`<dtml-var "this().meta_type.replace(' ', '-').replace('(', '').replace(')', '')">`
Maybe you got a better idea how to purge the class names in a more elegant/general way?

Many thanks
fh

dataflake:
Thank you @drfho, that's really nice. I appreciate the great work!

On your question about rewriting the DTML in `manage_page_header` I have to admit defeat. I have looked (and written) various solutions for a couple hours now, only to realize they won't work:

- added code into the `ZTUtils` module and made it public, only to realize that you cannot get to that module in DTML
- then I rewrote the entire template as a page template. Much nicer-looking code, and access to `ZTUtils`. But then I found out that importing `PageTemplateFile` from `Products.PageTemplates` into `App.Management` so I can declare the new template causes a circular import that breaks everything.

I also looked at all functions available inside DTML, but none offer an elegant solution. The last resort would be to define a method on one of the common base classes that would always be available, but that seems like too much just to create a CSS class name.

Since Michael likes your solution as well I will merge this now because it's a great win as it stands. We can revisit details later.

jugmac00:
> * then I rewrote the entire template as a page template. Much nicer-looking code, and access to `ZTUtils`. But then I found out that importing `PageTemplateFile` from `Products.PageTemplates` into `App.Management` so I can declare the new template causes a circular import that breaks everything.

As I recently worked on a Flask project, and for Flask it is pretty common to apply some workarounds to avoid circular imports...

Maybe you could either put the import statement at the end of the module, or inside a method.

I bet you knew this anyway.

PR #539, opened 2019-04-09
Revise section "HTTP Responses"

jugmac00:
- fix minor wording and comma issues
- improve formatting
- improve example code
- add note about empty return values / 204 status code

### DIFF
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -263,24 +263,38 @@ value like ``text/html; charset=UTF-8``.
 HTTP Responses
 ==============
 
-Normally the published method returns a string which is considered the
-body of the HTTP response. The response headers can be controlled by
-calling methods on the response object, which is described later in
-the chapter. Optionally, the published method can return a tuple with
-the title, and body of the response. In this case, the publisher
-returns an generated HTML page, with the first item of the tuple used
-for the HTML 'title' of the page, and the second item as the contents
-of the HTML 'body' tag. For example a response of::
+Usually, the published method returns a string which is considered
+the body of the HTTP response. The response headers can be controlled
+by calling methods on the response object, which is described later in
+the chapter.
 
-  ('response', 'the response')
+.. note::
+
+  When the return value is empty, e.g. an empty list, instead of
+  returning an empty page, Zope issues a header with a 204 status code.
+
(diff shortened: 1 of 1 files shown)


docs/zdgbook/ObjectPublishing.rst
```diff
@@ -263,25 +263,39 @@ value like ``text/html; charset=UTF-8``.
...
+  returning an empty page, Zope issues a header with a 204 status code.
+
+  Depending on the used client, it looks like nothing happens.
+
+
+Optionally, the published method can return a tuple with
+the title and the body of the response. In this case, the publisher
+returns a generated HTML page, with the first item of the tuple used
+for the value of the HTML ``title`` tag of the page, and the second
+item as the content of the HTML ``body`` tag.
```
icemac: I did not know about this feature.

@dataflake Do you think we can drop it? It does not make much sense to me.

dataflake:
No, please leave it. I had forgotten all about that "feature" myself but had to laugh when I realized that's exactly what I need today for some customer work.

dataflake:
P.S.: Which also gives me the opportunity to test is it still works

jugmac00:
@icemac My (biggest) motivation to revise all the docs actually is learning something new about Zope. But surprisingly, the issue with the 204 header is something I experienced, and I am happy to give something back.

@dataflake I manually test / review source code for all updated document sections - that's why it takes quite a while to do :-) Thinking of it, it would have been even better to check whether the special behaviours (tuple/204) are actually covered by tests.

icemac:
Having tests for the documented features would be a big bonus but maybe they are already in place.

jugmac00:
Here we go... you're right - there was already one:

```
    def test_setBody_2_tuple_wo_is_error_converted_to_HTML(self):
        EXPECTED = (b"<html>\n"
                    b"<head>\n<title>TITLE</title>\n</head>\n"
                    b"<body>\nBODY\n</body>\n"
                    b"</html>\n")
        response = self._makeOne()
        response.body = b'BEFORE'
        result = response.setBody(('TITLE', b'BODY'))
        self.assertTrue(result)
        self.assertEqual(response.body, EXPECTED)
        self.assertEqual(response.getHeader('Content-Type'),
                         'text/html; charset=utf-8')
        self.assertEqual(response.getHeader('Content-Length'),
                         str(len(EXPECTED)))
```

PR #552, opened 2019-04-15
Revise sub-sections about argument marshalling.

jugmac00:
- fix typos
- improve doc string
- improve formatting

### DIFF
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -611,18 +611,22 @@ The publisher marshals arguments from the request and automatically
 makes them available to the published object. This allows you to
 accept parameters from web forms without having to parse the
 forms. Your objects usually don't have to do anything special to be
-called from the web. Consider this function::
+called from the web.
+
+
+Consider this function::
 
       def greet(name):
-          "greet someone"
-          return "Hello, %s" % name
+          """Greet someone by name."""
+          return "Hello, %s!" % name
 
-You can provide the 'name' argument to this function by calling it
-with a URL like *greet?name=World*. You can also call it with a HTTP
-'POST' request which includes 'name' as a form variable.
+You can provide the ``name`` argument to this function by calling it
+with a **URL** like ``greet?name=World``. You can also call it with a **HTTP
+POST request** which includes ``name`` as a form variable.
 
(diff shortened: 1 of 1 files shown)


jugmac00:
Free beer to anybody who can show me the code where "URL0", "URL1", "URL2"... "URLPATHn", "BASEPATHn" and the likes get set (please add a link to the source code).

dataflake:
I believe the code to set `BASE` and `URL` variables is here:

https://github.com/zopefoundation/Zope/blob/e82a33aa90578941d76905a87f63d8d73f71e7de/src/ZPublisher/HTTPRequest.py#L1452-L1466

dataflake:
By the way, I also searched for `BASEPATH` and `URLPATH`. They show up in the code in a few places, but I have not found code that actually sets those on the Request object, and if you render a Request object they aren't there.

jugmac00:
> By the way, I also searched for `BASEPATH` and `URLPATH`. They show up in the code in a few places, but I have not found code that actually sets those on the Request object, and if you render a Request object they aren't there.

Thanks for the update! Fix comes with #555

jugmac00:
> I believe the code to set `BASE` and `URL` variables is here:
> 
> [Zope/src/ZPublisher/HTTPRequest.py](https://github.com/zopefoundation/Zope/blob/e82a33aa90578941d76905a87f63d8d73f71e7de/src/ZPublisher/HTTPRequest.py#L1452-L1466)
> 
> Lines 1452 to 1466 in [e82a33a](/zopefoundation/Zope/commit/e82a33aa90578941d76905a87f63d8d73f71e7de)
> 
>  # Cache URLN and BASEN in self.other. 
>  # This relies on a side effect of has_key. 
>  n = 0 
>  while 1: 
>      n = n + 1 
>      key = "URL%s" % n 
>      if key not in self:  # NOQA 
>          break 
>   
>  n = 0 
>  while 1: 
>      n = n + 1 
>      key = "BASE%s" % n 
>      if key not in self:  # NOQA 
>          break

@dataflake  Thank you! Shame on me. I only grepped for occourences in combination with + :-)

Hm, and I do not really get a grasp of the code snippets.

The local variable key gets something assigned, but afterwards key is no longer used. Hm... except the "in" operator has a side effect, the code snippet imho is a noop.

Issue #558, opened 2019-04-18
Several possible problems within the ZPublisher Converters module


jugmac00:
When working on #557 I spotted several possible problems within the Converters module.

The converters which inherit from `_unicode_converter` all fail under Python 3 and byte input.

Testcase:
```
def test_field2utext_with_binary_string(self):
    from ZPublisher.Converters import field2utext
    to_convert = b'abc'
    if PY2:
        expected = u'abc'
        self.assertEqual(field2utext(to_convert), expected)
    if PY3:
        expected = 'abc'
        self.assertEqual(field2utext(to_convert), expected)

Failure in test test_field2utext_with_binary_string (ZPublisher.tests.test_Converters.ConvertersTests)
Traceback (most recent call last):
  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.6/unittest/case.py", line 622, in run
    testMethod()
  File "/home/jugmac00/Projects/Zope/Zope/src/ZPublisher/tests/test_Converters.py", line 259, in test_field2utext_with_binary_string
    self.assertEqual(field2utext(to_convert), expected)
  File "/usr/lib64/python3.6/unittest/case.py", line 846, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/python3.6/unittest/case.py", line 1220, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/lib64/python3.6/unittest/case.py", line 687, in fail
    raise self.failureException(msg)
AssertionError: "b'abc'" != 'abc'
- b'abc'
+ abc
```

This problem applies to field2utokens, field2utext, field2string. field2ulines on the other hand does not inherit from _unicode_converter and has to be examined separately.

I spotted https://github.com/zopefoundation/Zope/blob/5733da7d210d8fe322bfd03a01852bb02b529706/src/ZPublisher/Converters.py#L186 as the root problem for this behaviour, at least with Python 3.

----

Another possible problem is in https://github.com/zopefoundation/Zope/blob/5733da7d210d8fe322bfd03a01852bb02b529706/src/ZPublisher/Converters.py#L209 - as line breaks only get normalized when there is input of type byte.

[...]

PR #560, opened 2019-04-19
[Do not merge] Fix converter basics

jugmac00:
Most of the converters rely on field2string and field2bytes.

field2bytes was broken for Python 3 and file-like input.

This has been fixed and a test has been added to prevent regression.

Also, the docstrings has been improved.

This is the first pull request for #558

### DIFF
--- a/src/ZPublisher/Converters.py
+++ b/src/ZPublisher/Converters.py
@@ -31,9 +31,12 @@ default_encoding = 'utf-8'
 
 
 def field2string(v):
-    """Converts value to native strings.
+    """Converts a value into a native string.
 
     So always to `str` no matter which Python version you are on.
+    That means:
+    - str / byte string for Python 2
+    - str / "unicode string" for Python 3
     """
     if hasattr(v, 'read'):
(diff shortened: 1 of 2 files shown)


jugmac00:
I think I have to setup a real form with an file upload and step through.

And maybe using a cgi.FieldStorage.fle instead of the fake for testing would a much better idea.

Once again, thanks a lot @dataflake

jugmac00:
@dataflake Your initial thoughts on where the buggy behaviour is located was right.

I stepped through a file upload with a "string" type converter attached to the field and that was the result:
```
python3
(Pdb) item
<ZPublisher.HTTPRequest.FileUpload object at 0x7fb79e74e588>
(Pdb) type(item.file.read())
<class 'bytes'>


python2
(Pdb) item
<ZPublisher.HTTPRequest.FileUpload object at 0x7f907c06b7d0>
(Pdb) type(item.read())
<type 'str'>
```

Thanks a lot for preventing me from introducing a new bug!

I am closing this pr.

PR #563, opened 2019-04-26
Add setting to disable registered error views for debugging in WSGI setups

### LINKED ISSUE #562: Make debugging under WSGI easier by allowing exceptions through
Right now if you have blanket exception views registered or if you have a `standard_html_error` (also a blanket exception view) all exceptions will stop there. This is fine and expected in production, but inconvenient for developers:

- errors are masked in the browser unless the error view exposes the traceback; which would be unusual because that's an information leak
- you cannot use debugging middleware like the nifty `werkzeug` debugger because the exceptions never get there
- potentially helpful `waitress` settings like `expose_tracebacks` become meaningless because, again, the exception never gets there

I propose to change the behavior for `debug_mode`. If an instance is running in debug mode  with `bin/runwsgi -d ...` or, in the case of `plone.recipe.zope2instance`, `bin/zopeinstance fg`, to allow exceptions to propagate and not look for/execute configured exception views.

dataflake:
Fixes #562 

Makes life much easier for developers, and enables us to use debugging and exception view tools that plug into the WSGI pipeline


d-maurer:
There should be a way to disable the "reraise" even in debug mode: The "reraise" is useful only when you are using specialized tools; other features controlled by debug mode are useful in general. There should be no implicit coupling.

dataflake:
The reraising is now done with a separate configuration setting.

d-maurer:
Jens Vagelpohl wrote at 2019-4-26 20:24 -0700:
>The reraising is now done with a separate configuration setting.

Thank you!

CHANGES.rst
```diff
@@ -50,6 +50,9 @@ Features
 Other changes
 +++++++++++++
 
+- Exceptions during publishing are now re-raised in a new exceptions debug mode
```
```suggestion
- Exceptions during publishing are now re-raised in a new exceptions debug mode,
  see also `documentation <https://zope.readthedocs.io/en/latest/wsgi.html#werkzeug>`_.
```

src/ZPublisher/tests/test_WSGIPublisher.py
```diff
@@ -631,6 +636,30 @@ def testHandleErrorsFalseBypassesExceptionResponse(self):
...
+            response.debug_exceptions = True
+            return response
+
+        # With debug_exceptions, the exception view is not called.
+        with self.assertRaises(BadRequest):
+            self._callFUT(environ, start_response, _publish,
+                          _response_factory=response_factory)
+
+        # Clean up view registration
+        unregisterExceptionView(IException)
```
icemac: I think there should be a `try-finally` block, so the `unregisterExceptionView` is even called when the tests might fail so it does not mess up other tests.

docs/wsgi.rst
```diff
@@ -210,6 +210,92 @@ section will pull in the correct dependencies:
...
+use debugging tools, you can start your Zope instance in `exceptions debug
+mode`. This will disable all registered exception views including
+``standard_error_message`` so that exceptions are not masked or hidden.
+
+This is how you run Zope in exceptions debug mode using the built-in
+``runwsgi`` script:
+
+.. code-block:: console
+
+   $ bin/runwsgi -ev etc/zope.ini
```
icemac: Is the `-v` actually needed? From reading the text I was thinking the new option is called `-ev`. Maybe `--debug-exceptions` could be used here for clarity.

CHANGES.rst
dataflake: Yes, good idea. I added the link to the documentation.

src/ZPublisher/tests/test_WSGIPublisher.py
dataflake: try/finally added

docs/wsgi.rst
dataflake: I added the `-v` purely out of habit. During development it's always `bin/runwsgi -dv etc/zope.ini`, and while I know what `-d` does, I don't even know if `-v` makes any difference. I dropped it in the document.

Now I can use `bin/runwsgi -dev etc/zope.ini`, which is kind of cool ;-)

PR #564, opened 2019-04-26
Revise subsection "Character Encodings for Arguments"

jugmac00:
Subsection has been almost rewritten completely, as nowadays "UTF-8" is the
de facto standard on the Internet and the publisher also uses
"UTF-8" as default encoding for Python 3.

Added links to further information on charsets, Python's supported
encodings and on the accept-charset attribute of the form tag.

**Note:**
As this subsection was almost completely rewritten, please take extra care when reviewing this pull request, especially the part about Python 3 and not needing to manually set an encoding for the argument conversion.

I both had a firm look at the source code and stepped through several examples, both on Python 2 and Python 3.


dataflake:
UTF-8 is the default for *both* Python 2 and Python 3 under Zope 4. It used to be Latin-1 under Zope 2.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -759,67 +759,58 @@ Character Encodings for Arguments
...
-   Every field needs that character encoding name appended to is
-   converter. The tag parser insists that tags must only use
-   alphanumberic characters or an underscore, so you might need to
-   use a short form of the encoding name from the Python 'encodings'
-   library package (such as utf8 rather than UTF-8).
+browser to encode the submitted form fields. In the past, this could
+have been a complicated topic.
+
+Nowadays, as UTF-8 is the de facto standard for encoding on the
+Internet, things became much simpler.
```
dataflake: ... **are** much simpler  (with "nowadays" you refer to the present, "became" refers to the past)

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -759,67 +759,58 @@ Character Encodings for Arguments
...
+
+**Best practice**
+
+If you are using Python 3 and set the the ``charset`` meta tag to
```
dataflake: I don't see many places, except maybe in our own auto-generated default content, where the meta charset tag is set by hand, or used at all. Browsers are supposed to go by the Content-type header which can contain charset information. If you set ``default-zpublisher-encoding`` in the Zope configuration then the Content-type header will reflect the encoding set there. The default is UTF-8.

dataflake: So you should emphasize the Zope configuration and de-emphasize the HTML tag. Right now it seems the HTML tag is really important, but it isn't.

jugmac00:
> UTF-8 is the default for _both_ Python 2 and Python 3 under Zope 4. It used to be Latin-1 under Zope 2.

Right - but one place where the publisher differentiates between Python 2 and 3 is when creating the ZopeFieldStorage - and there for Python 3 the publisher sets the encoding to self.charset which resolves to default_encoding => "utf-8" - actually I referred to this spot, which I assume is the reason that you do not have to use a `:utf8` converter for Python 3, but you do have to set it for Python 2.

https://github.com/zopefoundation/Zope/blob/056e4aa4a08c1838f06f06d513de180b6b8734c8/src/ZPublisher/HTTPRequest.py#L519-L524

So it maybe just better to leave out the "default encoding" info, and just say that for Python 2 you have to set the encoding converter, but not for Python 3.

dataflake:
Have you done some simple manual testing to prove this difference? I don't know off the top of my head if this is true.

dataflake:
Let me revise that: I don't know if that bit of code that sets the encoding specifically is needed or if it makes sense.

docs/zdgbook/ObjectPublishing.rst
jugmac00: Thank you - will fix it later.

docs/zdgbook/ObjectPublishing.rst
jugmac00: Thank you - actually this is the first time I heard of **default-zpublisher-encoding**.

Previously, I thought only
- meta tag charset
- the bom
- web server (ie Nginx) settings
- and the accept-charset attribute of the form tag can set/overwrite the encoding
:-)

Gonna rewrite the paragraph later.

jugmac00:
> Have you done some simple manual testing to prove this difference? I don't know off the top of my head if this is true.

=>

>and stepped through several examples, both on Python 2 and Python 3.

I am sure/convinced, that you do not need to use :utf8 for Python 3, but you need to set it for Python 2. I am not 100% sure why, but I'd bet on the ZopeFieldStorage creation.

I will manually step through some more examples and document them here.

PR #566, opened 2019-04-26
Revise subsection "Method arguments"

jugmac00:
- fixed minor grammar problems
- improved formatting
- deleted note about not to use more than one method field, as
it is a common use case to create forms with several input buttons,
each with a method argument

### DIFF
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -826,33 +826,31 @@ Method Arguments
 ----------------
 
 Sometimes you may wish to control which object is published based on
-form data. For example, you might want to have a form with a select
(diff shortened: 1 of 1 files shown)


docs/zdgbook/ObjectPublishing.rst
```diff
@@ -826,33 +826,31 @@ Method Arguments
 ----------------
 
 Sometimes you may wish to control which object is published based on
-form data. For example, you might want to have a form with a select
```
dataflake: You can leave the old line the way it is, it is correct

jugmac00: Thanks! Reverted it. Also the the other line in this subsection (which you reviewed in the other pull request, as I falsely pushed this one commit to both pull requests).

PR #572, opened 2019-04-28
Resurrect the Interfaces ZMI tab

### LINKED ISSUE #450: "Interfaces" tab gone?
Is there a reason why the Interfaces tab was removed from the Zope 4  ZMI.

dataflake:
Fixes #450


src/Products/Five/utilities/browser/marker.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
+        self.adapted = IMarkerInterfaces(context)
+        self.context_url = self.context.absolute_url()
+
+    def __call__(self, SAVE=None, add=(), remove=()):
+        if SAVE:
+            self.update(add, remove)
+            self.request.response.redirect(self.request.ACTUAL_URL)
```
icemac: Setting a `manage_tabs_message` here would be nice to hint the user that his save request was successful.

src/Products/Five/utilities/browser/marker.py
```diff
@@ -0,0 +1,61 @@
...
+            return ''
+        return self.index()
+
+    def _getLinkToInterfaceDetailsView(self, interfaceName):
+        return (self.context_url + (
+            '/views-details.html?iface=%s&type=zope.publisher.'
+            'interfaces.browser.IBrowserRequest' % interfaceName))
```
icemac: Where is this link used? I was not able to find it in the generated HTML code on `manage_interfaces` and there seems to be no view named `views-details.html` in Zope.

dataflake:
I simply copied the code from the moment it was removed by Hanno. I did not go over it with a fine tooth comb, I just made sure the ZMI part works as intended.

There was a test module, see https://github.com/zopefoundation/Zope/commit/c4c1b4acf51b9d6f2b699801373d750b50e74673. This test module is a doctest. I spent an hour trying to make it work but couldn't, due to the fact that doctests are utterly impossible to debug in a sane way. I'm not going to spend any more time on that module.

I'm not invested in this PR. This was just in reaction to complaints in #450 and the fact that several people replied wanting it back.

src/Products/Five/utilities/browser/marker.py
dataflake: You're right, I added that now.

src/Products/Five/utilities/browser/marker.py
dataflake: I removed that method now.

dataflake:
@icemac After converting the tests from &%$^ doctests I had a chance at debugging and fixing them, they are back and passing now. Take another look when you have a minute.

PR #578, opened 2019-05-01
[WIP]Revise section exceptions

jugmac00:
While revising the subsection **Exceptions** and verifying the documenation, I cleaned up HTTPResponse a bit.


d-maurer:
Jürgen Gmach wrote at 2019-4-30 23:46 -0700:
>While revising the subsection **Exceptions** and verifying the documenation, I cleaned up HTTPResponse a bit.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/zopefoundation/Zope/pull/578
> ...
>  * Simplify HTTPResponse.excecption

: 662 - tb = None
: 669 . tb = None

Resetting "tb" can be important to prevent circurlar references
(via tracebacks).

While Python now has a cyclic garbage collector and may recycle
cyclic structures (under some circumstances),
uncontrolled cycles (as those via tracebacks)
should still be avoided, as a single object with "__del__" may
prevent the collection.

I strongly recommend to keep the "tb = None" above -- even
though I did not check whether there is really a chance for
the creation of cyclic structures via a not reset "tb".

jugmac00:
@d-maurer Thanks a lot for that one. I saw the return statement right after the variable assignment and did not think there could be any valid reason for the reset.

You won't come to next weeks Zope sprint in Halle so I could buy you a beer?
https://blog.gocept.com/2019/03/07/zope-spring-cleaning-sprint/

d-maurer:
Jürgen Gmach wrote at 2019-5-1 01:38 -0700:
I saw the return statement right after the variable assignment and did not think there could be any valid reason for the reset.

And you were likely right - in the special case.

Those references (even indirect ones) to tracebacks are dangerous
only, when they can get themselves into a traceback (e.g.
caused by a following exception or a following raise/reraise).
There is no danger right before a return statement (provided
a potential return expression does not raise an exception).

Nevertheless, it is good practive to always reset variables
containing (direct or indirect) references to tracebacks -
as soon as they are no longer needed. This lets you be on the "safe side" -
whatever the following code might do.


I will not be in Halle.

PR #581, opened 2019-05-02
Update to secure version

icemac:
https://requires.io/github/zopefoundation/Zope/requirements/?branch=master indicates the used version as insecure.


dataflake:
@icemac when I did the dependency updates 5 days ago there was some reason why I could not update urllib3. I don't recall what that was, though. The tests are not failing, so that's fine.

@jugmac00 Sphinx cannot be updated just in those configuration files. The latest versions are Python 3-only.  What I have done in other packages (but not applicable here since the buildout is not allowed to just pick a version) is to remove the pin from the buildout config etc and specify it inside the docs setup.py extra where you can insert conditions based on Python version.

Example:

https://github.com/dataflake/dataflake.wsgi.bjoern/blob/fed7dfc20c3b75be3aeec9800a5fc0a2cfbeecd5/setup.py#L73-L74

dataflake:
@icemac I found what the issue was, there's a collision between the pins in the `requests` package and the new version:

```
<buildout blowup>
Version and requirements information containing urllib3:
  [versions] constraint on urllib3: 1.25.2
  Requirement of requests: urllib3<1.25,>=1.21.1
While:
  Installing scripts.
Error: The requirement ('urllib3<1.25,>=1.21.1') is not allowed by your [versions] constraint (1.25.2)
```

icemac:
This ist still the case even on master of requests, see https://github.com/kennethreitz/requests/blob/e73a6c52d205915c4d84e24ba01a5cd67e37d094/setup.py#L47

But we could upgrade to 1.24.3 which should fix the security issue: https://pypi.org/project/urllib3/1.24.3/

icemac:
Hm, it did not help to update to the version containing the fix: https://requires.io/github/zopefoundation/Zope/requirements/?branch=master still shows `urllib3` as insecure although the suggested version does not even contain the security fix. So I wrote to the support of requires.io.

dataflake:
It was worth a try, thanks for the research.

icemac:
It is now fixed on https://requires.io/github/zopefoundation/Zope/requirements/?branch=master after they updated their database.

Issue #435, opened 2018-12-18
Improper handling with non-ascii characters in response.redirect() in Zope 4/Python 3


zopyx:
Situation:
- user enters the name of form which is send as parameter `name:unicode` to Plone 5.2/Python 3
- the browser view receives the parameter as `str` and creates response.redirect to
  `self.context.absolute() + '/view' + name`. In my case this is `/Plone2/s3/view/ä` 
- the browser redirects to `/Plone2/s3/view/%E4`  and waitress throws this exception:

```
2018-12-18 11:37:44,424 ERROR [waitress:341][waitress] Exception when serving /Plone2/s3/view/ääää
Traceback (most recent call last):
  File "/home/ajung/src/xmldirector.connector/eggs/waitress-1.1.0-py3.7.egg/waitress/channel.py", line 338, in service
    task.service()
  File "/home/ajung/src/xmldirector.connector/eggs/waitress-1.1.0-py3.7.egg/waitress/task.py", line 169, in service
    self.execute()
  File "/home/ajung/src/xmldirector.connector/eggs/waitress-1.1.0-py3.7.egg/waitress/task.py", line 399, in execute
    app_iter = self.channel.server.application(env, start_response)
  File "/home/ajung/src/xmldirector.connector/eggs/Zope-4.0b7-py3.7.egg/ZPublisher/httpexceptions.py", line 30, in __call__
    return self.application(environ, start_response)
  File "/home/ajung/src/xmldirector.connector/eggs/Zope-4.0b7-py3.7.egg/ZPublisher/WSGIPublisher.py", line 251, in publish_module
    path_info = path_info.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 16: invalid continuation byte

```

My browser view can clearly deal with a request to `/Plone2/s3/view/ä`  - however something is going here with the unicode <-> utf8 conversion - either as part of the generation of the redirection URL and/or with improper `%e4` as part of the redirection URL.

zopyx:
I think that the redirect() call generates an improper redirection URL here:

The correct redirection URL for `/Plone2/s3/view/ä` must be `/Plone2/s3/view/%C3%A4` but not `/Plone2/s3/view/%E4`.

dataflake:
The `redirect` call does not seem to be the issue. It's how paths are handled in the WSGI publisher.

[...]

PR #586, opened 2019-05-03
Do not quote `@` on redirect.

icemac:
This is a regression introduced in zopefoundation/Zope#435.


d-maurer:
Jens Vagelpohl wrote at 2019-5-3 11:16 +0000:
>dataflake approved this pull request.
>
>I wasn't aware that quoting the `@@` would break those paths. Are there more of those special characters? Like e.g. `++resources++xxx`?

This may have some relation to our discussion about recoding of "PATH_INFO":
"PATH_INFO" may contain url-escaped characters (e.g.
resulting from quoting "@@") - and those character sequences need to
be url-decoded (then recoded).
The current problem might indicate that this url-decoding is missing.

dataflake:
This is indirectly related. Our earlier discussion was about URLs that come into Zope for processing. 
 This is about URLs that are generated by Zope during redirect handling.

What comes into `RESPONSE.redirect` is parsed and the path element is URL-quoted inside the redirect method to generate a correct and safe representation for the `Location` header. Apparently (I did not test this case) quoting view paths that have `@@` in them will result in an incorrect representation.

d-maurer:
Jens Vagelpohl wrote at 2019-5-3 09:22 -0700:
>This is indirectly related. Our earlier discussion was about URLs that come into Zope for processing.
> This is about URLs that are generated by Zope during redirect handling.

I wantet to stress, that,
apparently, the incoming side currently lacks a necessary "url-decode"
(otherwise, the (unnecessary) outgoing quoting of `@` would have been undone
on the incoming side).


>What comes into `RESPONSE.redirect` is parsed and the path element is URL-quoted inside the redirect method to generate a correct and safe representation for the `Location` header.

What is your motivation for this? The argument to `redirect` must
be an url. It should already have all necessary quoting.
Especially, it may contain url-encoded characters which must not
be encoded a second time.

dataflake:
The motivation is that we have URLs coming in from application code that are *not* quoted. And to make sure I don't quote twice I actually unquote incoming path elements.

Take a look for yourself at https://github.com/zopefoundation/Zope/blob/43e3337826fe84b902f0e6bac551cd643799b1fa/src/ZPublisher/HTTPResponse.py#L205

d-maurer:
Jens Vagelpohl wrote at 2019-5-3 11:26 -0700:
>The motivation is that we have URLs coming in from application code that are *not* quoted. And to make sure I don't quote twice I actually unquote incoming path elements.
>
>Take a look for yourself at https://github.com/zopefoundation/Zope/blob/43e3337826fe84b902f0e6bac551cd643799b1fa/src/ZPublisher/HTTPResponse.py#L205

In general, the approach is not correct: one must not url-decode
(nor url-encode) the complete path, as the decoding may expose characters with
a semantic meaning (e.g. "/" (path separator) or ";" (parameter separator)
or "?" (query string separator)).

Note that the redirect may go to another site. Thus, even though
Zope does not support "/" in ids nor uri parameters (introduced with ";"),
the `location` parameter to `redirect` may use those features
and "unquoting/quoting" may interfere in their use.


A safer approach for "non-ASCII id handling" would be to handle
the quoting of the *individual* ids elsewhere, e.g. in
`OFS.Traversable.Traversable.absolute_url`, potentially giving
this class an additional method `get_id_for_url` (for partial
url construction).
It could then become the responsibilty of the application to
provide a proper `location` value to `redirect`.

icemac:
Maybe we should assume that the `location` entring the `redirect` method is already correctly quoted. We could raise an exception if we get something which is not plausible as a correctly quoted location, e. g. a number or `None` or maybe `bytes`.

d-maurer:
Michael Howitz wrote at 2019-5-5 23:08 -0700:
>Maybe we should assume that the `location` entring the `redirect` method is already correctly quoted. We could raise an exception if we get something which is not plausible as a correctly quoted location, e. g. a number or `None` or maybe `bytes`.

When I remember correctly, then Python (3) allows "bytes" as urls
(conceptually, urls are indeed a sequence of bytes).

I am not sure, what the WSGI specification requires: it might
require that "bytes" are transformed into a "native string" (via
"latin-1" decoding for Python 3).

In any case, the WSGI use of "native string" and the associated restrictions
("str" restricted to the "latin-1" charset) for the description
of request/response metadata (everything except the body which is
"bytes") must be respected (at least for the WSGI publisher).

dataflake:
I think a discussion about what *should* be passed into the `redirect` method is not helpful here. The method is called by application code and in reality no one cares what URLs they create and pass along. The job of `redirect` is to basically take almost anything and try to repair it into usable shape.

icemac:
There might be additional characters like `?` and `;` which should not be quoted.
I am not sure how to handle `&`.

dataflake:
Take a look at the code, the quoting is only applied to the path element after sending the URL through urllib.parse. AFAIK at least `?` is illegal in path elements as it signifies the start of a query string.

d-maurer:
Jens Vagelpohl wrote at 2019-5-6 04:20 -0700:
>I think a discussion about what *should* be passed into the `redirect` method is not helpful here. The method is called by application code and in reality no one cares what URLs they create and pass along.

Even application code has responsibilities. I prefer to require a
greater effort from the application than do something wrong
in a library which the application has difficulty to work around:
the current code with its "unquote/quote" can be such a case -
it makes it e.g. very difficult for the application to redirect
to an url which contains an uri parameter (introduced with ";").

dataflake:
You're still going on about how you think it should work or how you would prefer it to work. Those are all valid points. But not helpful in this situation.

There's a huge body of code out there and we have to make an effort to support it. You can't just turn around and say "sorry, from now on your redirects must all conform to specification X".

If ';'  is important it can be added to the `safe` parameter of the quote function.

d-maurer:
Jens Vagelpohl wrote at 2019-5-6 09:28 -0700:
>You're still going on about how you think it should work or how you would prefer it to work. Those are all valid points. But not helpful in this situation.
>
>There's a huge body of code out there and we have to make an effort to support it.

The situation is not so grave as you paint it: most (all?) urls in existence
are ASCII-only. For those urls the "quote/unquote" is not necessary
but can introduce problems. Only non-ASCII urls can profit from
automatic "quote/unquote" -- they can come from new code - which could
be made responsible to adhere to restrictions; or from Zope ids.
Many issues with Zope ids can (and in my view should) be handled in general
classes, e.g. `OFS.Traversable.Traversable`.

I agree that there are likely some cases where non-ASCII ids can
require a new approach (such as a new method in `Traversable` to
get an id properly encoded for the use in an url path).
More than a decade ago, I have implemented a Zope extension
"arbitrary ids" (to support better WebDAV interaction by directly
supporting "normal" file system names). It was straight forward
to handle all problems related to those general ids - without the need
to massively change existing application code
(agreed, the Zope extension had no
need to be compatible with the complete existing Zope application code base,
just the one of my then employer).

d-maurer:
Jens Vagelpohl wrote at 2019-5-6 09:28 -0700:
> ...
>If ';'  is important it can be added to the `safe` parameter of the quote function.

From a formal point of view (with the stress on conformity to
relevant standards), a global "unquote + quote" cannot get it right:
if you want to support ";" as parameter separator, then there
are cases, where ";" should be used as "normal" character (and not
as parameter separator). In those cases, the ";" would be quoted;
but the "unquote" would destroy this information -- you lose
the possibility to support ";" in two different functions.

You have the same problem already now with "/" -- it will
always be treated as path separator -- even if it was quoted
in the input to `redirect` indicating that it should be treated as
"normal" character.


If you want to handle arbitrary input in `redirect`, you could approach
it as follows: do not use `unquote`; instead apply `quote` on the individual
character sequences invalid in urls (excluding sequences of allowed ASCII
characters - there will be some ambiguitiy for "%", however).
This way, portions of the input already representing url-escaped characters
would remain as they were. You could then use `quote` with
the separators in paths ("/" and ";") as `safe` and let the application
use pre url-encoding if they should not be handled in this way.

Alternatively, `redirect` could use special means to be informed
about the application's wish to fully control the url (i.e. to
deactivate any magic inside `redirect`) -- maybe, via a new parameter,
or interpreting a "bytes" input as "the application has already done
everything necessary".
I agree that it will be rare that the application needs that level
of control.

dataflake:
I would like to stop this discussion right now. Thank you.

dataflake:
Of course I broke down and played with this again. Dieter, if you're still interested, here's what I did:

First of all, instead of blindly unquoting I take a hacky guess if something has already been quoted by looking for `%`. If there's a `%` in the path component I do nothing at all.

Otherwise I apply quoting and pass all those characters as "safe" that either *should not* be escaped (https://tools.ietf.org/html/rfc2396 section 2.3):

```
Unreserved characters can be escaped without changing the semantics
   of the URI, but this should not be done unless the URI is being used
   in a context that does not allow the unescaped character to appear.
```

plus those that have semantic meaning within the path component (https://tools.ietf.org/html/rfc2396 section 3.3).

d-maurer:
Jens Vagelpohl wrote at 2019-5-6 16:32 -0700:
>Of course I broke down and played with this again. Dieter, if you're still interested, here's what I did:
>
>First of all, instead of blindly unquoting I take a hacky guess if something has already been quoted by looking for `%`. If there's a `%` in the path component I do nothing at all.
>
>Otherwise I apply quoting and pass all those characters as "safe" that either *should not* be escaped (https://tools.ietf.org/html/rfc2396 section 2.3):
>
>```
>Unreserved characters can be escaped without changing the semantics
>   of the URI, but this should not be done unless the URI is being used
>   in a context that does not allow the unescaped character to appear.
>```
>
>plus those that have semantic meaning within the path component (https://tools.ietf.org/html/rfc2396 section 3.3).

I feel good with this approach.

It might do something wrong (if the url contains `%`)
but in those rare cases, the application can work around this on its own.

PR #588, opened 2019-05-03
Extend zope-ecosystem.cfg by Products.ExternalMethod.

jugmac00:
... as opposed to PythonScripts, with ExternalMethod it is also possible
to set a breakpoint, raise exceptions or - more general - use not only
a restricted set of Python syntax.

This make debugging and setting up special cases, e.g. to validate
Zope behaviour for the documenation overhaul much easier.

### DIFF
--- a/zope-ecosystem.cfg
+++ b/zope-ecosystem.cfg
@@ -13,6 +13,7 @@ allow-picked-versions = true
 show-picked-versions = true
 additional_packages =
     Products.CMFCore
+    Products.ExternalMethod
     Products.MailHost
     Products.PythonScripts
     Products.Sessions


jugmac00:
Thanks for merging. I am aware of your concerns.

External methods are a perfect fit right now for quickly building a new Zope instance and injecting some Python code so I can verify certain documented behaviour, e.g. the **Exceptions section** of the **Object Publisher** by manually raising exceptions or setting a breakpoint - which is not possible with PythonScripts, as the Python syntax there is heavily restricted.

ExternalMethod makes this possible without the overhead of creating a custom product.

When I ever finish the ObjectPublishing chapter, the very next one is "Zope Products", which already describes how to create a product on the file system :-) as opposed to the Zope Book, which is more on the side of scripts.

dataflake:
Yeah, you found a perfectly good use for it.

By the way, while trying to pull together the various resources for www.zope.org I noticed there are a few lesser-known resources that you could combine with those parts of the Developer Guide that touch building a Zope Product. Take a look at https://www.zope.org/documentation/zope_secrets/startup.html, especially the two sections about products. A lot of that stuff is old, but much is still relevant.

d-maurer:
Jürgen Gmach wrote at 2019-5-5 14:55 -0700:
> ....
 by manually raising exceptions or setting a breakpoint - which is not possible with PythonScripts, as the Python syntax there is heavily restricted.

Note that is is possible to set (code) breakpoints in "PythonScripts".
For this, you must use `allow_module` (maybe differently spelled)
for the module `pdb`.
I think, it is also possible to raise exceptions in "PythonScripts".

PR #590, opened 2019-05-07
- Add a configuration flag to show bookmarkable URLs in the ZMI

### LINKED ISSUE #580: Make ZMI URL manipulation opt-in
Currently the ZMI uses a Javascript snippet in `manage_page_footer` to set the URL in the browser URL bar to the URL of the right-hand frame. Under Zope 2, the main URL was always stable unless you explicitly opened a frame separately. Frank has explained his reasoning and it makes sense for some use cases, but this issue is not about questioning that choice.

Today I found myself patching `manage_page_footer` for a customer so their experience moving from Zope 2 to Zope 4 will be smoother and they are not faced with unexpected behavior. They use the ZMI a lot.

I would really like that URL manipulation to be opt-in instead of the default. I'm not sure where or how to store such a preference. A new configuration key for the Zope configuration seems too heavy for this.

dataflake:
Fixes #580 

This patch exposes a new boolean Zope configuration flag named `zmi-bookmarkable-urls` and uses it to toggle the Javascript URL manipulation of the browser URL bar.


dataflake:
My decision was purely based on "what will give switchers the smoothest experience". This could be turned on by default for Zope 5 maybe?

icemac:
Zope 4 is already a big step from Zope 2.13. Adding this change (bookmarkable URLs) does not make it worse. Changing the default in Zope 5 is not very appealing to me.

dataflake:
Coming from Zope 2 the user experience with the "fake URLs" in the URL bar is very very bad. The browser is constantly lying about what the real URL is and as soon as you're trying to use browser elements that use the URL bar and its history, like reload or just going back, strange things happen. Boom, left menu gone, you've slipped out of the frameset.

The more I think about it the more I believe that approach of trying to keep the old frameset while trying to suggest you're not in the frameset with the fake URLs needs to be re-thought. The user experience is unexpected and bad.

What really needs to happen is this:

- get rid of frames entirely to get rid of the URL-faking
- the left-hand menu is entirely javascript driven

dataflake:
P.S.: I'll switch the default now after my local `tox` test has run through. I still think the whole approach is broken.

drfho:
@dataflake 
Hi, as a consequence of the `getBookmarkableURLs()`-call in 
`Zope\src\App\dtml\manage_page_footer.dtml`
 there might occur an error if the function is not available, eg. in ExternalMethod-objects. I'd like to suggest an easy fix with a try-nesting. Maybe you got a more profound solution?

```
<dtml-unless "REQUEST.get('zmi_dialog','window')=='modal'">
  <dtml-unless "'/manage_menu' in REQUEST.URL">
    <script>
      // Helpers for Menu Handling <dtml-var "REQUEST.URL">
      var manage_menu = typeof window.parent.frames!="undefined"&&typeof window.parent.frames.manage_menu!="undefined";
+     <dtml-try>
        <dtml-if "getBookmarkableURLs() and REQUEST.get('method','') == 'GET'">
          window.parent.history.replaceState('','Main','<dtml-var URL>')
        </dtml-if>
+       <dtml-except>
+     </dtml-try>
    </script>
  </dtml-unless>
  </body>
</html>
</dtml-unless>
```
Regards
f

dataflake:
@drfho You can have the same effect ("don't blow up if something doesn't exist") simpler, see e7192fbd55aa9120b0c5b425e71034fa53bf01c0

Why the External Method object cannot acquire the method from its container like every other object is a riddle to me.

drfho:
@dataflake Works perfect, thanks!

PR #593, opened 2019-05-08
Fixes #575 by improved documentation.

### LINKED ISSUE #575: Nonsensical PATH_INFO handling in WSGIPublisher.publish_module
The code in ``publish_module`` that wants to clean up ``environ['PATH_INFO']`` is broken under Python 3.

- it assumes PATH_INFO is an unencoded string (is this always true under Python3? I don't know.)
- it encodes PATH_INFO in `Latin-1` in one line and decodes it from `UTF-8` in the next line. That's clearly broken.

Commits 6c15c71 by @sallner and 707df73 by @icemac left the code in this state. Please look this over and fix it.

sallner:
Please have a look, if it is BIG enough :wink: 

Fixes #575.


src/ZPublisher/WSGIPublisher.py
```diff
@@ -283,10 +283,16 @@ def publish_module(environ, start_response,
...
-        # again with latin-1 to get a utf-8 encoded bytestring.
+        # BIG Comment, see discussion at
+        # https://github.com/zopefoundation/Zope/issues/575
+        #
+        # The WSGI server automatically treats headers, including the
+        # PATH_INFO, as latin-1 encoded bytestrings, according to PEP-3333.
+        # Typically this is a false assumption nowadays as the browser delivers
+        # utf-8 url-encoded PATH_INFO, like '/t%C3%A4st', resulting in '/tÃ¤st'
+        # after latin-1 decode by the wsgi-server at this point. We, therefore,
+        # need to encode it again with latin-1 to get a utf-8 encoded
```
dataflake: This sentence makes no sense. Encoding with latin-1 gives you a latin-1 encoded byte string.

src/ZPublisher/WSGIPublisher.py
```diff
@@ -283,10 +283,16 @@ def publish_module(environ, start_response,
...
+        # bytestring.
         path_info = path_info.encode('latin-1')
         # But in Python 3 we need text here, so we decode the bytestring.
         path_info = path_info.decode('utf-8')
```
dataflake: This is actually the head scratcher for me and needs commenting. The decode right after encode, which seems like a non-op, but with the added danger of using a different encoding. Code points that exist in latin-1 but not in UTF-8 make this fail.

d-maurer:
Jens Vagelpohl wrote at 2019-5-8 07:55 -0700:
> ...
>+        # The WSGI server automatically treats headers, including the

>+        # PATH_INFO, as latin-1 encoded bytestrings, according to PEP-3333.

>+        # Typically this is a false assumption nowadays as the browser delivers

>+        # utf-8 url-encoded PATH_INFO

No!

WSGI delivers "latin-1" encoded bytestrings, because that is the neutral, zero
loss transformation from `bytes` to `str`; it works for any
encoding and lets later steps use their knowledge about the right
encoding to interpret the information.

In addition, "latin-1" is also the HTTP 1.1 charset for all HTTP request
and response headers. If a header needs something else, it must
use RFC 2047 encoding. 

The url parts are special -- they must be pure ASCII; anything
non-ASCII "pct" encoded.

src/ZPublisher/WSGIPublisher.py
sallner: not if the original byte string was something else. `b'/t\xc3\xa4st'.decode('latin-1').encode('latin-1')` is still a utf-8 encoded byte string.

src/ZPublisher/WSGIPublisher.py
sallner: I try to clarify this further with the description of how the `PATH_INFO` was derived up to this point.

src/ZPublisher/WSGIPublisher.py
dataflake: That claim is only true for those unicode code points shared by latin-1 and utf-8. If there is any mentioning in PEP-3333 that you will only ever get those code points in the WSGI pipeline please mention that. Otherwise the general claim is wrong and confusing (to me at least).

sallner:
> No! WSGI delivers "latin-1" encoded bytestrings, because that is the neutral, zero loss transformation from `bytes` to `str`; it works for any encoding and lets later steps use their knowledge about the right encoding to interpret the information. In addition, "latin-1" is also the HTTP 1.1 charset for all HTTP request and response headers. If a header needs something else, it must use RFC 2047 encoding. The url parts are special -- they must be pure ASCII; anything non-ASCII "pct" encoded.

That is true, I tried to capture that in a language, which is less diminishing of latin-1 than before.  When traversing to object-ids with unicode characters, we use the pct-encoding to render the URIs in Zope, so this also the way we can safely handle the traversal here.

d-maurer:
Jens Vagelpohl wrote at 2019-5-8 16:02 +0000:
> ...
>That claim is only true for those unicode code points shared by latin-1 and utf-8. If there is any mentioning in PEP-3333

PEP-3333 states that it is using "native string" for HTTP request and
response metadata. This includes everything in a request/response
apart from the body. The body is a sequence of bytes.

It explains that for Python 3, "native string" is in fact
a "latin-1" decoded byte string: the "str" must not contain
any codepoint above 255.
It explains that it uses `str` rather than `bytes` just for
convenience.


This does not mean that the byte string underlying
a "native string" really represents
a "latin-1" encoded text. The "latin-1" decoding is a purely
technical means to treat a sequence of bytes as an `str` -- for
convenience purposes: `str` is much better supported than `bytes`.

The bytes string underlying an url (and their parts)
is really a (restricted) ASCII sequence (everything non-ASCII was PCT encoded).
For them, "latin-1" and "utf-8" are equivalent.


Applied to the individual segments of `PATH_INFO`, `urllib.parse.unquote`
automatically performs the necessary recoding
(effectively, though not literally: `.encode("latin-1").decode("utf-8")`)
and delivers the correctly decoded id (of type `str`)
corresponding to the segment.

d-maurer:
Steffen Allner wrote at 2019-5-8 09:03 -0700:
> ...
>That is true, I tried to capture that in a language, which is less diminishing of latin-1 than before.  When traversing to object-ids with unicode characters, we use the pct-encoding to render the URIs in Zope, so this also the way we can safely handle the traversal here.

Right.

dataflake:
Thanks to the information from @d-maurer I have expanded the comment a little more. Let me know if that's OK.

sallner:
> Thanks to the information from @d-maurer I have expanded the comment a little more. Let me know if that's OK.

I have the impression that your expansion works not so well for my brain, but that was the same with my initial description. So let's keep it and I will merge it later, if @d-maurer has no other blocking concern.

d-maurer:
Steffen Allner wrote at 2019-5-9 07:59 +0000:
>> Thanks to the information from @d-maurer I have expanded the comment a little more. Let me know if that's OK.
>
>I have the impression that your expansion works not so well for my brain, but that was the same with my initial description. So let's keep it and I will merge it later, if @d-maurer has no other blocking concern.

I do not have a blocking concern: we speak about an (effective) no-op
-- and I was highly surprised to have seen the (primary) issue (#575)
classified as "release blocking".

I am still convinced that the WSGI publisher should not do
the recoding - and that the problem with the failing "non-ascii-url"
test should be tackled elsewhere (likely in "webOb") -
but this is cosmetics only and far from "blocking".



--
Dieter

sallner:
In I think the label "release blocking" was mainly introduced, as it was not clear, whether a deletion of this code branch turns out to break something substantial or not. As it turns out with the `test_can_handle_non_ascii_URLs` (which actually tests the handling of non-ascii _object_ids_) it would break something in the current case. Your concern with the potentially present `/` or `%2F` is mitigated in a way, that `/` are not allowed in _object_ids_ in Zope, so if there is a slash in the path, it should be treated as path segment delimiter.

d-maurer:
Steffen Allner wrote at 2019-5-9 02:43 -0700:
>In I think the label "release blocking" was mainly introduced, as it was not clear, whether a deletion of this code branch turns out to break something substantial or not. As it turns out with the `test_can_handle_non_ascii_URLs` (which actually tests the handling of non-ascii _object_ids_) it would break something in the current case.

It broke a test -- not a "real" use.

I am quite confident that the breakage is caused by a buggy test
component (but have not yet investigated).

d-maurer:
Dieter Maurer wrote at 2019-5-9 12:07 +0200:
>Steffen Allner wrote at 2019-5-9 02:43 -0700:
>>In I think the label "release blocking" was mainly introduced, as it was not clear, whether a deletion of this code branch turns out to break something substantial or not. As it turns out with the `test_can_handle_non_ascii_URLs` (which actually tests the handling of non-ascii _object_ids_) it would break something in the current case.
>
>It broke a test -- not a "real" use.
>
>I am quite confident that the breakage is caused by a buggy test
>component (but have not yet investigated).

I was wrong.

I have commented out the recoding and a "real" use broke.
This means that at this point `environ["PATH_INFO"]` is
no longer ASCII only (as it should be).

For the limited Zope application (no "/" in Ids; no support
for url paramters (introduced by `;`)), this may be acceptable.
But, it means that the WSGI frontend heavily depends on those
restrictions -- without any need.

PR #596, opened 2019-05-08
Revise section "Exceptions".

jugmac00:
Beginning with Zope 4, the processing of exceptions has been changed
entirely, as Zope 4 come with the *waitress* WSGI server by default.

**Note about line length**
ReStructered Text does not allow a line breakn with a bullet list
=> line length exceeded in favor of readabilty in rendered document.


dataflake:
Your first comment here is wrong. The exception handling has nothing to do with `waitress`. It's different because the (default) publisher switched to WSGI.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,46 @@ button will also deactivate all other radio buttons from the other records.
...
-2. If the exception type is "No Content", then no body will be
-   returned.
-
-When a body is returned, traceback information will be included in a
-comment in the output.
+When the object publisher catches an unhandled exception, it tries to
+match it with a set of predifined exceptions coming from the
+Products.zExceptions package, such as **HTTPNoContent**, **HTTPNotFound**,
+**HTTPUnauthorized** and many, many more. If there is a match, the custom
+exception gets upgraded into an zException, which then results in a proper
```
dataflake: upgraded **to a** zException

Instead of punting by just saying "many, many more" I would list them like the old documentation did.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,46 @@ button will also deactivate all other radio buttons from the other records.
...
+Zope comes with **waitress** and by default **waitress** returns an error
+message as follows::
+
+  Internal Server Error
+
+  The server encountered an unexpected internal server error
+
+  (generated by waitress)
+
+.. note::
```
dataflake: Instead of listing particulars here just point to this: https://zope.readthedocs.io/en/latest/wsgi.html#debugging-zope-applications-under-wsgi

jugmac00:
> Your first comment here is wrong. The exception handling has nothing to do with `waitress`. It's different because the (default) publisher switched to WSGI.

You mean in the commit message?

Gonna fix it later - we are heading to the pub in a second.

jugmac00:
> Your first comment here is wrong. The exception handling has nothing to do with `waitress`. It's different because the (default) publisher switched to WSGI.

Thanks for pointing out - updated the commit message.

docs/zdgbook/ObjectPublishing.rst
jugmac00: I removed the "many, many more" phrase and added a link to the package for a full list of exceptions..

The old documentation included more, but also only a partial list of all the possible exceptions.

Plus, if there will be a change to the list of exceptions of the zExceptions package (which probably won't ever happen), nobody will think about updating the Zope developer guide.

Also, if I really want to implement a custom exception I am probably more interested to choose the right status code. Choosing the name is only a way to get there. Either way, I'd have to lookup the source code of zExceptions anyway.

docs/zdgbook/ObjectPublishing.rst
jugmac00: Right! Added a link to the **debugging Zope applications under WSGI** documentation. Btw - thanks a lot for writing it! There was really a need for it.

jugmac00:
> In my Zope 4 installation, the "zExceptions" package is `zExceptions`, not `Products.zExceptions`.

Good catch!

> I suggest to drop the `and many, many more`.

Dropped it.


> I am fine with just giving a few `zException` examples (rather than an exhaustive enumeration): the upgrading seems to me an implementation detail which associates the proper HTTP status code (and message) with the exception. Few people will be interested in the details - and then, they will need to look at the code anyway.

I agree with that and updated the text accordingly, ie kept the few examples and added a link to the source code of zExceptions.


> I suggest to replace `the custom exception gets upgraded into a zException` by `the exception gets upgraded into the matching zException`.

Indeed - sounds better.

Thank you @d-maurer !

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
-   returned.
-
-When a body is returned, traceback information will be included in a
-comment in the output.
+When the object publisher catches an unhandled exception, it tries to
+match it with a set of predifined exceptions coming from the
+**zExceptions** package, such as **HTTPNoContent**, **HTTPNotFound**,
+**HTTPUnauthorized**.
+
+If there is a match, the exception gets upgraded into the matching
```
dataflake: upgraded **to**

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+    a ``standard_error_message``.
+
+    There are two ways to catch and render an exception:
+
+    - create a ``standard_error_message``, which can be a **DTML Method**, **DTML Document**, **Python Script** or **Page Template**
+    - create an ``exception view``, see blog post `Catching and rendering exceptions <https://blog.gocept.com/2017/10/24/zope4-errorhandling/>`_
+
+If the exception is not handled, it travels up the WSGI stack.
+
+What then happens depends entirely on the WSGI middleware. By default
```
dataflake: **happens then** instead of then happens

docs/zdgbook/ObjectPublishing.rst
jugmac00: Thanks for taking care even of those details!

docs/zdgbook/ObjectPublishing.rst
jugmac00: Thanks for taking care even of those details!

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
-When a body is returned, traceback information will be included in a
-comment in the output.
+When the object publisher catches an unhandled exception, it tries to
+match it with a set of predifined exceptions coming from the
+**zExceptions** package, such as **HTTPNoContent**, **HTTPNotFound**,
+**HTTPUnauthorized**.
+
+If there is a match, the exception gets upgraded to the matching
+**zException**, which then results in a proper response returned to the
+browser, including a HTTP status code.
```
d-maurer: browser, including an appropriate HTTP status code.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+     For a full list of exceptions please directly refer to the
+     implemented exception classes within the
+     `zExceptions package
+     <https://github.com/zopefoundation/zExceptions/blob/master/src/zExceptions/__init__.py>`_.
+
+
+.. attention::
+
+    When you create a custom exception, please make sure not to inherit
+    from **BaseException**, but e.g. from **Exception**, otherwise you'll
```
icemac:
```suggestion
    from **BaseException**, but from **Exception** or one of its child classes, otherwise you'll
```

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+    run into an waitress exception.
```
icemac:
```suggestion
    run into an exception in waitress.
```

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+
+.. note::
+
+    Beginning with Zope 4, a standard installation no longer comes with
+    a ``standard_error_message``.
+
+    There are two ways to catch and render an exception:
+
+    - create a ``standard_error_message``, which can be a **DTML Method**, **DTML Document**, **Python Script** or **Page Template**
```
icemac:
```suggestion
    - create a ``standard_error_message``, which can be a **DTML Method**, **DTML Document**, **Script (Python)** or **Page Template**
```

Changed to the name used in the add types menu in the ZMI.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+    - create an ``exception view``, see blog post `Catching and rendering exceptions <https://blog.gocept.com/2017/10/24/zope4-errorhandling/>`_
+
+If the exception is not handled, it travels up the WSGI stack.
+
+What happens then depends entirely on the WSGI middleware. By default
```
icemac:
```suggestion
What happens then depends entirely on the possibly installed WSGI middleware or the WSGI server. By default
```

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -926,42 +926,56 @@ button will also deactivate all other radio buttons from the other records.
...
+Zope comes with **waitress** and by default **waitress** returns an error
```
icemac:
```suggestion
Zope by default uses **waitress** which returns an error
```

PR #597, opened 2019-05-08
Change generated WSGI confinguation names

### LINKED ISSUE #571: A couple of questions about default-zpublisher-encoding
When revising #564 @dataflake made me aware of the `default-zpublisher-encoding` option.

I have a couple of questions /annotations about that one.

1. The only documentation about that option seems to be the CHANGES.rst and the NEWS.rst.

Both refer to zope.conf.

I installed Zope via
`$ bin/buildout -c zope-ecosystem.cfg`
and under `etc/` there is only a wsgi.conf and a zope.ini, no zope.conf.

Afaik `zope.conf` came with ZServer, which is no longer the default, and zope.ini / wsgi.conf comes with the new wsgi way.

If that's correct, we should update the news.rst.

2. Can anybody please explain what is the "wsgischema.xml" for?

https://github.com/zopefoundation/Zope/blob/9b2d948b2072a0c4cdc2220fa33b09b7f3bf79e4/src/Zope2/Startup/wsgischema.xml#L266-L273

Is this a config file or only documentation?

I'd assume it's a config file, as it gets loaded here..
https://github.com/zopefoundation/Zope/blob/882629f3deb0a188e1be548f7da9bbcd39e4862d/src/Zope2/Startup/options.py#L52-L60

3. Wrong docstring

Having a look at...

https://github.com/zopefoundation/Zope/blob/882629f3deb0a188e1be548f7da9bbcd39e4862d/src/Products/PageTemplates/unicodeconflictresolver.py#L47-L69

I'd think the docstring is wrong.

a) the encoding is not used from "`default-zpublisher-encoding`" but from 
"`ZPublisher.HTTPRequest.default_encoding`" which is hardcoded to "`utf-8`."

https://github.com/zopefoundation/Zope/blob/882629f3deb0a188e1be548f7da9bbcd39e4862d/src/ZPublisher/HTTPRequest.py#L55-L56

b) the fallback is "`iso-8859-15`", and not "`sys.getdefaultencoding`".

c) Does the class even get used?

The class gets used at the bottom of the module
https://github.com/zopefoundation/Zope/blob/882629f3deb0a188e1be548f7da9bbcd39e4862d/src/Products/PageTemplates/unicodeconflictresolver.py#L127-L132

but these new identifiers are only used for tests.

So it looks like the code could be removed.

I have done a quick search on github - no results except the tests. Or is that a public API we'd have to deprecate?

dataflake:
Fixes #571 

To stay in line with existing documentation that normally uses "zope.conf" for the Zope configuration, this PR changes the name of the configurations generated by `mkwsgiinstance`. The Zope configuration will get the old name "zope.conf", and the WSGI configuration is named "zopewsgi.ini".

I also fixed one docstring pointed out by Jürgen.


dataflake:
I have changed `zopewsgi.ini` to `zope.ini` now.

Issue #591, opened 2019-05-08
Zope allows ids to start with `@@` but cannot publish corresponding objects ...


d-maurer:
... because on traversal, it interprets the `@@` as view indicator.

Same problem with other id structures which can be misinterpreted as view indicators, such
as `++namespace++name`.

We likely should forbid ids which could be mistaken as view indicators.
For publishing itself, we could use (url-) quoting to distinguish the use of `@` as part of an id from that as view indicator. However, this would not fit well with programmatic traversal.

d-maurer:
An alternative to forbidding affected ids could be to specify that one interpretation has priority over the other interpretation, e.g. first try to interpret a name as a view and if this fails try to interpret it as an id. Such an approach is already used for names not starting with `@` or `+`.

dataflake:
Fixed in fe0bc43c142d152e304d0a2fde255841cb1dafcc

PR #598, opened 2019-05-08
Prevent the creation of object IDs that look like views

dataflake:
The change disallows objects with IDs that start with either `@@` or `++` so users cannot create something that will then be unreachable because Zope tries to look up a view with that name.


dataflake:
Fixes #591

src/OFS/ObjectManager.py
```diff
@@ -125,6 +125,10 @@ def checkValidId(self, id, allow_dup=0):
         raise BadRequest(
             'The id "%s" is invalid because it '
             'ends with two underscores.' % id)
+    if id.startswith('@@') or id.startswith('++'):
+        raise BadRequest(
+            'The id "%s" is invalid because it starts with characters '
+            'reserved for Zope views lookup.' % id)
```
icemac:
```suggestion
            "reserved for Zope's view lookup." % id)
```

dataflake: Das ist eine Verschlimmbesserung, geht nicht.

icemac: Sorry, maybe I did not understand the intent of the sentence of the exception.

dataflake: It's a matter of better English

PR #604, opened 2019-05-09
Add warning box about know issue with records

### LINKED ISSUE #568: Wrong result with records and checkboxes
When revising the "records arguments" section of the Zope developer book #567, I was reminded of an old bug, that bit me a couple of years ago.

Consider a form like this:

```
<form action="records_parse">

    <p>
    <input type="checkbox" name="index.enabled:records" value="0" />
    <input type="text" name="index.name:records" value="index 1" />
    <p>
    <input type="checkbox" name="index.enabled:records" value="1" checked="checked" />
    <input type="text" name="index.name:records" value="index 2" />
    <p>
    <input type="submit" name="submit" value="send" />

</form>
```
You'd expect a result as:
```
[{'name':'index 1'}, {'enabled':'1', 'name':'index 2'}]
```

But you get this
```
[{'enabled': '1', 'name': 'index 1'}, {'name': 'index 2'}]
```

So essentially the checkbox slips in the wrong list element.

This bug was reported already in 2010 by @petschki (hi!) on launchpad, and seemingly was forgotten to move to Github:
https://bugs.launchpad.net/zope2/+bug/619081

I can confirm that this bug is still present in the latest head of Zope 4.

As far as I understand the current test suite does not cover this bug as the test suite does not take the input types into account, and this bug only occurs with checkboxes.

As a workaround I used a hidden input field as the first input field as then the bug does not occour, in other words the bug only occurs, when the empty checkbox is the first input field of the form.

jugmac00:
When using a checkbox as the first form field, the object publisher
may not be able to match checked checkboxes with the correct record.

As this behaviour cannot be fixed, a warning box in the documentation
was added.

closes #568


docs/zdgbook/ObjectPublishing.rst
```diff
@@ -918,10 +918,43 @@ Record marshalling provides you with the ability to create complex
...
+  Records do not work with input fields of type radio as you might
+  expect, as all radio fields with the same name are considered as one
+  group - even if they are in different records. That means, activating
+  one radio button will also deactivate all other radio buttons from
+  the other records.
+
+.. attention::
+
+    When using records, please note that there is a known issue when
+    you use a form, where checkboxes are used in the first "column".
```
dataflake: take out the comma please

d-maurer:
Jens Vagelpohl wrote at 2019-5-9 05:33 -0700:
>dataflake approved this pull request.
> ...
>> -Please note, that records do not work with input fields of type radio as you
>-might expect, as all radio fields with the same name are considered as one

Note: I have not tested it, but I suspect that
`:records` (and maybe other sequence type
suffixes) might lead to surprises with the use of the ":default" suffix.

The purpose of the ":default" suffix it to provide a default value
which can later be overwritten by the the real value.

It is interesting to ask how this is interpreted (by the current
implementation) for cases where a new component is started
when a value is already present (as for `records`).

In my opinion, an existing default value should get overridden (and
no new component started) and a new default value should (potentially)
start a new component. If the current implementation does this,
then `checkboxes` (and other inputs for which the browser does not
reliably pass on values) could be handled by the creation
of an appropriate "hidden" input defining the default value.

jugmac00:
@d-maurer

As this pr is already merged, would you please create a new issue so your information does not get lost?

dataflake:
@jugmac00 please don't treat the issue tracker as a scratchpad or list of reminders to yourself. That's why I closed Dieter's issue.

dataflake:
@jugmac00 sorry, false alarm, I didn't realize #618 was a standalone issue and not just a followup to an existing issue that should have been handled in there.

Issue #463, opened 2019-01-26
Improved "Testing" support for Python 2/3 compatible doctests


d-maurer:
Working on [https://github.com/zopefoundation/five.formlib/pull/7](https://github.com/zopefoundation/five.formlib/pull/7), I noticed that it is quite difficult to write Python 2/3 compatible doctests.

I noticed the following problem areas:
- doctest checked unicode return values
- different input value types for `http` and `browser`
- necessity to have a correct "Content-Length" header for Python 3

To handle doctest checked unicode return values, it would be good to get a "doctest" option which allows to ignore the "u" string literal prefix in Python 3 for the comparison. In fact, this would best be in Python's "doctest" itself.

It would be good, if `Testing` would provide a function which converts a unicode value to the correct input type for `http` and the `browser` related methods.

It would be good, if `Testing.ZopeTestCase.zopedoctest.functional.http` would get a higher level companion which (among others) computes "Content-Length" correctly. Especially due to above point (different input types for `http` - which sometimes leads to a different length), it is very difficult to get a correct cross Python version "Content-Length" header for some `http` calls (those involving not ASCII input values).

ale-rt:
Hi @d-maurer, usually this is done defining a custom checker, look for "class Py23DocChecker". here some examples I found from a quick search but there are more:
- https://github.com/plone/plone.behavior/blob/master/plone/behavior/tests.py
- https://github.com/plone/plone.browserlayer/blob/master/plone/browserlayer/tests/test_doctest.py
- https://github.com/plone/plone.namedfile/blob/master/plone/namedfile/tests/test_doctests.py

d-maurer:
@ale-rt Thank you for your comment!
Obviously, it addresses the first point ("doctest checked unicode values"). **If** Zope plans to support both Python 2 and 3 for some time, then it should contain such a `Py23DocChecker` - avoiding the need to reinvent something like this in many packages.

dataflake:
A personal observation: among those people who are currently maintaining Zope: I see no one writing doctests, no enthusiasm for doctests and thus no enthusiasm for improving doctest support in Zope itself.

icemac:
The plans currently are to start working on Zope 5 (aka dropping Python 2 support) right after the Zope 4 final release is out of the door. As Python 2 has its sunset by the end of this year I don't see a big value in adding code to ease its support.

PR #610, opened 2019-05-10
Work on fixing `allowed_attributes` and `allowed_interface` on pages

### LINKED ISSUE #397: BrowserViews implement IPublishTraverse in Zope 4 which breaks allowed_attributes and allowed_interface

Example:

```xml
  <browser:page
      name="foo"
      for="*"
      class=".dummy.Foo"
      permission="zope2.View"
      allowed_attributes="bar"
      />
```
dummy.py:

```python

class Foo(BrowserView):

    def __call__(self):
        return 'foo'

    def bar(self):
        """bar"""
        return 'foobar'
```

```python
else:
    if adapter is None:

```

```
 <class 'object'>)
```

And like this after it:
```
 <class 'object'>)
```

dataflake:
Fixes #397 

This is work in progress so this is a "draft" PR that cannot be merged.

Open issues:

- [x] ~~Move added implementation of `publishTraverse` to an adapter (as requested by @jaroel)~~ does not help, the problem is that authentication is done _after_ traversal
- [x] Convert newly written doctest to a unittest.
- [x] Add a similar test for the behaviour when accessing a not published attribute (aka `mouse` instead of `eagle`).
- [ ] Implement a pair of tests using `allowed_interface`.


icemac:
@dataflake What do you think about the implementation I did together with @sallner?

dataflake:
@icemac See that lone failure on Python 2 because it renders as `u'The eagle has landed'`? Is there some doctest way of fixing it? The only reason I did that as doctests, which I otherwise hate, was because all other examples there use it.

@pbauer Can you run the whole Plone unittest battery against this branch? I am sure Plone would exercise all of this a lot more.

I'm fine with any solution that makes all tests succeed. I do want to fix up that non-running doctest that uses PythonScripts, though.

sallner:
After a lengthy discussion over a good lunch, we decided that a 404 should be the correct response in case an attribute is not allowed. This will be resolved after we decided on a good implementation in an upcoming bugfix release.

dataflake:
I'm fine with pushing this out a little bit, but I'm sure the Plone guys will be unhappy that they basically cannot use 4.0 until that bugfix is out, right?

sallner:
@jensens assured that there are workarounds for all relevant parts and therefore no blocker.

pbauer:
Plone itself does not use this feature a lot but there are some use-cases that are broken. Among them is enabling ical-upload to a folder: There is a action that enables this with the url:
http://localhost:8080/Plone/folder/@@ical_import_settings/enable

Here ist the registration:
```xml
    <browser:page
        for="plone.folder.interfaces.IFolder"
        name="ical_import_settings"
        class=".importer.IcalendarImportSettingsFormView"
        allowed_attributes="enable
                            disable"
        permission="plone.app.event.ImportIcal"
        layer="..interfaces.IBrowserLayer"
        />
```
In 5.2 this feature is broken so far (it has no tests that use the action...). 
With this branch the feature works.

pbauer:
So: There are cases without workarounds. There are a lot more cases where `allowed_attributes` and `allowed_interfaces` appears in Plone's zcml and where features might be broken. Sadly I'm ill so I'm not much help 😞

dataflake:
@icemac @sallner @pbauer @jaroel Current status: Tests are now passing.

View methods that are not in `allowed_attributes` cause a 401 now instead of the previous 404.

There's no code in Zope itself that uses `allowed_interfaces` so I can't be sure if that's fixed.

d-maurer:
Jens Vagelpohl wrote at 2019-5-10 04:26 -0700:
>@icemac See that lone failure on Python 2 because it renders as `u'The eagle has landed'`? Is there some doctest way of fixing it? The only reason I did that as doctests, which I otherwise hate, was because all other examples there use it.

My question about Python2+3 doctest support was closed, because
Zope plans to drop Python 2 support in the near future.
Thus, maybe, you would be content with a workaround.

For ASCII strings (as in your case), you can replace
the output via an equality test, e.g. instead of
```
	>>> f()
	"<some ASCII text>"
```
you can use in your doctest the equivalent
```
        >>> f() == "<some ASCII text>"
	True
```
This uses that Python2 automatically converts `str` to `unicode`
when the two come together.


Alternatively, you could set up a "custom checker". You find
details in @ale-rt 's comment to #463.

dataflake:
@d-maurer The only reason I used a doctest here was to make it easier to copy existing tests. I worked around the test failure and it's fixed now.

icemac:
After implementing a version which does not lead to HTTP-401 on attributes which are not allowed to be accessed, I added a list of open issues in the description of this PR. I can fix them myself, but I am not sure when I'll have the necessary time to do so.

jaroel:
This should work for the problem at hand, though we no longer can provide our own `IPublishTraverse` implementation - as the publisher checks if `IPublishTraverse` is already provided?

note: I can see why a 401 when accessing attributes that aren't in allowed_attributes or allowed_interfaces would make sense - though that's debatable of course.

icemac:
@jaroel wrote:
> This should work for the problem at hand, though we no longer can provide our own `IPublishTraverse` implementation - as the publisher checks if `IPublishTraverse` is already provided?

I see your point. Moving the implementation to an adapter should fix it. I added it to the list of open issues in the description of the PR.

> note: I can see why a 401 when accessing attributes that aren't in allowed_attributes or allowed_interfaces would make sense - though that's debatable of course.

If the attribute is neither in `allowed_attributes` nor in `allowed_interfaces` it could be considered as not published aka not existing for the browser thus leading to HTTP-404. (That's what we have discussed last week on the Zope sprint in Halle.)

pbauer:
With https://github.com/plone/plone.app.event/pull/306 this still results in a 404:

```
Error in test test_enable_ical_import (plone.app.event.tests.test_ical_import.TestICALImportSettings)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 615, in run
    testMethod()
  File "/Users/pbauer/workspace/plip/src/plone.app.event/plone/app/event/tests/test_ical_import.py", line 58, in test_enable_ical_import
    self.browser.open(f1_url + '/ical_import_settings/enable')
  File "/Users/pbauer/.cache/buildout/eggs/zope.testbrowser-5.3.2-py3.7.egg/zope/testbrowser/browser.py", line 251, in open
    self._processRequest(url, make_request)
  File "/Users/pbauer/.cache/buildout/eggs/zope.testbrowser-5.3.2-py3.7.egg/zope/testbrowser/browser.py", line 275, in _processRequest
    resp = make_request(reqargs)
  File "/Users/pbauer/.cache/buildout/eggs/zope.testbrowser-5.3.2-py3.7.egg/zope/testbrowser/browser.py", line 249, in make_request
    return self.testapp.get(url, **args)
  File "/Users/pbauer/.cache/buildout/eggs/WebTest-2.0.33-py3.7.egg/webtest/app.py", line 335, in get
    expect_errors=expect_errors)
  File "/Users/pbauer/.cache/buildout/eggs/zope.testbrowser-5.3.2-py3.7.egg/zope/testbrowser/browser.py", line 94, in do_request
    expect_errors)
  File "/Users/pbauer/.cache/buildout/eggs/WebTest-2.0.33-py3.7.egg/webtest/app.py", line 628, in do_request
    res = req.get_response(app, catch_exc_info=True)
  File "/Users/pbauer/.cache/buildout/eggs/WebOb-1.8.5-py3.7.egg/webob/request.py", line 1310, in send
    application, catch_exc_info=True)
  File "/Users/pbauer/.cache/buildout/eggs/WebOb-1.8.5-py3.7.egg/webob/request.py", line 1278, in call_application
    app_iter = application(self.environ, start_response)
  File "/Users/pbauer/.cache/buildout/eggs/WebTest-2.0.33-py3.7.egg/webtest/lint.py", line 201, in lint_app
    iterator = application(environ, start_response_wrapper)
  File "/Users/pbauer/.cache/buildout/eggs/plone.testing-7.0.1-py3.7.egg/plone/testing/_z2_testbrowser.py", line 38, in wrapped_func
    return func(*args, **kw)
  File "/Users/pbauer/.cache/buildout/eggs/plone.testing-7.0.1-py3.7.egg/plone/testing/_z2_testbrowser.py", line 65, in __call__
    wsgi_result = publish(environ, start_response)
  File "/Users/pbauer/workspace/plip/src/Zope/src/ZPublisher/WSGIPublisher.py", line 337, in publish_module
    response = _publish(request, new_mod_info)
  File "/Users/pbauer/workspace/plip/src/Zope/src/ZPublisher/WSGIPublisher.py", line 243, in publish
    obj = request.traverse(path, validated_hook=validate_user)
  File "/Users/pbauer/workspace/plip/src/Zope/src/ZPublisher/BaseRequest.py", line 540, in traverse
    return response.notFoundError(URL)
  File "/Users/pbauer/workspace/plip/src/Zope/src/ZPublisher/HTTPResponse.py", line 986, in notFoundError
    raise exc
zExceptions.NotFound: http://nohost/plone/f1/ical_import_settings/enable
```

icemac:
The problem is that authentication is done _after_ traversal. So `guarded_getattr` always raises an `Unauthorized` for all attributes which are not accessible by `Anonymous`.

I also tried to move the `guarded_getattr` call to a post_traverse hook. But this does not help either because these hooks are called after authorisation. In a case of an `Unauthorized` they are not called at all. This way it is not possible to convert the `Unauthorized` of a method which is not mentioned in `allowed_attributes` to a `NotFound`.

Currently I am low on ideas and out of time. The next step could be to find out, how it ever worked before the changes in https://github.com/zopefoundation/Zope/compare/9bc2e7f0b64ff887891e4d8a7dc17e496bdf29df...da088f0b0d849c3d0bd512979a5672852197ea5a.

I adapted the test so it fails quite the same way as it does in `plone.app.event`.

davisagli:
@icemac I think that the way it worked in Zope 2.13 is that DefaultPublishTraverse would find the attribute via getattr, but it would fail access control validation so return a 401.

Why are you intent on returning a 404 rather than a 401 when an attribute is not allowed? That's not consistent with what the Zope publisher does in other cases when an object is traversable but not authorized. And as you've found, the way that the publisher is structured makes it impossible to customize traversal based on the user's access.

If the 404 is critical, you might be able to achieve it by making traversal return a DeferredAttrAuthProxy like this:

    class DeferredAttrAuthProxy(object):
        def __init__(self, context, name):
            self.context = context
            self.name = name

        def __call__(self, *args, **kw):
            try:
                attr = guarded_getattr(self.context, self.name)
            except (AttributeError, Unauthorized):
                raise NotFound(self.name)
            return attr(*args, **kw)

Make sure that DeferredAttrAuthProxy is declared as public (I forget the best way to do this.) Then make SimplePublishTraverse.publishTraverse do `return DeferredAttrAuthProxy(self.context, name)`. This way traversal returns an object that is public but that will check for access and delegate to the intended attribute when called.

icemac:
@davisagli Thank you for investigating this issue. I implemented your suggestion in this PR.
@pbauer Could you please check the current status of the PR against Plone?

pbauer:
Will do so asap

pbauer:
This breaks a lot of tests and features. Enabling ical-support now works but the a CSRF-check does not kick where it should.
 
Among the broken features with tests are: 

* adding content-rules with http://localhost:8080/Plone/+rule/plone.ContentRule (now raises a 404)
* adding portlets with http://localhost:8080/Plone/++contextportlets++plone.leftcolumn/+/portlets.News 
* Changing a template for folder with http://localhost:8080/Plone/folder/selectViewTemplate?templateId=summary_view

See https://jenkins.plone.org/job/pull-request-5.2/239/ (python2.7) and https://jenkins.plone.org/job/pull-request-5.2-3.7/213/ (python 3.7) for the test-results.

MatthewWilkes:
David's approach looks sensible to me, but I very much agree with his point that it's not normal for traversing to a view that you fail a permissions check for to issue a 404.

David's approach is very much like my one in #397 except that he creates a wrapper object where I mutate the `simple(...)`. I've tried my suggested change and the content rules, portlets and template change all work. It does have a CSRF error on ical changes, I'm not sure if that's correct or not.

I've pushed my changes to `zope.browserpage` to a branch (`wilkes/397-allowed-attributes`) and would have created a PR to demonstrate them but I also have to add the whitelist setter line to `Products.Five.browser.metaconfigure.page` and that involves creating a Zope branch but branching from master caused a version issue for me and I'm not sure where to go from.

MatthewWilkes:
Okay, I worked out how to build it. Sorry, I'm out of practice with the coredev buildout.

I've got a branch of Plone building at https://github.com/plone/buildout.coredev/pull/588 that includes @pbauer's plone.app.event and my changes to zope.browserpage and Zope core. I don't know if this will work, or if it's the route you'd want to take, but hopefully it's useful information.

d-maurer:
Matthew Wilkes wrote at 2019-5-27 11:57 -0700:
>David's approach looks sensible to me, but I very much agree with his point that it's not normal for traversing to a view that you fail a permissions check for to issue a 404.

I agree.

However, it is exactly this what security experts recommend:
let a resource inaccessible with the current proviledges behave
exactly like a non existing resource -- in order to
give a potential attacker the least information possible.

On the other hand, Zope does not follow this recommendation:
One of its operation principles is to allow a reauthentication
when the user tries to access a resource with insufficient
priviledges. Therefore, there is no need to implement the
recommendation at an isolated place.

icemac:
In #642 I created a variant which does not insist to have a HTTP-404 for view attributes which are not in `allowed_attributes` but this leads to HTTP-401 there. Maybe this is the easiest solution which additionally allows to use custom traversal adapters.

icemac:
@MatthewWilkes wrote:
> I've got a branch of Plone building at [plone/buildout.coredev#588](https://github.com/plone/buildout.coredev/pull/588) that includes @pbauer's plone.app.event and my changes to zope.browserpage and Zope core. I don't know if this will work, or if it's the route you'd want to take, but hopefully it's useful information.

Cool to see an approach which seems to work. I am not sure whether I like the necessary changes in `zope.browserpage`. If #642 does not work or if we could not agree on having HTTP-401 for disallowed attributes, I'd take this route.

pbauer:
#642 has the same problems as this one with portlets, content-rules and changing templates.

The approach by @MatthewWilkes (https://github.com/zopefoundation/Zope/commit/ab64f98d59f718765f02af092de4fde3fe8a1c2e plus https://github.com/zopefoundation/zope.browserpage/commit/f8dedc92a370d55081f56ef4ff3a7de3764edf30) seems to fix all features and tests in Python 3. In Python 2.7 there is the issue that enabling ical-support raises ``TypeError: unbound method __call__() must be called with simple instance as first argument (got nothing instead)``. See https://github.com/plone/buildout.coredev/pull/588#issuecomment-496400951 for the full traceback.

Forgive me for not having a opinion regarding 404 vs. 401 or implementation-details at this point. I only focus on finishing Plone 5.2

jensens:
I lost a bit track here about implementation details. 
My opinion to 404 vs 401: I just can repeat what I tried to say in our lunch-discussion in Halle at Saltlabs sprint:

1. If the attribute is defined at the interface or given in allowed attributes, but the user has no permission to access it, it is a 401. 
2. If the attribute exists on the view, but does not exist at the interface or given in allowed attributes it has to be a 404.

my 0.02€

MatthewWilkes:
@icemac The reason I was keen to do the bulk of this in `zope.browserpage` is because that is where both the code to support `allowed_attributes` and the code that prevents traversing to it is. My main complaint is that it involves making changes to `Zope` because of the duplicated metaconfigure.

I can see why you'd be a bit concerned about the `__call__` mutation, but I'm thinking of this as effectively allowing users to change the `__page_attribute__` on the fly through traversal, rather than traversing to a new view. The change to `__call__` is there because so many users write their own `__call__` rather than `index.html` or whatever the other default methods are.

icemac:
@MatthewWilkes If `zope.browserpage` is the best place to change the behaviour we should go for it. So this PR can be closed after #643 was successfully merged. I already ported the newly written tests there.

@jensens The approach in #643 seems to implement the behaviour you are expecting.

icemac:
Superseded by #643.

Issue #252, opened 2018-02-16
Update Zope Developer’s Guide to Zope 4


icemac:
The Zope Developer’s Guide at https://zope.readthedocs.io/en/latest/zdgbook/index.html has many references to Zope 2 and should be updated to work with Zope 4 on Python 3.

icemac:
@jugmac00 Thank you very much. I added a checklist to the issue description.

ghiodor:
Hello, I'm trying to convert Products.FSDump to Zope 4 and I was following the developers guide and examples in github. However, chapter 3 is not specific to Products and chapter 5 is not up to date. Currently I'm stuck with error "Products.FSDump setup command: Distribution contains no modules or packages for namespace package 'Products'", which I don't understand. More important, how can I develop a zope product and test it in Zope using buildout develop?

tseaver:
@ghiodor I'm not sure what your starting point is for doing this migration.  The [release tarball](https://files.pythonhosted.org/packages/50/3a/a76ef772f3bc0b5db163d67b039ffee31704b92ec90f37882fa555061224/Products.FSDump-0.9.5.tar.gz) has the correct package structure for a namespace package.

jugmac00:
@ghiodor Updating the documentation is a mammoth project - I am working on this like for 9+ months - both in my spare time and without having contributed to the underlying source code, so I am learning by documenting :-) And currently my spare time is almost non existing :-(

Can't help you with the error message - maybe create a new ticket as here it could be easily overlooked? 

Or ask in the new discussion board over at the Plone forum:
https://community.plone.org/c/zope

d-maurer:
> Distribution contains no modules or packages for namespace package 'Products'", which I don't understand.

@ghiodor Zope no longer needs special things for "products": any subpackage of the namespace package `Products` can be a product (but likely, it still has a top level `initialize` function). The error you are seeing therefore is not a Zope message but one of `setuptools`. Apparently, `setuptools` cannot find the actual package. To get the packaging right, you no longer need Zope specific knowledge; [Python packaging knowledge](https://packaging.python.org/tutorials/packaging-projects/#requirements-for-packaging-and-distributing) is sufficient.

ghiodor:
[...]

PR #614, opened 2019-05-10
Revise section "XML-RPC"

jugmac00:
- improve formatting
- update syntax and imports for Python 3
- remove paragraph about DTML
- remove wrong information about missing basic authentication support
- remove wrong comment that "None" is no valid return value via XML-RPC
(as Zope enables "allow_none")


jugmac00:
>  This stuff still works under WSGI? I had no idea.

Afaik it was removed, but reintroduced by @davisagli again. And yes, it works - I tried all code examples with my test instance of Zope.

jugmac00:
Chapter 4 of Zope dev guide is done... only took me half a year :-)

@dataflake and @d-maurer - thank you very much for all the reviews!

jugmac00:
only 5 chapters left cf #252

PR #621, opened 2019-05-13
wheel 0.33.3 was removed from pypi

pbauer:
See https://github.com/pypa/wheel/issues/294#issuecomment-491544685


dataflake:
@icemac I don't think this has anything to do with being conservative about versions. If you look back we don't do those dependency updates all that often, and someone just removing a recent release egg from PyPI is a rare (and really unexpected) occurrence. I think we're doing just fine the way we have been handling those updates.

pbauer:
A problem is now that all builds that depend on https://raw.githubusercontent.com/zopefoundation/Zope/4.0/versions.cfg will break unless they override that version-pinn. I updated this for Plone in https://github.com/plone/buildout.coredev/pull/586. 

Should we release 4.0.1 soon? If yes then I would like https://github.com/zopefoundation/Zope/pull/624 and another yet nonexistent fix needed for Plone 5.2 final that fixes a issue inZPublisher/HTTPResponse.py (see https://community.plone.org/t/please-help-with-failing-tests-that-block-5-2-final-release/8512/8)

dataflake:
I personally don't mind a relatively quick release, but that call is up to @icemac as he is doing all the work.

dataflake:
P.S.: I'd really like to see the fix for #397 in a bugfix. I believe the PR for it, #610 , would only need to include the desired behavior change (return 404 instead of 401).

PR #627, opened 2019-05-13
Resurrect ZODB packing from the ZMI

### LINKED ISSUE #623: Packing a non-ZEO Zope instance broken
For Zope 2.13, there was a button in the ZMI which allowed packing a DB - the button used to call ApplicationManager.manage_pack.

Both the button and the method got removed back in 2016
- without updating the documentation ( https://zope.readthedocs.io/en/latest/zopebook/MaintainingZope.html#packing-and-backing-up-the-filestorage-database )
- without offering an alternative, like a script in the bin directory (at least I could not find one) (UPDATE: there is a script in the bin dir for zeo server, when you build a separate zeo server, called zeopack)

Ideally
- we get back the button in the ZMI (I am not the only one who wants the button back; there was a quick discussion at last weeks Zope sprint)
- find some time to update the documentation
- get a script for a standalone Zope setup?

dataflake:
Fixes #623 

I have no idea why that was taken out, which made packing a non-ZEO ZODB near impossible.


dataflake:
I have also chosen to remove the "advice" about automatically packing the database:

- that procedure wouldn't even have worked because `manage_pack` requires a `POST` request

- it wouldn't have worked because `manage_pack` requires authentication as a Manager-type user

- advising anyone to use scripts that call up a URL and require you to put cleartext credentials into the script to make it work is just horrifically bad. It's worse than not providing any advice.

Anyone who is in need of automatic packing will already use ZEO, which comes with ready-made scripts for that purpose.

src/App/ApplicationManager.py
```diff
@@ -250,7 +250,27 @@ def manage_minimize(self, value=1, REQUEST=None):
...
 
         if REQUEST is not None:
-            REQUEST.RESPONSE.redirect(REQUEST['URL1'] + '/manage_main')
+            msg = 'ZODB in-memory caches minimized.'
+            url = '%s/manage_main?manage_tabs_message=%s' % (REQUEST['URL1'],
+                                                             msg)
+            REQUEST.RESPONSE.redirect(url)
+
+    @requestmethod('POST')
+    def manage_pack(self, days=0, REQUEST=None, _when=None):
```
icemac: Should this method be restricted to Manager users?

dataflake: It is because it has a `manage_` prefix. You should know this.

dataflake:
I did not write a test because there's no point.

What exactly am I testing? Pack functionality? That's complete overkill. All I am doing is to provide a button and a method to access it.

When you say "it doesn't break in the future" then the only test would be "can I access packing?", because that's what broke. But how would you even test that? Look for a method name like `manage_pack`? Look for a button on a web page?

There's simply too little that makes sense to test here.

src/App/ApplicationManager.py
jugmac00: The new form / button in the ZMI is the only caller of **manage_pack** and there is no `_when` input field - so `_when` could be removed from both the signature of the method and the body.

d-maurer:
Jürgen Gmach wrote at 2019-5-15 00:30 -0700:
>jugmac00 requested changes on this pull request.
>
>**My 2 cents on tests**:
>Discussions about if to write tests could be completely avoided if we'd keep to a single rule - no new code without tests, no bug fix without tests. Tests are not only about proving the new code works, but especially about laters having no fear to refactor or change code, and to make sure no features get removed accidentally.

There are cases where the effort required to write a test is in a very bad
relation to the gain.

My most favorit example is a cometic UI change:
for example, the UI designer looks at a page and
decides the grid needs a small adjustment
to look good (maybe even averaged for serveral devices). In such a case,
a test will not help much; it is far more efficient that the UI designer just
looks at the result. A test checking that the effective value is what
the designer has fixed is not very helpful because a significant UI change
has a good chance that the value must change as well.

dataflake:
I'm 100% with Dieter here.

- this is not new code. It is old code I am just copying back in
- testing whether a pack works belongs in the underlying ZODB code, not here
- testing a UI feels pointless

This is a place where it feels like everyone wants a test just to have a test and no one is questioning whether that makes any sense. I do not want to spend time writing tests just for tests sake, when tests are a <German word alert> Selbstzweck.

I fully agree that tests are needed for new code. But making tests a condition without considering whether a test makes sense at all or whether writing a test is in a sane relation to the actual code change just erects unnecessary barriers to contributions

 The goal of this project is not 100% test coverage. It's to provide software that works well for as many people as possible. Don't punish people who want to reach the goal with unreasonable test requirements.

jugmac00:
> Don't punish people who want to reach the goal with unreasonable test requirements.

By no means I want to spoil anybody's intent to contribute. Personally, I do enjoy writing tests (I am used to pytest only, though) - both at home and at work. And as I usually write tests first, there is no question about adding tests or not.

Obviously, there is no agreement on my suggestion - that is fine for me, too.

tseaver:
FWIW, the rationale for "testing UI" here is to ensure that any logic in it doesn't have bugs.  For the case of `manage_pack`, there are a couple of bits which might be buggy: 

- The `days` variable might not be marshalled correctly as an `int` (maybe `float` would be better. :).  This would be a bug in the DTML, but the method could handle forcing the conversion.
- The computation of the point-in-time in the past might be wrong.

In addition to testing those two parts of the method, mocking out the value returned by `self._getDB()` lets you assert that its `pack` method is actually called with the expected value.

src/App/ApplicationManager.py
icemac: I did not know that. It seems to happen in `InitializeClass`:
https://github.com/zopefoundation/AccessControl/blob/099c03fb5bcb7e3cf8e9f3f959a1a30508cd5422/src/AccessControl/class_init.py#L54-L57
Thank you for the hint.

dataflake:
Test has been added so I am finally going to merge this now.

PR #633, opened 2019-05-18
Do not cache (implicit) request access to form data or cookies

### LINKED ISSUE #630: "ZPublisher.HTTPRequest": minor security risk: tainted value access not reliable
`HTTPRequest` takes immense efforts to mark unsafe values as "tainted" - but gets unsafe via caching in `other`.
Assume that access to the request variable "x" is unsafe. If the first access to "x" is with `HTTPRequest.get(...returnTaints=True...)`, then all accesses to "x" deliver a value marked as tainted. However, if the first access is with `returnTaints=False`, then all accesses deliver a value without "taint mark" (because the first access caches the returned value in `other` and the `other` lookup no longer honours `returnTaints`).

d-maurer:
This PR fixes #630 by not caching the result of an implicit request lookup (in the form `request.get` or `request[...]` in contrast to `request.{form|cookies}{.get,[...]}`) in `other`.

There is a minor backward incompatibility affecting applications which might rely on finding form data/cookies in `other` after the first access.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ Fixes
 - Fixed logic error in exceptions handling during publishing. This error would
   prevent correct Unauthorized handling when exceptions debug mode was set.
 
+- Do not cache (implicit) request access to form data and cookies in ``other``
+  (`#630 <https://github.com/zopefoundation/Zope/issues/630>`_).
+
 
 4.0 (2019-05-10)
 ----------------
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -145,10 +145,8 @@ class HTTPRequest(BaseRequest):
       - Environment variables
 
         These variables include input headers, server data, and other
-        request-related data.  The variable names are as <a
-        href="http://hoohoo.ncsa.uiuc.edu/cgi/env.html">specified</a>
-        in the <a
-        href="http://hoohoo.ncsa.uiuc.edu/cgi/interface.html">CGI
+        request-related data.  The variable names are as specified
+        in the <a href="https://tools.ietf.org/html/rfc3875">CGI
         specification</a>
 
       - Form data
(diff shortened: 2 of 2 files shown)


d-maurer:
Jens Vagelpohl wrote at 2019-5-18 08:36 -0700:
>dataflake requested changes on this pull request.
> ...
>First of all, the comment "Issue 630" is completely meaningless to anyone just looking at that code. And as we have seen in the past, when bug trackers change we lose access to that information even if the viewer knew what "bug 630" means. If you want to leave a comment, please leave a meaningful comment, like "we are not caching this value in ``other`` because..."

The `CHANGES.rst` is full of references to bug tracker entries.
Those references, too, will lose some of their usefulness
should the bug tracker change.

A code reviewer who really is interested in the details
about "Issue 630" (and does not know how to look up
a detailed description in the bug tracker) can look in `CHANGES.rst`
for some details.

>Secondly, please don't just comment out the offending code and leave it hanging there. Remove it. A meaningless comment will explain what happened here before.

I will remove all the comments.
You will then need to get information about the change via the
git commit log.

d-maurer:
Jens Vagelpohl wrote at 2019-5-18 09:11 -0700:
>dataflake commented on this pull request.
>Question: Why not still cache those values that are correctly looked up and marked as tained? That would solve the issue in #630 and still preserve a little beneficiary caching, right?  Completely turning it off everywhere seems unnecessary.

Start optimizing when it is really necessary.

How often is the same cookie/the same form data accessed several times
in a request? I have never seen code that looks up a cookie/form variable
in a loop; thus one can expect the number of those accesses to
be in the order of 1 (and not 10, 100 or even more).

How many microseconds are saved by a cached access?
Caching saves typically a couple (4-6) of dict lookups
and two comparisons against the first character of *key*.
Only if *key* starts with "U" or "B", it can be a bit more expensive
(checking for the URL/BASE pattern).

Issue #636, opened 2019-05-20
Document ZPublisher request parameter controls


d-maurer:
`ZPublisher` (more precisely `ZPublisher.HTTPRequest,processInputs`) supports a large set of controls which guide the interpretation of request parameters. A request parameter (as supported by Zope) has the general form `name=value`. Controls are specified as suffixes of the form `:<control>` to the name part of the request parameter. They can specify for example a type for the value or how the parameter should be aggregated with other request parameters to form a complex form variable. Examples are `x:int=1` (gives a form variable `x` with integer value `1`) and `r.x:record=1` (gives a form variable `r` with attribute `x` with value `'1'`).

The discussions in #618 and #634 have shown that we need a specification/documentation of these features.

Where should such a documentation/specification be located?
 * my personal preference would be a file `request_parameter_controls.rst` alongside `HTTPRequest` with a reference in the docstring of its `processInputs` to this file

with alternatives:
 * directly put it in the docstring of `processInputs`
 * put it in the "Zope Developer Guide"
 * put it in the "Zope Book"

Such a document would have wide ranging effects:
 * it is a prerequisite for #619 
 * it would provide answers for #634 
 * it would contribute to the resolution of #558

icemac:
I like the idea to have the documentation near the code although it should be in the documentation, too to be easily findable.

d-maurer:
Michael Howitz wrote at 2019-5-20 23:15 -0700:
>I like the idea to have the documentation near the code although it should be in the documentation, too to be easily findable.

We could first put is near the code (I have seen that this
is already used in `Products.Transience`) and put it in a more
prominent place later - once it is finished and "polished".
Or we put a reference at a prominent place to the location
in the code tree.

d-maurer:
Closed via #640

PR #640, opened 2019-05-24
Document request parameter handling

d-maurer:
see #636


src/ZPublisher/request_params.rst
```diff
@@ -0,0 +1,296 @@
+=========================
+Request parameter support
+=========================
+
+Introduction
+============
+
+Zope responds to requests, specified via an url, request headers
```
jugmac00: url => URL

src/ZPublisher/request_params.rst
```diff
@@ -0,0 +1,296 @@
...
+
+An aggregator directive tells Zope how to process parameters with
+a similar name.
+
+Zope supports the following aggregators:
+
+list
+  collect all values with this name into a list.
+
+  If there are more than a single parameter with the same name,
```
jugmac00: "If there is..." - but I am not 100% sure about this.

src/ZPublisher/request_params.rst
```diff
@@ -0,0 +1,296 @@
...
+  the ``default`` directive.
+
+record
+  this directive assumes that the parameter name starts with
+  *var*\ ``.``\ *attr*.
+
+  It tells Zope to create a request variable *var*
+  with as value a 
+  record (i.e. a ``ZPublisher.HTTPRequest.record`` instance) and
+  set its attribute *attr* to the parameter value.
```
jugmac00: Sounds a bit rough.... maybe... "It tells Zope to create a request variable *var* of type record...

src/ZPublisher/request_params.rst
```diff
@@ -0,0 +1,296 @@
...
+
+It then loops over the request parameter sequence.
+
+
+For each request parameter, the processing consists of the following steps:
+
+1. Some variables are set up:
+
+   isFileUpload
+     does the parameter represents an uploaded file?
```
jugmac00: represents => represent

src/ZPublisher/request_params.rst
```diff
@@ -0,0 +1,296 @@
...
+
+After all request parameters have been processed,
+request variables from ``defaults`` are put into ``form`` if this
+does not yet contain such a variable.
+If a method directive has been processed, then the traversal
+path is extended accordingly.
+
+As a security measure, mainly for DTML use, request variables
+are not only made available in the request attribute ``form``;
+a (somewhat) securized version of them is made available in
```
jugmac00: Does the word "securized" exist?

jugmac00:
@d-maurer Thank you very much for this extensive documentation! I really wished this existed before before I worked on the Object Publisher part of the Zope Developer documentation, as your text really makes it very clear what happens and what are the rough edges.

I also feel your documentation offers way more value than parts of the Zope Developer documentation, and I feel it sells itself short as it is located in a package (and possibly not rendered to html) instead of in the documentation. At minimum in the long term it should get a more prominent location.

I am a bit in a rush (on my way to Pycon Web in Munich), so I could not review it in detail - but will catch up later or beginning of next week.

src/ZPublisher/request_params.rst
d-maurer: I have reformulated to ensure the plural is correct.

d-maurer:
@jugmac00 
> ...
> I also feel your documentation offers way more value than parts of the Zope Developer documentation, and I feel it sells itself short as it is located in a package (and possibly not rendered to html) instead of in the documentation. At minimum in the long term it should get a more prominent location.

I plan to soon propose a major overhaul of the request parameter handling - one that gets rid of most of the known bugs/caveats. The new implementation should get much clearer, too; it seems for example not necessary, that each subcase has parallel code for the tainted and untainted variant; instead construct the untainted ``form`` dict and then derive the tainted variant thereof.

dataflake:
Thank you Dieter, that is great documentation. You may have noticed that I did some grammar/wording changes earlier.

The only aspect that makes me unhappy is hiding it away in a place where no one will find it. With my last commit I have merged the contents into the Zope Developer Guide. Please take a look and let me know if this is OK. If yes then we can delete the document at the location in the ZPublisher folder.

d-maurer:
Jens Vagelpohl wrote at 2019-5-24 17:01 -0700:
>Thank you Dieter, that is great documentation. You may have noticed that I did some grammar/wording changes earlier.

I am not happy with the section title change
`Request parameter directive kinds` -->
`Request parameter directives` because the section actually describes only
the directive kinds and not the individual directives.
This is valid also for the last paragraph of the introduction.

I think english prefers `converter specific` over `converter-specific`.

The description in section "Processing model" is too detailed for
an end user documentation. You have edited out my explanation
why I have chosen this level of details (--> peculiar logig leading
to surprises): the target audience for my documentation has been
the "developers **OF** Zope", not the "developers **WITH** Zope".
In the documentation for "developers with Zope", the whole
section "Processing model" (and references thereof) can go away.

In list item 3. of section "Processing model" the `in addition` and `instead` described details of some importance: typically, only variables set up in step 1. are updated; `:tuple` updates `flags` (it ors `SEQUENCE`) and in addtion `tuple_items`; method (all variants of this kind) directives update `method` instead. The peculiar handling of `:tuple` explains the later described bug that it often is equivalent to `:list`.

In the description of the `defaults --> form` handling, a spurious `if` has slipped in.

dataflake:
I have incorporated your suggestions. My goal was to have the text in the Zope Developer's Guide as the only copy and remove the text file in the ZPublisher module, so if you could concentrate just on the ZDG parts that would be great.

You're correct that some of the details are unsuitable for normal application developers or Zope users and I have added a note to that effect at the start of the processing model section in the ZDG. However, we do not have any canonical place for documentation aimed at Zope developers, the ZDG is the closest we have. People won't have a hard time finding it there.

d-maurer:
@dataflake I am quite happy with the current state.

I have made some minor modifications:
* fixed the documentation regarding the request lookup order
* removed some items from the "converter"s list which are now classified as "aggregator"s
* used the same formatting for "aggregator"s and "converter"s.

There is still one point which does not satisfy me fully: the description is now placed into the context of object publishing, i.e. focused on the use of request variables as arguments for the call of the "published object".  Maybe, we should add a special note that this is only one use of request variables and that they are often accessed via the request object.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -727,7 +759,10 @@ many converters:
 - **ulines**, **utokens**, **utext** -- like **lines**, **tokens**,
   **text**, but always converts into unicode strings.
 
-If the publisher cannot coerce a request variable into the type
+The full list of supported converters can be found
+in ``ZPublisher.Converters.tpye_converters``.
```
icemac:
```suggestion
in ``ZPublisher.Converters.type_converters``.
```

Additionally could there be a link to the module?

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -951,6 +935,220 @@ simple as possible.
...
+   flags
+     to indicate which processing types are requested via directives
+
+     Processing types are "ignore", "aggregate as sequence",
+     "aggregate as record", "aggregate as records", "use as default",
+     "convert" (using ``converter_type`` and ``character_encoding``)
+
+2. The parameter value is checked to see if it is a file upload.
+   In this case, it is wrapped into a ``FileUpload``, and ``isFileUpload``
+   is updated
```
icemac:
```suggestion
   is updated.
```

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -951,6 +935,220 @@ simple as possible.
...
+
+   character_encoding
+     the most recently seen encoding from an encoding directive
+
+   flags
+     to indicate which processing types are requested via directives
+
+     Processing types are "ignore", "aggregate as sequence",
+     "aggregate as record", "aggregate as records", "use as default",
+     "convert" (using ``converter_type`` and ``character_encoding``)
```
icemac:
```suggestion
     "convert" (using ``converter_type`` and ``character_encoding``).
```

PR #642, opened 2019-05-28
Fixing fixing `allowed_attributes` - Raise HTTP-401

### LINKED ISSUE #397: BrowserViews implement IPublishTraverse in Zope 4 which breaks allowed_attributes and allowed_interface

icemac:
Fixes #397.

This PR is #610 but whit the difference that HTTP-401 is returned for attributes which are not in `allowed_attributes`.

In my mind this is the easiest solution if we can decide that HTTP-401 is a valid enough status code for this case.


icemac:
@pbauer @MatthewWilkes could you please test this branch against Plone?

pbauer:
This has the same issues as mentioned in https://github.com/zopefoundation/Zope/pull/610#issuecomment-496113775. Adding portlets and content-rules is broken as well as changing a folders view.

icemac:
This PR can be closed when #643 got successfully merged.

icemac:
No it can be closed now, as it is a dead end.

PR #644, opened 2019-05-29
Simplify linting configuration and lint docs, too.

icemac:
I did not port the code in the examples to a current version of `zope.interface` as this PR is only about linting.


icemac:
Thank you.

Issue #641, opened 2019-05-28
Modernize request parameter handling


d-maurer:
Zope's current request parameter handling has several known bugs/caveats:
1. There is almost no error handling:
    - unrecognised directives are silently ignored
   - if a request parameter contains several converter directives, the 
     leftmost wins
   - if a request parameter contains several encoding directives, the
     leftmost wins
   - if a request parameter contains an encoding but no converter
     directive, the encoding directive is silently ignored
   - some directive combinations do not make sense (e.g. ``:record:records``);
     for them, some of the directives are silently ignored

2. Usually, the order of aggregator directives in a request parameter does
   not matter. However, this is not the case for the ``:tuple`` directive.
   To really produce a tuple request variable, it must be the left most
   directive; otherwise, it is equivalent to ``:list``.

   In addition, ``:tuple`` is always equivalent to ``:list`` for
   request variables aggregated as record or sequence of records.

3. The main use case for the ``:default`` directive is to provide a
   default value for form controls (e.g. checkboxes) for which the browser may
   or may not pass on a value when the form is submitted.
   Unfortunately, this only works at the top level.
   It does not work for subcomponents, e.g. an attribute of a "record".
   As a consequence, if a request parameter combines ``:default`` with
   another aggregator directive, the result may be unexpected.

4. The request preprocessing happens at a very early stage, before
   traversal has taken place. As a consequence,
   important configuration for application specific error handling
   may not yet have taken effect. Exceptions raised during this stage
   are reported and tracked only via "root level" error handling.

In addition, for Python 3, Zope's handling of encoding directives is currently broken (--> #634) and converter support is partially broken (#638, #558). Furthermore, Zope's encoding support is not in line with modern standards, e.g. [HTML5 form support](https://www.w3.org/TR/html5/sec-forms.html#multipart-form-data).

This issue proposes a major overhaul of Zope's request parameter handling:
* either drop support for the `zpublisher-default-encoding` configuration option and fix the encoding to `utf-8` (this seems to be in line with current standards, see e.g. [HTML living standard](https://html.spec.whatwg.org/multipage/semantics.html#charset); they seem to view any non-`utf-8` encoding as "legacy support" only) or fully support
[HTML5' form encoding practices](https://www.w3.org/TR/html5/sec-forms.html#multipart-form-data).
[...]

Issue #634, opened 2019-05-18
"HTTPRequest.processInputs": broken "character_encoding" handling


d-maurer:
```
from ZPublisher.HTTPRequest import HTTPRequest
try: from urllib.parse import quote
except ImportError: from urllib import quote

def epi(qs):
  r = HTTPRequest(
    None,
    dict(
      SERVER_NAME="nohost",
      SERVER_PORT="80",
      SCRIPT_NAME="",
      REQUEST_METHOD="GET",
      QUERY_STRING=qs),
    None,
    )
  r.processInputs()
  return r.form

bstring = u'\x80\x81\x82' # 'latin1' decoded sequence of bytes
epi("bstring:latin1:bytes=" + quote(bstring.encode("latin1")))
```
produces three different results in Zope2, Py3/Zope4 and Py2/Zope4:

 Zope2: `{'bstring': '\x80\x81\x82'}` (the expected result)

 Py3/Zope4: `{'bstring': b'\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'}`

 Py2/Zope4: `{'bstring': '\xc2\x80\xc2\x81\xc2\x82'}`

Zope 4 assumes that the query string content is encoded via the configuration dependent `zpublisher_default_encoding`. This is acceptable (only) when the request parameter does not explicitly specify its own character encoding.

`dm.zopepatches.ztutils` explicitly uses `:utf8` to reliably pass around unicode values independent from a potentially different `zpublisher_default_encoding`.

dataflake:
After eyeballing the code in both Zope 4 and Zope 2 the statement "Zope 4 assumes that the query string content is encoded via the configuration dependent zpublisher_default_encoding" appears to be true in Zope 2 as well. It's just that the default encoding value has changed.

Using your specific example will always work in Zope 2, as your `latin-1` encoded sample string matches the Zope 2 default `latin-15` well enough.

Looking at results where the sample string encoding matches neither the Zope 2 nor the Zope 4 defaults may be more enlightening.

dataflake:
I have stepped through the code under Python 2 and it seems to do everything correctly. The input gets decoded using the character set you're explicitly passing in, and then re-encoded using the default ZPublisher encoding. The result on Zope 4 is a correctly encoded UTF-8 binary string. Do you expect something different?

Under Python 3 the field value is already messed up when it reaches the code around line 663 that deals with fields that have an explicit character encoding.

[...]

Issue #619, opened 2019-05-11
Enhance ZPublisher "type suffixes" and "ZTUtils" ...


d-maurer:
... by functionality from `dm.zopepatches.ztutils`?

`ZTUtils.make_query` (and friends) allow together with `ZPublisher`'s type suffixes to pass typed information between requests. However, there are some weaknesses.

`dm.zopepatches.ztutils` addresses the following weaknesses:
 - missing support for `None`
 - missing support for empty lists/tuples
 - limited support for nested structures
 - fixed set of supported types

by

 - allow for the registration of extensions (essentailly new type suffixes and corresponding converters); used to handle `None` and deeply nested JSON serializable structures (via a `json` extension)
 - use `:tokens:default` to provide a default for empty list/tuple

dataflake:
You're welcome to create a PR that integrates functionality directly, meaning without introducing dependencies to outside packages.

PR #648, opened 2019-06-12
Modernize request parameter handling #641

### LINKED ISSUE #638: :lines converter converting to list of bytes instead of list of str
An textarea like

```
<textarea name="mytext:lines">
..
</textarea>
```

is converted by the new ZPublisher in Zope 4/Python 3 to list of byte strings instead of list of `str`.

This is inconsistent with the default behavior of `<input type="text"> being returned as `str` inside `request.form`.

I think you could specify the encoding in older Zope versions like `name="mytext:lines:unicode">` or `name=mytext:lines:utf8` however the specifying an encoding does not make a difference.

d-maurer:
Adresses #641.

From `CHANGES.rst`: 
 - fully recursive aggregation which can handle structures of arbitrary depth
 - simplified processing model
 - support for special HTML5 features: ``_charset_`` informs
    about the used form encoding; character references used
    to work around encoding limitations
 - treats parameter values internally as text
    (this means unicode for Python 2).
    For Python 2, the conversion to unicode is skipped if it
    results in a ``UnicodeDecodeError``. The value is then used
    as is.

    For Python 2,
    the final values for parameters without converter and encoding directive
    are encoded with Zope's default encoding; character references
    are used for characters which cannot be encoded.
- Errors encountered during request parameter processing are
    not reported immediately (at this stage, application specific
    error handling has not yet been set up). Instead, a
    ``post_traverse`` is registered which will raise
    a ``RequestParameterError`` exception after the traversal
    in the proper application context.
    The ``RequestParameterError`` describes all errors
    encountered during request parameter processing.
 - ``FileUpload`` has new attributes ``type`` (the associated
    MIME type or ``None``) and the ``dict`` ``type_options``
    (containing the provided MIME type parameters).

Backward incompatibilities:
 - a parameter must now follow a corresponding default parameter
    to override the default paramter; formerly the relative order
    of parameter and default parameter was of no importance.

    There is a new directive "conditional" which can also be used
    to define a default value. A conditional parameter is ignored,
    if there is already a corresponding parameter, and
    otherwise acts like a default parameter. Thus, its behaviour
    is comparable to the former bahaviour of "default".
- aggregators are now applied from left to right; especially,
    their relative order is important.
    Formerly, aggregators were applied in a fixed order --
    independent of the order in which they were specified.
 - the converter *functions* (in ``ZPublisher.Converters``)
    no longer support the conversion of files (because they
    do not know the encoding applicable for the file).
    The converter *directives*, however, can still be applied
    to (uploaded) files. They use the encoding explicitly
    specified via an encoding directive or fall back to
    Zope's default encoding.

Fixes #638, #634, #558.

Due to the backward incompatibilities, we might want to wait until Zope 5. In this case, and if Zope 5 drops Python 2 support, some Python 2 complexities could be removed.


icemac:
@d-maurer Wow, this one is huge! Thank you for tackling these issues.
The current [Zope roadmap](https://blog.gocept.com/2019/05/10/zope-roadmap/) says that Zope 4 will only a short term release. So there should be no 4.x feature releases. On the other hand Zope 5 should only drop Python 2 compatibility thus code successfully running on Zope 4 on Python 3 should also run without modifications on Zope 5. This means I'd like to postpone this PR until Zope 5 is released for the Zope 5.1 release.
Is this okay for you?

d-maurer:
Michael Howitz wrote at 2019-6-12 23:01 -0700:
>@d-maurer Wow, this one is huge! Thank you for tackling these issues.
>The current [Zope roadmap](https://blog.gocept.com/2019/05/10/zope-roadmap/) says that Zope 4 will only a short term release. So there should be no 4.x feature releases. On the other hand Zope 5 should only drop Python 2 compatibility thus code successfully running on Zope 4 on Python 3 should also run without modifications on Zope 5. This means I'd like to postpone this PR until Zope 5 is released for the Zope 5.1 release.
>Is this okay for you?

Okay.

But it would imply that a minor release (5.1) would introduce
backward incompatibilities. Would it not be better to have
those incomoatibilities in a major release?

--
Dieter

dataflake:
> Okay. But it would imply that a minor release (5.1) would introduce backward incompatibilities. Would it not be better to have those incomoatibilities in a major release?

I'd agree generally, but that pushes it out even further.

We already have internal agreement that we want to make the switch from 4.x to 5.0 as easy as possible so that there is no good reason to stick with 4.x. Release 5 will be purged of any Python 2 workarounds and thus will be a much easier platform to maintain going forward. But that means no new incompatibilities in release 5.0, so this PR cannot go into it.

Michael is suggesting 5.1 as a compromise so this won't have to wait all the time for a release 6, which has no clear time line at the moment.

gforcada:
*just an idea:* what about wrapping the code with an environment variable/configuration option/something of the sorts, so that it can be merged *right away* and be enabled as needed.

Together with some deprecation warnings might be enough signaling to encourage users to enable the new behavior.

d-maurer:
Gil Forcada Codinachs wrote at 2019-6-13 08:36 -0700:
>*just an idea:* what about wrapping the code with an environment variable/configuration option/something of the sorts, so that it can be merged *right away* and be enabled as needed.

I could publish the new request parameter handling in a
`dm.zopepatches.*` package. This way, anyone interested could
use it right now -- independent on any Zope release decision.

However, there may be no need to hasten:
The request parameter features have 3 use cases:

1. support the calling of (a bit more complex) functions/scripts over
   the web, e.g. for testing/debugging purposes

2. providing elementary form support

3. act as a deserializer for the serializers of `ZTUtils` (i.e.
   `make_query` and `make_hidden_input`).

The current features work sufficiently well for simple cases (only
converters). This is sufficient for 1.: you
would not want to manually construct complex parameters.

For form support, you would nowadays use a form subframework
(e.g. `z3c.form`, `zope.formlib`). Such a framework gives much better
(easier to understand, better user feedback in case of errors,
more boiler plate, ...) support.

Remains 3. Again: `make_query/make_hidden_input` works for
simple cases. It can utterly fail for more complex structures,
but few people need to pass on complex objects across requests.


My personal interest is in 3: for a customer, I have extended
`make_query/make_hidden_query` (--> `dm.zopepatches.ztutils`)
to improve its ability to pass on more complex values.
However, it will still fail in some cases (e.g. inhomogenous lists
of records); at the moment, however, it is sufficient for
my customer's needs.

I am also interested that software behaves as its documentation suggests.
When e.g. the Zope Developer Guide promises that I can use
":default" to define a default value, then I expect that this
works also in a more complex case such as `x.a:default:records` (which
currently behaves in a strange way).
Having fixed the tests which broke due to my changes, I found
out that `x:default=2&x:default=3&x=1` currently results in
`x = ['1', '2', '3']` and was highly surprised that this was deemed
acceptable behaviour for `:default`. Apparently, some current
tests have not been used to verify that the implementation
behaves as promised but rather to document its (sometimes strange)
behaviour. But Zope has lived with the discrepancy for years,
there should be no problem to extend this for a few more months.

icemac:
I think a feature toggle makes the code unnecessary complex as it requires to keep both variants. I find another `dm.zopepatches.*` package okay. We could add a ticket here for Zope 5.1 or even Zope 6 to reintegrate the package into Zope.

@d-maurer Would this be okay for you?

d-maurer:
Michael Howitz wrote at 2019-6-20 23:20 -0700:
>I think a feature toggle makes the code unnecessary complex as it requires to keep both variants. I find another `dm.zopepatches.*` package okay. We could add a ticket here for Zope 5.1 or even Zope 6 to reintegrate the package into Zope.
>
>@d-maurer Would this be okay for you?

I will postpone the publication of a `dm.zopepatches`
package for this until this should become necessary --
for the following reason:
The `dm.zopepatches` use monkey patching to change Zope code.
Most of the changes in this PR are easily achievable via monkey
patching (essentially replacing
`ZPublisher.HTTPRequest.HTTPRequest.processInputs`).
There is however one important exception: postponing the exception
reports via `post_traverse` (to get them in the application context)
requires a changed handling of `BaseRequest._post_traverse`.
Currently, this attribute is set at the start of `BaseRequest.traverse`
and deleted at its end. `post_traverse` is therefore currently not available
for `processInputs`; to change this, `BaseRequest.traverse` needs to
change. But this is a very complex method; I would hate to have
a (modified) copy of it in my patch. It might be possible to avoid
changing `BaseRequest.traverse` by a nontrivial descriptor for
`_post_traverse` but I do not yet know the details.


If you agree, I start working on
"https://github.com/zopefoundation/Zope/issues/619"
based on the branch "modernize_request_parms#641",
because the new request parameter handling makes
`ZTUtils` easier, more reliable and more flexible
(my primary aim in implementing the new parameter handling).

dataflake:
This is a great candidate for Zope 6

d-maurer:
I ask myself whether I should resume work on this PR.

Part of it has already arrived in the release: sanitize  "taint" logic

2 parts remain:
1. postpone request parameter error reports until after traversal (to report them via the "correct" `error_log`)
2. more general but backward incompatible request parameter control

Currently, I think 1. could still be beneficial.
Regarding 2., developers with sophisticated need for request parameter control likely will use a form library (e.g. `z3c.form` or `zope.formlib`) rather than enhanced `ZPublisher` requst parameters. In view of this  2. is likely not worth the trouble caused by the backward incompatibility.
What do you think?

dataflake:
I agree.

d-maurer:
Most parts of this PR are now available in `master`.  Still missing are:
* generalized (and more consistent) request parameters
* better reports for errors during request parameter processing.
Both would lead to backward incompatibilities. Likely, this is not worth it.

icemac:
@d-maurer Is the branch still needed or can it be deleted?

d-maurer:
Michael Howitz wrote at 2024-11-7 23:51 -0800:
>@d-maurer Is the branch still needed or can it be deleted?

It might still contain one interesting element:
delaying the report of errors during parameter request processing
until after the traversal.

Recently, we have got an error report concerning the handling
of such errors: they have not been handled at all by `Products.SiteErrorLog`.
This problem is meanwhile fixed but it seems unnatural that
such errors are reported to the top level `error_log` rather
than that selected after traversal.

Implementing the delay is not trivial because it may want
to report not a single but multiple exceptions.
The approach in the branch above may not yet be optimal.


Should we not be interested in delaying the parameter request
processing error reports, the branch can go away.

PR #655, opened 2019-06-18
Prevent header spoofing via underscore/dash conflation

jmuchemb:
See https://www.djangoproject.com/weblog/2015/jan/13/security/


jmuchemb:
We found this security flaw while working on WSGI support for ERP5. Waitress fixed it years ago and we have to change our frontends to stop using headers with `_`. But we need Medusa to be fixed before we're ready to move all our instances to WSGI.

src/ZServer/HTTPServer.py
```diff
@@ -235,7 +235,7 @@ def get_environment(self, request,
             value=value.strip()
             if h2ehas(key) and value:
                 env[h2eget(key)]=value
-            else:
+            elif "_" not in key:
```
tseaver: Ugh, this module has no existing test coverage!

Can you add a comment explaining the rationale, e.g.:

```suggestion
            elif "_" not in key:  # Headers with underscores, might spoof real ones.
```

tseaver:
ISTM that this fix should be propagated to the [standalone ZServer reposttory](https://github.com/zopefoundation/ZServer/blob/d1f3789e1f1d63e23c86bbfec221ff139bef7aa7/src/ZServer/HTTPServer.py#L250).

src/ZServer/HTTPServer.py
jmuchemb: I changed it a little not to exceed 80 chars.

jmuchemb:
I didn't know the ZServer repository. Does it make sense today ?

dataflake:
Ported to the ZServer standalone repo in zopefoundation/ZServer@2fd0d281721a88de17aecd6c23d7dd19f1843130

viktordick:
This change created a production stop for one of our customers. I will create a separate issue, but the essential problem seems to be that Zope now throws away any headers containing underscores, even those that are already normalized by a Proxy. In our case, we use HAProxy (or, I think in this case, Apache), which already converts headers with dashes and (presumably) throws away headers with underscores. The headers with underscores are now given to Zope, which no longer sees headers like
HTTP_X_SSL_CLIENT_SHA1, HTTP_X_SSL_CLIENT_S_DN, HTTP_X_SSL_CLIENT_SERIAL, HTTP_X_SSL_CLIENT_VERIFY

PR #677, opened 2019-07-25
Remove unneeded if

ale-rt:


ale-rt:
Best reviewed removing whitespaces: https://github.com/zopefoundation/Zope/pull/677/files

ale-rt:
I was really curious why that "if 1" was there. 
I thought that it was some relic from the past that could go away.
Feel free to close this PR if you do not like it!

jugmac00:
> `if 1` is the string I search for to find the place where authentication takes place to debug authentication problems

Maybe you, and a few long term contributors know what `if 1` means, but I think this is a bad choice.

Imho there should be either a comment, or a even a single method/idiom which triggers authentication.

icemac:
I think it would be nice to move the code to a clearly named method. Only removing the if statement does IMHO does not help much.

PR #678, opened 2019-07-26
Small code improvements

ale-rt:


ale-rt:
It seems to me that the proposed version is equivalent and removes the use of `hasattr`.
If I am not wrong `hasattr` tryes to to a `getattr` anyway behind the scenes.

CHANGES.rst
```diff
@@ -28,6 +28,8 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 - Fix broken ZMI when using non-root deployments.
   (`#647 <https://github.com/zopefoundation/Zope/issues/647>`_)
 
+- Small code improvements
```
icemac: This change log entry does not help and can be omitted.

PR #682, opened 2019-08-07
Update to current releases of the dependencies.

icemac:


versions.cfg
```diff
@@ -26,30 +26,30 @@ collective.recipe.cmd = 0.11
...
 pycparser = 2.19
-pyparsing = 2.4.0
+pyparsing = 2.4.2
 python-gettext = 4.0
 readme-renderer = 24.0
 repoze.sphinx.autointerface = 0.8
 requests = 2.22.0
 requests-toolbelt = 0.9.1
-snowballstemmer = 1.2.1
+snowballstemmer = 1.9
```
sallner: That is a big version jump, but there is really no other release on pypi.

Issue #679, opened 2019-07-29
Deprecate / Remove usage of tempstorage


dwt:
Looking at zopefoundation/tempstorage#8 but also the discussion on  zopefoundation/tempstorage#12 and the linked issues it seems to me that the consensus of the senior developers is:

- That tempstorage is pretty thoroughly broken
- That this is common knowledge
- That it really never worked, and therefore nobody can use it correctly
- That e.g. Plone has completely moved away from it for some years because of that knowledge.

I propose that we start going the last step, deprecate it's usage und start removing it from Zope proper, especially from the generator script that seems to advise it's usage every time a new Zope instance is created.

What do you guys think?

dwt:
I have to admit, I was rather hoping that the generator-script change is something that does not have to wait till after 5.0 - as tempstorage is currently broken and will not be usable until that timeline. Also removing a generator feature, does not prevent anyone from using it's functionality if he wants to.

Apart from that: I think deprecation could be something that we can do immediately - although splitting it up to retain the mount-support in it's own plugin and getting rid of just tempstorage is something that we probably want to do after 5.0

icemac:
@dwt Okay, I moved the issue from future back to present.

dwt:
I have created two pull requests, zopefoundation/tempstorage#15 and zopefoundation/tempstorage#15 to move this forward - reviewers and enhancements to these are welcomed.

[Also linking again to zopefoundation/tempstorage#8 and zopefoundation/tempstorage#12 to get GitHub to pick up the links and generate back-references]

dwt:
And another one at #684 - some discussion is probably required. :-)

PR #684, opened 2019-08-12
Remove recommendation of tempstorage.

dwt:
This pull request tries to follow through on #679.

As a first suggestion I have just removed the recommendation of tempstorage from the generated config file. Since this is maybe impractical, an alternative suggestion could be to reformulate the commented out snippet like this:

```
# Uncomment this if you use Products.Sessions and Products.TemporaryFolder
# <zodb_db temporary>
#     <filestorage>
#       path $INSTANCE/var/Data.temporary.fs
#     </filestorage>
#     mount-point /temp_folder
#     container-class Products.TemporaryFolder.TemporaryContainer
# </zodb_db>
```

However, I do not pretend to understand the ramifications of this recommendation (nor have I tested it in depth), so I am putting it here as a discussion point.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,13 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 
 - Resurrect History ZMI tab and functionality
 
+- Removed commented out configuration for tempstorage (and server side 
+  sessions) as that was known not working for ages. This was removed so we do 
+  not lead unsuspecting developers to think that this is the right way to do 
+  session data. See
+  (`#679 <https://github.com/zopefoundation/Zope/issues/679>`_)
+  (`tempstorage#8 <https://github.com/zopefoundation/tempstorage/issues/8>`_)
+  (`tempstorage#12 <https://github.com/zopefoundation/tempstorage/issues/12>`_)
 
 4.1.1 (2019-07-02)
 ------------------
--- a/src/Zope2/utilities/skel/etc/zope.conf.in
+++ b/src/Zope2/utilities/skel/etc/zope.conf.in
@@ -1,3 +1,4 @@
+%define INSTANCE <<INSTANCE_HOME>>
 
 instancehome $INSTANCE
 
--- a/src/Zope2/utilities/skel/etc/zope.conf.in
+++ b/src/Zope2/utilities/skel/etc/zope.conf.in
@@ -1,4 +1,3 @@
-%define INSTANCE <<INSTANCE_HOME>>
 
 instancehome $INSTANCE
 
@@ -8,16 +7,6 @@ instancehome $INSTANCE
    </filestorage>
    mount-point /
 </zodb_db>
-
-# Uncomment this if you use Products.Sessions and Products.TemporaryFolder
-# <zodb_db temporary>
-#     <temporarystorage>
-#       name Temporary database (for sessions)
-#     </temporarystorage>
-#     mount-point /temp_folder
-#     container-class Products.TemporaryFolder.TemporaryContainer
-# </zodb_db>
-
 
 # Directive: locale
 #
--- a/versions.cfg
+++ b/versions.cfg
@@ -57,7 +57,6 @@ snowballstemmer = 1.9
 soupsieve = 1.9.2
 sphinx-rtd-theme = 0.4.3
 sphinxcontrib-websupport = 1.1.2
-tempstorage = 5.0
 toml = 0.10.0
 tox = 3.13.2
 tqdm = 4.32.2


d-maurer:
?rekcäH nitraM? wrote at 2019-8-12 00:05 -0700:
>This pull request tries to follow through on #679.
>
>As a first suggestion I have just removed the recommendation of tempstorage from the generated config file. Since this is maybe impractical, an alternative suggestion could be to reformulate the commented out snippet like this:
>
>```
># Uncomment this if you use Products.Sessions and Products.TemporaryFolder
># <zodb_db temporary>
>#     <filestorage>
>#       path $INSTANCE/var/Data.temporary.fs
>#     </filestorage>
>#     mount-point /temp_folder
>#     container-class Products.TemporaryFolder.TemporaryContainer
># </zodb_db>
>```

You likely would like to remove the `container-class`:
a `TemporaryContainer` is just a "normal" `Folder` with
its own `meta_type` and `icon`. Without a `container-class` spec,
a `Folder` is used -- which seems appropriate when `temp_folder`
is no longer RAM based.

The `mount-point` should likely no longer be named `temp_folder`
as its content is no longer temporary.
However, if you change this name,
then there will be no longer an automatic creation on startup.

Thus, (good) reformulation will not be easy.
I would favour appropriate documentation in `Products.Sessions`
with (potentially) a comment in the configuration file template
to this documentation.

d-maurer:
?rekcäH nitraM? wrote at 2019-8-12 00:14 -0700:
>Maybe recommending usage of `ZODB.MappingStorage.MappingStorage` could also be an interesting suggestion, as it's documentation looks like it might be closer to what tempstorage does - I have however no idea / experience about it's implementation qualities.

It has a reason that `tempstorage` has limited history:
RAM is a fairly scarce resource and therefore, you want to release
unnecessary RAM early. `MappingStore` removes its historical data
only when it is packed (like `FileStorage`).

Sessions tend to change far more frequently than "usual" Zope objects -
and therefore tend to produce far more historical data. Unless
a regular pack is performed, this historical data cumulates.
If you put sessions in a `MappingStore` it occupies precious RAM.
I do not think this is a good idea unless some mechanism ensures
quite frequent packs.

dwt:
> It has a reason that `tempstorage` has limited history:
> RAM is a fairly scarce resource and therefore, you want to release
> unnecessary RAM early. `MappingStore` removes its historical data
> only when it is packed (like `FileStorage`).
> 
> Sessions tend to change far more frequently than "usual" Zope objects -
> and therefore tend to produce far more historical data. Unless
> a regular pack is performed, this historical data cumulates.
> If you put sessions in a `MappingStore` it occupies precious RAM.
> I do not think this is a good idea unless some mechanism ensures
> quite frequent packs.

This would be simple with an accompanying cron job / systemd timer, or maybe even a zope-cron trigger (AFAIK I’ve seen something like this somewhere already, not sure this is still in zope 4).

Do you think that MappingStore could be a sensible replacement if one does not want to write any session data to disk (as a policy requirement) and can ensure a decent packing frequency (maybe once per day, or once per hour)?

Are there any known bugginesses with MappingStorage as there where with tempstorage?

d-maurer:
?rekcäH nitraM? wrote at 2019-8-12 06:47 -0700:
> ...
>Are there any known bugginesses with MappingStorage as there where with tempstorage?

I do not know of known problems.
But I would not use `MappingStorage` by default for sessions (e.g.
as example in the `zope.conf` template).

When you pack, keep a bit of history (to avoid the problems
seen for `tempstorage`).

dwt:
@icemac like this?

dwt:
@maintainers: does it make sense to also remove `tempstorage` from these places?

```diff
diff --git a/src/Zope2/Startup/wsgischema.xml b/src/Zope2/Startup/wsgischema.xml
index e84e32c07..21bbef9df 100644
--- a/src/Zope2/Startup/wsgischema.xml
+++ b/src/Zope2/Startup/wsgischema.xml
@@ -5,7 +5,6 @@
   <!-- type definitions -->
 
   <import package="ZODB"/>
-  <import package="tempstorage" condition="tempstorage"/>
 
   <sectiontype name="environment"
                datatype=".environment"
diff --git a/versions.cfg b/versions.cfg
index 0257fe062..c8a3e21ad 100644
--- a/versions.cfg
+++ b/versions.cfg
@@ -57,7 +57,6 @@ snowballstemmer = 1.9
 soupsieve = 1.9.2
 sphinx-rtd-theme = 0.4.3
 sphinxcontrib-websupport = 1.1.2
-tempstorage = 5.0
 toml = 0.10.0
 tox = 3.13.2
 tqdm = 4.32.2
```

There is also another mention of tempstorage in `requirements-red.txt` that I have no clue if it is neccessary to retain.

d-maurer:
?rekcäH nitraM? wrote at 2019-8-13 06:07 -0700:
>@maintainers: does it make sense to also remove `tempstorage` from these places?
> ... wsgischema.xml, ...

If you remove it from the "*schema.xml" files, then `tempstorage`
can no longer be configured in the zope configuration file.
This would be a very drastic form of deprecation -- forcing
Zope session users to look for a different solution (rather
than suggesting this).

dwt:
@d-maurer Since it is broken - would that actually be a good thing?

d-maurer:
?rekcäH nitraM? wrote at 2019-8-13 07:55 -0700:
>@d-maurer Since it is broken - would that actually be a good thing?

If I would use `tempstorage` (which is not the case) and
I had not seen problems myself, I would not be happy to be forced to use
a different solution.

I recommend to discourage new users to go this route
but not to force existing users to switch if they do not face
problems.

dwt:
@d-maurer from my experiments and the investigations I did on zopefoundation/tempstorage#8 I have to admit that I have a hard time imagining that this can work for anyone. It really looks to me that there is no chance in hell that this could have ever worked for anyone, and that this package _will_ corrupt it's data sooner or later for everyone.

That being said, we did use it in zope 2 / python 2 for a long time without noticing any problems - I just don't get how that was possible.

dwt:
@icemac what about the dependency in versions.cfg? Should we keep that?

icemac:
> @icemac what about the dependency in versions.cfg? Should we keep that?

I don't think so.

d-maurer:
?rekcäH nitraM? wrote at 2019-8-13 23:16 -0700:
> ...
>That being said, we did use it in zope 2 / python 2 for a long time without noticing any problems - I just don't get how that was possible.

A long time ago, I made a review of Zope's session machinery.
When I remember right, then somewhere (I forgot where
precisely) a special measure (I forgot what precisely)
was used to allow it to not
follow the ZODB's typical MVCC (i.e. see state as it was at the transaction
start) but to see more recent state. Something like this is important
for a storage with severely limited history (such as `TemporaryStorage`).

I suspect this no longer works (with modern ZODB versions)
or has been removed in the session machinery (because its importance
has not been recognized).

A way to restore such a possibility would be to let
the `loadBefore` of `TemporaryStorage` ignore its *tid*
parameter and always load the most current (committed) state.
I expect that this will make `TemporaryStorage` again as
usable as it was earlier.

dwt:
@d-maurer I am not sure I know how to achieve that - would you be willing to experiment on this? You could use the reproducer script in zopefoundation/tempstorage#8 to help with that?

In the meantime I'm going to merge this so we give potentially interested parties a reason to get involved.

d-maurer:
?rekcäH nitraM? wrote at 2019-8-14 00:21 -0700:
>@d-maurer I am not sure I know how to achieve that - would you be willing to experiment on this?

No.
Personally, I am not interested in `tempstorage`.

icemac:
@d-maurer Thank you for the suggestion how to possible solve this problem. I put it into a comment in zopefoundation/tempstorage#8  so someone who is interested can experiment with it.

PR #689, opened 2019-08-16
Remove further tempstorage dependencies.

### LINKED ISSUE #668: Installation of current head does not work for Python 2.7
```
git clone git@github.com:zopefoundation/Zope
virtualenv -p /usr/bin/python2.7 .venv
. .venv/bin/activate
pip install zc.buildout
buildout

Getting distribution for 'pathlib2'.
Got pathlib2 2.3.4.
While:
  Installing scripts.
  Getting distribution for 'pathlib2'.
Error: Picked: pathlib2 = 2.3.4
```

HEAD = 20b366eafc766521b1195d3f3897b8c9b6736427

dwt:
This tries to fix #668 by removing further `tempstorage` dependencies.

I am pretty sure that just removing `tempstorage` from the `[docs]` section of extras require is pretty safe as I can't find any mention of `tempstorage` in the documentation. I am a bit more hesitant about the requirements.rtd.txt - but am pretty sure that this is also not required.

Feedback welcome.

### DIFF
--- a/requirements-rtd.txt
+++ b/requirements-rtd.txt
@@ -1,4 +1,3 @@
 repoze.sphinx.autointerface
 Zope
 ZConfig==3.5.0
-tempstorage
--- a/requirements-rtd.txt
+++ b/requirements-rtd.txt
@@ -1,3 +1,4 @@
 repoze.sphinx.autointerface
 Zope
 ZConfig==3.5.0
+tempstorage
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,6 @@ setup(
             'Sphinx',
             'sphinx_rtd_theme',
             'repoze.sphinx.autointerface',
-            'tempstorage',
         ],
     },
     entry_points={


jugmac00:
@dataflake introduced both lines with `tempstorage` when he worked on #571 

Maybe it has to do with documenting the options?
https://zope.readthedocs.io/en/latest/operation.html#zope-configuration-reference

I am not very adept with RTD and Sphinx, so I have to pass on this review.

P.S.: buildout finally works for Zope
- though I'd still like to know why the tests, which also use buildout, passed but the manual buildout failed. I'd like to have failing tests in future when there is a problem with buildout.
- the buildout ends with this message

> requirements: Running '/home/jugmac00/Tests/Zope/bin/zopepy util.py'
Unused options for requirements: 'update-command'.

I think this has nothing to do with this pull request, though.

dwt:
I would argue, that the failing build is actually the test here. It tells you that there is a requirement for a package that is installed, which is not version pinned -> thus the build fails. Adding a test that buildout builds fail if a package is used but not version pinned seems to me to be a redundant test of a buildout feature.

dwt:
@jugmac00 Do you have any Idea where in the docs it mentions tempstorage? I couldn't find anything, so it must be without using the actual name - but then again I don't really know the documentation.

d-maurer:
?rekcäH nitraM? wrote at 2019-8-18 22:41 -0700:
>@jugmac00 Do you have any Idea where in the docs it mentions tempstorage? I couldn't find anything, so it must be without using the actual name - but then again I don't really know the documentation.

With Zope2, sessions have been avaialble automatically and used
`tempstorage` by default. Thus, there has been no need to document
`tempstorage` in entry level documentation. I recommend that experts directly
look at source level documentation (especially interfaces) and do not rely
on other documentation (which is often not complete and/or outdated).

requirements-rtd.txt
```diff
@@ -1,4 +1,3 @@
 repoze.sphinx.autointerface
 Zope
 ZConfig==3.5.0
-tempstorage
```
icemac: This is still required as long as the ZConfig schema contains the `tempstorge` declaration.
See https://readthedocs.org/projects/zope/builds/9537827/ for the failed RTD build.

dwt: done

Issue #659, opened 2019-06-22
Update documentation to Zope 5


icemac:
- [x] chapter "What's new"
- [x] chapter "Migration to Zope 5"
- [x] drop chapter "Zope 4 migration"
- [x] chapter "Installing Zope" --> Prerequisites
- [x] remove usage of ZServer see #592

jugmac00:
Afaik in the current documentation there are several mentions of Python 2, eg. https://zope.readthedocs.io/en/latest/zdgbook/ComponentsAndInterfaces.html#querying-an-interface

This should be revised, too.

PR #690, opened 2019-08-27
Drop Python 2 support

### LINKED ISSUE #592: Remove ZServer support
- [x] from `setup.py`
- [x] from buildout.cfg
- [x] from code
- [x] from documentation

icemac:
fixes #592.

Open issues:

- [x] Grep for `2.7`, `PY2` and drop Python 2 support code see #692
- [x] drop dependency on six see #692
- [X] Create a new `What's new section in the documentation` see #659
- [ ] maybe remove server directives from etc/example.conf
- [ ] Find out if ZApplicationWrapper.__bobo_traverse__ is actually only called by ZServer, see https://github.com/zopefoundation/Zope/blob/b9a346903fe4a6874face0930876184ac321853e/src/App/ZApplication.py#L45-L47

Maybe these open issues should be handled by later pull requests.


icemac:
Superseded by #696.

PR #691, opened 2019-08-29
Drop Python 2.7 support in package configuration.

icemac:
This PR prepares the master branch to start working on Zope 5 aka drop Python 2 support.

(It is only the first step dropping Python 2 from the tests and preparing the change log.)

- [x] Move the 4.x branch point after the 4.1.2 release.


icemac:
@dataflake wrote:
> The change itself looks fine, but can I request that we release 4.1.2 first?

That's a good idea, so we do not have the currently unreleased changes in both release.

> I was hoping to go over the outstanding issues this weekend for any low-hanging fruit for 4.1.2 as well.

Cool. So let's have the 4.1.2 release before merging this PR.

dataflake:
@icemac Could you create a AccessControl release and update the dependency in Zope? That will fix a few bugs across several packages when people try to access the result of dict.items/keys/values.

icemac:
@dataflake I created #694 to use AccessControl 4.1.

dataflake:
As far as I am concerned this is good to go after creating a Zope 4 branch.

icemac:
I (re-) created a 4.x branch just now, see https://github.com/zopefoundation/Zope/tree/4.x

PR #693, opened 2019-08-30
Reuse ``zope.publisher.http.splitport`` instead of defining our own

### LINKED ISSUE #683: Use splitport implementation from zope.publisher
In zopefoundation/zope.publisher#48 @mgedmin implemented the `splitport` function in a much easier way than I did using the suggested `urlparse` function in https://github.com/zopefoundation/Zope/blob/e9f10101b8a4db64a2440e3a51561a7805ea44f5/src/ZPublisher/HTTPRequest.py#L110-L128

Once the PR is merged and released we could use his function as we already depend on `zope.publisher`.

dataflake:
Fixes #683 

The behavior of the zope.publisher version is slightly different, so for the VirtualHostMonster code it was not a full drop-in replacement.

### DIFF
--- a/src/Products/SiteAccess/VirtualHostMonster.py
+++ b/src/Products/SiteAccess/VirtualHostMonster.py
@@ -10,12 +10,12 @@ from App.special_dtml import DTMLFile
 from OFS.SimpleItem import Item
 from Persistence import Persistent
 from zExceptions import BadRequest
+from zope.publisher.http import splitport
 from ZPublisher.BaseRequest import quote
 from ZPublisher.BeforeTraverse import NameCaller
 from ZPublisher.BeforeTraverse import queryBeforeTraverse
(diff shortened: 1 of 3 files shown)


src/Products/SiteAccess/VirtualHostMonster.py
```diff
@@ -149,7 +149,9 @@ def __call__(self, client, request, response=None):
                 stack.pop()
                 protocol = stack.pop()
                 host = stack.pop()
-                request.setServerURL(protocol, *splitport(host))
+                hostname, port = splitport(host)
+                port = (port and int(port)) or None
```
icemac:
```suggestion
                port = int(port) if port else None
```

dataflake: Agreed, thanks

Issue #649, opened 2019-06-13
FTP or WebDav access


georgpfolz:
I'm wondering if it will still be possible to access the ZODB via FTP, WebDav or maybe even SFTP in future Zope versions without ZServer. Or is there a way already which I missed?

I hope this is the right place to ask this kind of questions, if not please refer me to the right place to go.

d-maurer:
Georg Pfolz wrote at 2019-6-13 05:14 -0700:
>I'm wondering if it will still be possible to access the ZODB via FTP, WebDav or maybe even SFTP in future Zope versions without ZServer. Or is there a way already which I missed?
>
>I hope this is the right place to ask this kind of questions, if not please refer me to the right place to go.

WebDAV operates in fact on top of HTTP. There should be no big problem
to use it in a WSGI environment (i.e. without `ZServer`).

Difficulties result from the fact, that WebDAV has slightly
different expectations with respect to server behaviour than
a typical browser:

 * when you access an object over WebDAV, you expect to get
   the object's raw content;
   when you access an object via a browser, you typically
   want a "rendered" version (e.g. with navigation and actions)

 * WebDAV expects HTTP authentication; for a browser,
   typically a more user friendly form based authentication is
   used.

The `ZServer` WebDAV (source) server marked the request to
make different behaviour for the two use cases possible.
But there are other possibilities to mark the request in a similar way.

For example, one could use a view '@@webdav' to mark the request
on traversal as a WebDAV request and request WebDAV friendly behaviour.
With this approach a URL of the form `<host>/<path>` could
provide browser friendly access to the object identified by `<path>`
while `<host>/@@webdav/<path>` would give WebDAV friendly access
to the same object.

dataflake:
There are no plans to support any of this in Zope without ZServer because Zope has essentially outsourced all network protocol services to third-party packages that use WSGI to talk to Zope.

thet:
@datakurre worked on a ZServer implementation based on Twisted, which also supports http2, WebDAV and the like. See here: https://community.plone.org/t/moonshot-for-migrating-zserver-to-python-3-at-plog-2019/8191/19

It's not finished, but his statements sound very promising.

@d-maurer instead of using a specific ``@@webdav`` browser view, I'd prefer to mark the request as WebDAV request based on the ``webdav://`` protocoll. By adapting a view to a webdav request, we'd have the same flexibility like with normal browser views.

datakurre:
[...]

PR #696, opened 2019-09-09
Rip out ZServer support.

### LINKED ISSUE #592: Remove ZServer support

icemac:
Open questions/tasks:

- [x] Should zserver directives be removed from etc/example.conf? ← they already have been removed
- [x] Find out if `ZApplicationWrapper.__bobo_traverse__` is actually only called by ZServer, see https://github.com/zopefoundation/Zope/blob/b9a346903fe4a6874face0930876184ac321853e/src/App/ZApplication.py#L45-L47 ← yes, it is actually used by a ZServer based Zope but not by a WSGI based one.

fixes #592


src/ZPublisher/Publish.py
```diff
@@ -14,28 +14,7 @@
...
-    publish_module_standard=(
-        'ZServer.ZPublisher.Publish:publish_module_standard'),
-    set_default_debug_mode=(
-        'ZServer.ZPublisher.Publish:set_default_debug_mode'),
-    set_default_authentication_realm=(
-        'ZServer.ZPublisher.Publish:set_default_authentication_realm'),
-)
-
-# BBB Zope 5.0
+# BBB Zope 6.0
```
dataflake: Was this not ready for removal?

icemac: You are right it can be removed in 5.0 but in another PR.

d-maurer:
Michael Howitz wrote at 2019-9-9 23:10 -0700:
> ...
>-# BBB Zope 5.0
>+# BBB Zope 6.0
>
>You are right it can be removed in 5.0 but in another PR.

Please keep "webdav" support (which apparently accidentally
was moved into the `ZServer` package).

icemac:
Zope 4 using WSGI does not support WebDAV. There is #649 track re-enabling it. It would be a huge benefit if Zope 5 could support WebDAV again.
I personally neither have the need nor resources to re-implement WebDAV in Zope, so any help is welcome.

d-maurer:
Michael Howitz wrote at 2019-9-10 23:19 -0700:
>Zope 4 using WSGI does not support WebDAV. There is #649 track re-enabling it. It would be a huge benefit if Zope 5 could support WebDAV again.
>I personally neither have the need nor resources to re-implement WebDAV in Zope, so any help is welcome.

"WebDAV" used to be a top level package in Zope 2.
I assume that it was put below `ZServer` under the (wrong) assumption
it were dependent on `ZServer`.
Could you move it into a separate "zopefoundation/webdav"?

`OFS` and (I think) `ZPublisher` have special handling for `WebDAV`.
It would be quite difficult (complex monkey patching) to have
`WebDAV` without this special code.
In Zope 4, this was dependent on `ZServer` and it therefore risks
to get lost with the outphasing of `ZServer`.
The dependency should become one of `WebDAV`.

dataflake:
I'm sure no one would complain if you just went ahead and did this yourself. 😉

d-maurer:
Jens Vagelpohl wrote at 2019-9-11 03:12 -0700:
>I'm sure no one would complain if you just went ahead and did this yourself. ?

I do not yet know how I can move parts from one git repository
(including its history) to a different or a new one.

dataflake:
There's a howto in the GitHub help pages: https://help.github.com/en/articles/splitting-a-subfolder-out-into-a-new-repository

drfho:
Hi @icemac , @dataflake ,
today our Zope-latest testing system (working with Products.CMFCore) cannot start anymore, because some CMFCore-methods expect a variable OFS.bbb.HAS_ZSERVER. 
This comaptibility problem can quickly be solved by providing a reduced module OFS.bbb just for setting the variable:
[bbb_py.txt](https://github.com/zopefoundation/Zope/files/3600325/bbb_py.txt)
`HAS_ZSERVER = False`
Or do you prefer fixing Products.CMFCore?

dataflake:
Zope `master` is the precursor of Zope 5. Since this is a major release backwards incompatibilities are OK. A lot of the removed code had already been emitting DeprecationWarnings so it is up to the packages that want to use the deprecated code to get fixed. Fix CMFCore.

icemac:
@d-maurer I'd prefer to have `webdav` back here in the Zope repository. I am not sure if it could be  used standalone. As you said using it requires some changes in Zope which I do not want to have conditionally again.

Maybe only parts of 24d566b475ec60a5f872d39cdb63cb730fdc0f82 have to be reverted to get it back including its history.

dataflake:
FYI: Instead of trying to fiddle with some git magic it is sufficient to simply copy the contents of the files in `src/webdav` from that commit diff (or in case they were updated in ZServer, from there) back into their old location. When you do a `git add` their history will be right back.

d-maurer:
Michael Howitz wrote at 2019-9-11 23:17 -0700:
>@d-maurer I'd prefer to have `webdav` back here in the Zope repository. I am not sure if it could be  used standalone. As you said using it requires some changes in Zope which I do not want to have conditionally again.

I have no objections against `webdav` directly in the Zope repository.

PR #699, opened 2019-09-10
Drop Python 2 support Part 1

icemac:
Dropping references to 

* 2.7
* PY2

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -16,8 +16,7 @@ available:
 - A supported version of Python, including the development support if
   installed from system-level packages.  Supported versions include:
 
-  * 2.7
-  * 3.5 - 3.8
+  * 3.5 up to 3.8
 
 - Zope needs the Python ``zlib`` module to be importable.  If you are
   building your own Python from source, please be sure that you have the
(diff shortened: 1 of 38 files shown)


docs/INSTALL.rst
```diff
@@ -16,8 +16,7 @@ available:
 - A supported version of Python, including the development support if
   installed from system-level packages.  Supported versions include:
 
-  * 2.7
-  * 3.5 - 3.8
+  * 3.5 up to 3.8
```
dataflake: Don't we use "Python 3.5 and higher" in other places instead of limiting to 3.8?

docs/zope4/news.rst
```diff
@@ -29,15 +29,14 @@ found in this release.
 
 Extended Python version support
 -------------------------------
-Zope 4 supports Python 2.7 and Python 3.5 up to Python 3.8.
+Zope 5 supports Python 3.5 up to Python 3.8.
```
dataflake: See above - saying "Python 3.5 and up" may be better

docs/INSTALL.rst
icemac: I explicitly changed it some weeks ago to mention the currently supported maximum version of Python as it tells about the actual current status of supported versions. (It is more explicit than the implicit meaning that only Python 3 versions are supported which are known at the time when this part of the documentation was written.)

Yes, this requires to adapt the documentation once we support Python 3.9, but this is also an explicit step.

PR #709, opened 2019-10-03
Fix VirtualHostMonster not being able to set mappings under Python 3.

### LINKED ISSUE #708: VHM mappings can't be set
I have a Plone 5.2 site installed on top of Zope 4.1.1 / Python 3.7

Setting a VHM map in http://localhost:8080/virtual_hosting/manage_edit like ``l1/Plone`` leads to this error:

```
2019-10-03 09:57:43,956 ERROR   [Zope.SiteErrorLog:18][waitress] 1570089463.9533460.9165882821177458 http://localhost:8080/virtual_hosting/set_map
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 155, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 337, in publish_module
  Module ZPublisher.WSGIPublisher, line 255, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module Products.PDBDebugMode.wsgi_runcall, line 60, in pdb_runcall
  Module Products.SiteAccess.VirtualHostMonster, line 107, in set_map
TypeError: catching classes that do not inherit from BaseException is not allowed
```
It's because of the ``'LineError'`` which is not a real exception and I couldn't find any information about.

This is just the error with the catching branch. The real error happens earlier in Products/SiteAccess/VirtualHostMonster:
```
  69                     pp = filter(None, path.split('/'))                                                                                                                                                                                   
  70                     if pp:                                                                                                                                                                                                               
  71  ->                     obpath = pp[:]          
TypeError: 'filter' object is not subscriptable
(Pdb++) pp
<filter object at 0x7f3a442e4550>
(Pdb++) list(pp)
['Plone2']
```

thet:
Fixes #708

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.0a1 (unreleased)
 ------------------
 
+- Fix VirtualHostMonster not being able to set mappings under Python 3. Fixes #708
 - Improve documentation for Zope's error logging services.
 
 Backwards incompatible changes
--- a/src/Products/SiteAccess/VirtualHostMonster.py
+++ b/src/Products/SiteAccess/VirtualHostMonster.py
@@ -104,7 +104,7 @@ class VirtualHostMonster(Persistent, Item, Implicit):
                 if hostname not in host_map:
                     host_map[hostname] = {}
                 host_map[hostname][port] = pp
-            except Exception as msg:
+            except ValueError as msg:
                 line = '%s #! %s' % (line, msg)
             new_lines.append(line)
         self.lines = tuple(new_lines)
--- a/src/Products/SiteAccess/VirtualHostMonster.py
+++ b/src/Products/SiteAccess/VirtualHostMonster.py
@@ -65,7 +65,7 @@ class VirtualHostMonster(Persistent, Item, Implicit):
                 except Exception:
                     raise ValueError(
                         'Line needs a slash between host and path: %s' % line)
-                pp = filter(None, path.split('/'))
+                pp = list(filter(None, path.split('/')))
                 if pp:
                     obpath = pp[:]
                     if obpath[0] == 'VirtualHostBase':
@@ -104,7 +104,7 @@ class VirtualHostMonster(Persistent, Item, Implicit):
                 if hostname not in host_map:
                     host_map[hostname] = {}
                 host_map[hostname][port] = pp
-            except 'LineError' as msg:
+            except Exception as msg:
                 line = '%s #! %s' % (line, msg)
             new_lines.append(line)
         self.lines = tuple(new_lines)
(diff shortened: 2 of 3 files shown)


src/Products/SiteAccess/VirtualHostMonster.py
```diff
@@ -104,7 +104,7 @@ def set_map(self, map_text, RESPONSE=None):
                 if hostname not in host_map:
                     host_map[hostname] = {}
                 host_map[hostname][port] = pp
-            except 'LineError' as msg:
```
thet: @icemac You've added Products.SiteAccess with this commit.
I don't find the original repo anymore and can't do a git blame there.
What's meant with a ``'LineError'``?

d-maurer:
Johannes Raggam wrote at 2019-10-3 02:35 -0700:
>thet commented on this pull request.
>> @@ -104,7 +104,7 @@ def set_map(self, map_text, RESPONSE=None):
>                 if hostname not in host_map:
>                     host_map[hostname] = {}
>                 host_map[hostname][port] = pp
>-            except 'LineError' as msg:
>
>@icemac You've added Products.SiteAccess with this commit.
>I don't find the original repo anymore and can't do a git blame there.
>What's meant with a ``'LineError'``?

`'LineError'` must be very old -- from a time when Python exceptions have
been strings. I think, modern Python versions no longer support string
exceptions. Should this be true, then there is no longer any way to
raise a `'LineError'` and the `except 'LineError'` could go away.

You may search the `SiteAccess` source for `Error` and try to
find out whether the raising part as meanwhile been modernized
(i.e. the exception transformed from a string to a subclass
of `BaseException`).

thet:
@d-maurer tnx for the explanation! I couldn't find any references to LineError. IMO Exception is sufficient here, even it's normally bad practice to use such an generic Exception type.

src/Products/SiteAccess/VirtualHostMonster.py
icemac: The code was actually added in f58021f472a0d7490b6913c8168d723f50f269cd 18 years ago. The code raised the `LineError` exception those days. In fb33a42d9bc934c540829ceb9373dda1d76551ac it was changed to `ValueError` so the except statement should be changed, too.

CHANGES.rst
```diff
@@ -9,6 +9,7 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.0a1 (unreleased)
 ------------------
 
+- Fix VirtualHostMonster not being able to set mappings under Python 3. Fixes #708
```
```suggestion
- Fix VirtualHostMonster not being able to set mappings under Python 3. 
  (`#708 <https://github.com/zopefoundation/Zope/issues/708>`_)

```

src/Products/SiteAccess/VirtualHostMonster.py
```diff
@@ -104,7 +104,7 @@ def set_map(self, map_text, RESPONSE=None):
...
+            except Exception as msg:
```
icemac:
```suggestion
            except ValueError as msg:
```

This would restore the original behavior. (See my other comment.)

icemac:
Additionally the pylint errors should be fixed.

PR #713, opened 2019-10-22
Set ``REMOTE_USER`` in wsgi environ using Zope user authentication

mamico:
Set REMOTE_USER in wsgi publisher for next apps in wsgi pipeline. 
e.g. useful for have user in access logs, ...


jugmac00:
I'd love to see the WSGI documentation updated with an example how to use this new environment variable.

https://zope.readthedocs.io/en/latest/wsgi.html

dataflake:
The REMOTE_USER variable is documented in the WSGI specification itself. You don't "use" it in Zope, anyway. This is purely for the benefit of other code in the WSGI pipeline.

Issue #674, opened 2019-07-19
Python 3: unrestrictedTraverse checks for str


reinhardt:
unrestrictedTraverse checks whether path is an instance of ```str```:

https://github.com/zopefoundation/Zope/blob/master/src/OFS/Traversable.py#L175

From the comment and the way it works with python 2 it seems that unicode strings are forbidden without being explicitly rejected. Everything that is not a binary string is just assumed to be a sequence of binary strings.

However, with python 3, ```str``` is the unicode type, so now you *have* to pass unicode for it to work. That seems unintended.

I guess the solution is to replace ```str``` with ```six.binary_type```, or am I missing something?

ale-rt:
`six.binary_type` will not work on Python3, you need something like the `safe_nativestring` function defined here: https://github.com/plone/Products.CMFPlone/blob/bd98bd05a3eaf1c5b9579a7fb49da9c37f6721fd/Products/CMFPlone/utils.py#L525

reinhardt:
In what way do you think it won't work?
Note that I'm not suggesting to convert the string, only to fix the isinstance check.

ale-rt:
Because on Python3:
```
>>> isinstance(b"", str)
False
```

reinhardt:
No, I mean, why will this not work:
```
if isinstance(path, six.binary_type):
```

ale-rt:
That will work on Python2 only because on Python3 path is expected to be a `str` instance in both Python2 and Python3.

reinhardt:
If it is expected to be ```str``` in both then it is effectively expected to be something different in Python 3 than in Python 2. Can you explain the reasoning behind this?

ale-rt:
> Can you explain the reasoning behind this?

Not really, at least in a few sentences :)

reinhardt:
OK, but you're saying it's correct like this? So you could also put it like this then:
```
if six.PY2 and isinstance(path, six.binary_type) or six.PY3 and isinstance(path, six.text_type):
```
which is of course longer but doesn't look like a mistake so much.

Also, the comment "Unicode paths are not allowed" is wrong, isn't it? In Python 3 unicode paths are just fine IIUC.

ale-rt:
I think it is correct for this to be a native string and there is no issue with the code and I do not feel it being a mistake.

> Also, the comment "Unicode paths are not allowed" is wrong, isn't it? In Python 3 unicode paths are just fine IIUC.

[...]

PR #720, opened 2019-10-28
Update comment

ale-rt:
Refs #674

### DIFF
--- a/src/OFS/Traversable.py
+++ b/src/OFS/Traversable.py
@@ -172,7 +172,7 @@ class Traversable(object):
             return self
 
         if isinstance(path, str):
-            # bytes paths are not allowed
+            # Only native strings are allowed
             path = path.split('/')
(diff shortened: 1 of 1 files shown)


jugmac00:
@ale-rt Thanks for takting your time!

This pr should be against the Zope 4 branch ( https://github.com/zopefoundation/Zope/tree/4.x ) - as master only supports Python 3, which means there should be also an updated comment, but only referring to Python 3 then.

ale-rt:
Done in #721 and I updated the PR!

ale-rt:
In my opinion the comment was better before, because it was more coherent with what we have now.
Anyway I will submit a patch that will raise a `ValueError` is the type is wrong.
Now that method also accepts not orderable objects like sets or dicts and IMHO that should be fixed because it is not meant to work like that...

PR #722, opened 2019-10-31
Make unrestrictedTraverse more pedantic about the path

### LINKED ISSUE #674: Python 3: unrestrictedTraverse checks for str

ale-rt:
The method ``unrestrictedTraverse`` raises an error when the argument ``path`` is not something it can work with

Refs #674 #722


src/OFS/Traversable.py
```diff
@@ -168,15 +168,20 @@ def unrestrictedTraverse(self, path, default=_marker, restricted=False):
...
+            if isinstance(path, str):
+                path = path.split('/')
+            else:
+                path = list(path)
+                for part in path:
+                    if not isinstance(part, str):
+                        raise TypeError(
+                            "path should be a string or an iterable of strings"
+                        )
+
```
dataflake: I suggest you retain the original order and drop this code below the `if not path` test.

ale-rt: Instead I would leave the patch as it is because path changes under `if path` and so it make sense to reevaluate it after the first if.

dataflake: You're right that the path may change, but I see no possibility in that code that it can change from a truth value to a non-truth value. That is not possible.

ale-rt: Well, we would need an object that is truish but casted to a list returns an empty list...
```
ipdb> self.context.invokeFactory("Folder", "folder")
'folder'
ipdb> bool(self.context["folder"])
True
ipdb> !list(self.context["folder"])
[]
```

Not that I am encouraging people to use a folder as an argument to `unrestrictedTraverse`, but you know:

> I've seen things you people wouldn't believe...

So in my opinion leaving the patch as it is is better :)

ale-rt: @dataflake did I make you change your mind?

dataflake: I don't think the argument that people might throw garbage at this method is valid. The method has very clear documentation about what it accepts. Everything else simply deserves to lose. You cannot guard against everything, and attempting to do that is a non-goal.

The original intent of your changes was to tell people if they sent in the wrong kind of string. But you conflate that with changing the control flow, which was very clear and obvious before your changes. Please keep it that way.

dataflake:
To break the stalemate I have reverted the control flow change.

PR #705, opened 2019-09-19
Drop dependency on six.

icemac:
See #692

PR #724, opened 2019-11-03
Run ``pyupgrade`` against the code base, remove ``six`` dependency

dataflake:
These changes are the result of running `pyupgrade --py3-plus --keep-percent-format` against the codebase, with some cleanup afterwards.

I chose the `--keep-percent-format` flag because that removed a whole lot of noise. Even so, the PR is pretty big.

Since this work got most of the way to removing the `six` dependency I pushed that all the way to completion, so this PR also replaces #705 . The only "difficulty" was replacing six.reraise, for which there is no simple replacement in Python 3. So I copied its implementation into the one module where it was used.

As a refresher, here's what `pyuograde` does: https://github.com/asottile/pyupgrade

Closes #705.


src/Zope2/utilities/copyzopeskel.py
```diff
@@ -182,7 +182,7 @@ def copyskel(sourcedir, targetdir, uid, gid, **replacements):
                 copydir(root, dirs, files, targetdir, replacements, uid, gid)
         finally:
             os.chdir(pwd)
-    except (IOError, OSError) as msg:
+    except OSError as msg:
```
icemac: Cool, I had to look up that `IOError` has been merged into `OSError`, see https://docs.python.org/3/library/exceptions.html#OSError

dataflake:
Yes, I was very impressed myself. And the followup PRs are a good idea as well.

Issue #730, opened 2019-11-08
ZMI Security tab: Make permissions list fixed when scrolling left/right


dataflake:
The ZMI Security tab gets unwieldy really fast if the list of roles requires scrolling to the right. Please change the form so that the leftmost table columns (the first column with the checkboxes and the second column with the permission names) remain in view when scrolling to the right.

@drfho Could you take a look at this?

dataflake:
I have merged PR #731 the way it was right now because I need it for 4.1.3. The Safari issue remains open.

dataflake:
The solution works well, thanks for your work!

Issue #729, opened 2019-11-07
ZMI: dbmain form, ordered folder icon


drfho:
Hi, found 2 small css probs in ZMI:

![zmi_cssprbs](https://user-images.githubusercontent.com/29705216/68435160-03a6da00-01bb-11ea-96a8-51a2e0f1d16c.png)

I will provide a PR for that...  
got some connection issues with VSCode now - so the  code diff here in advance:

![zms_base_css](https://user-images.githubusercontent.com/29705216/68436962-c729ad00-01bf-11ea-8232-fec4d9b00f76.png)
```
/Zope/src/zmi/styles/resources/zmi_base.css
L621
%< - - - - - -
/* CUSTOM ICON: Folder (Ordered) */
i.fa-folder.zmi-icon-folder-ordered:after {
	content: "\f338";
	font-family: "Font Awesome\ 5 Free",Fontawesome;
	font-weight:900;
	margin: .475rem 0 0 -.7rem;
	position:absolute;
	font-size:14px;
}
.zmi .breadcrumb i.fa-folder.zmi-icon-folder-ordered:after {
	font-size:12px;
}
%< - - - - - -
```

![dbMain](https://user-images.githubusercontent.com/29705216/68436971-cf81e800-01bf-11ea-8f79-3d927a6286de.png)
```
/Zope/src/App/dtml/dbMain.dtml
L49
%< - - - - - -
  <div class="zmi-controls mb-5">
    <form action="&dtml-URL1;/manage_pack" method="post">
      <div class="input-group">
        <input class="btn btn-primary" id="pack" type="submit"
          name="submit" value="Pack" />
        <input class="form-control col-sm-1 col-3 ml-2 text-right" type="text"
          name="days:float" value="0" size="3" />
        <label class="mt-2 ml-2">days</label>
      </div>
      <small class="form-text text-muted">
        Pack: Remove previous object revisions older than the selected
        number of days.
      </small>
    </form>
  </div>
%< - - - - - -
```

![zmi_changes](https://user-images.githubusercontent.com/29705216/68436976-d6a8f600-01bf-11ea-8464-a113d7c783e4.png)

dataflake:
@drfho Don't worry about the PR, I copied the changes from your description and tested them manually. It's all done.

dataflake:
I'll need to reopen this:

[...]

PR #731, opened 2019-11-08
Issues #729, #730: ZMI Security tab: Make permissions list fixed when scrolling left/right

drfho:
Hi @dataflake ,
 I added a JS check of the table size for jumping into the compressed/responsive view of that table, if it's standard/desktop view is too  broad for the window size. I think in most cases this will be perfect:
src/OFS/dtml/access.dtml
src/zmi/styles/resources/zmi_base.css

zmi_userperms_table

If the table is super-super-broad, a column fixing would be helpful, but in the mobile view, it does not work in general, because the the first two cols are too broad. So I need to think a little more about it. 
Anyway another modification will be added to the one shown here, so this change could be merged already.

Thanks for the hint.
Best regards
f

PS: Maybe you could add the .gitignore conf for ".vscode"?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ pip-selfcheck.json
 Pipfile
 Pipfile.lock
 pyvenv.cfg
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 .project
 .pydevproject
 .tox
+.vscode
 /bin/
 /build/
 /_build/
@@ -32,4 +33,3 @@ pip-selfcheck.json
 Pipfile
 Pipfile.lock
 pyvenv.cfg
-.vscode
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,8 @@ matrix:
           env: TOXENV=py36
         - python: "3.7"
           env: TOXENV=py37
-          dist: xenial
         - python: "3.8"
           env: TOXENV=py38
-          dist: xenial
 
 install:
     - travis_retry pip install -U pip setuptools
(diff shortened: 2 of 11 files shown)


dataflake:
@drfho Friendly reminder: When you create a branch from `master`, **PLEASE** do a `git fetch` and a `git pull` first so your local `master` is fully up to date. I didn't notice that and kept adding commits on this branch before realizing what happened. I had to rebase the branch and fix a bunch of conflicts.

That said, this solution looks nice and works for me on Chromium and Firefox, but not on Safari. I'd also still like to add something like I suggested where you will always have the permission name in view, no matter how wide the table gets. In my case, we currently have about 50 roles...

drfho:
Hi @dataflake,
seems we worked in parallel: actually I tried hard to display the first two cols of the permission table as fixed. But the table is too komplex; nothing worked.  So I added the permission-name to the title (as data-title, shown by CSS), similar to your proposal based on the standard-tooltip.

security

This is a simple solution and even your 50 roles case should work with it. 
Besides, I found out, that a normal HDMI screen shows the table with more than 40 roles, if you scale the viewport to 80%.

security_42roles

For some small css fixes  and the safari prob I will add a new PR on the base of your merged master.
Many thanks
f

dataflake:
No, thank you for your work.

I expanded on the tooltip for the table cells with the checkbox by adding both the role and permission to it, by the way. I am happy with the way it is  now, just the Safari issueshould be fixed at some point.

Funny enough, in our specific setup much of the issue has disappeared with a pending complete rewrite of how that site handles roles and permissions, which brought it down to about 25 roles from several hundred.

dataflake:
P.S.: Just noticed you had done the exact same thing with the tooltip, just formatted differently. Mine just used a line break in the title and qualifies both lines with what they are to be super clear.

drfho:
yes, your variant is better.

Issue #734, opened 2019-11-22
 The 'Paste' distribution was not found and is required by the application


jugmac00:
There seem to be a bunch of related issues, but all closed #651 #452 #664 

I followed the official installation document at 
https://zope.readthedocs.io/en/latest/INSTALL.html
https://zope.readthedocs.io/en/latest/operation.html

**Steps to reproduce**
```
python3.7 -m venv zope
cd zope/
bin/pip install -U pip
bin/pip install Zope==4.1.2 -c https://zopefoundation.github.io/Zope/releases/4.1/constraints.txt
bin/mkwsgiinstance -d .
bin/runwsgi -v etc/zope.ini
```

The last command fails
```
bin/runwsgi -v etc/zope.ini 
Traceback (most recent call last):
  File "bin/runwsgi", line 8, in <module>
    sys.exit(main())
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/Zope2/Startup/serve.py", line 248, in main
    return command.run()
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/Zope2/Startup/serve.py", line 189, in run
    global_conf=vars)
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/Zope2/Startup/serve.py", line 217, in loadapp
    return loadapp(app_spec, name=name, relative_to=relative_to, **kw)
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 253, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 277, in loadobj
    global_conf=global_conf)
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 302, in loadcontext
    global_conf=global_conf)
  File "/home/jugmac00/Tests/Zope4.1.2.installation/zope/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 326, in _loadconfig
    return loader.get_context(object_type, name, global_conf)
[...]

Issue #664, opened 2019-06-28
Minimum buildout.cfg is not enough


icemac:
history:
```
 2161  python3.8 -m venv zope
 2162  cd zope 
 2164  cp ../buildout.cfg .
 2165  bin/pip install -U pip zc.buildout
 2166  bin/buildout
 2167  bin/mkwsgiinstance 
 2168  vim etc/zope.ini
 2169  bin/runwsgi -v etc/zope.ini

```
and buildout.cfg:

```
[buildout]
extends = https://zopefoundation.github.io/Zope/releases/4.1/versions-prod.cfg
parts = zopescripts

[zopescripts]
recipe = zc.recipe.egg
interpreter = zopepy
eggs =
  Zope
#  Paste
  Products.TemporaryFolder
  Products.ExternalMethod
  Products.PythonScripts
  Products.Sessions
  Products.SiteErrorLog
  Products.MailHost
  Products.ZSQLMethods
  Products.ZMySQLDA

```

_Originally posted by @ghiodor in https://github.com/zopefoundation/Zope/issues/452#issuecomment-506225698_

icemac:
Even using the minimum [buildout.cfg from the documentation](https://zope.readthedocs.io/en/latest/INSTALL.html#minimum-configuration) is not enough.

The traceback is:

```python
Traceback (most recent call last):
  File "bin/runwsgi", line 85, in <module>
    sys.exit(Zope2.Startup.serve.main())
  File "/.../Zope-4.1-py3.8.egg/Zope2/Startup/serve.py", line 248, in main
    return command.run()
  File "/.../Zope-4.1-py3.8.egg/Zope2/Startup/serve.py", line 188, in run
    app = self.loadapp(app_spec, name=app_name, relative_to=base,
  File "/.../Zope-4.1-py3.8.egg/Zope2/Startup/serve.py", line 217, in loadapp
    return loadapp(app_spec, name=name, relative_to=relative_to, **kw)
  File "/.../PasteDeploy-2.0.1-py3.8.egg/paste/deploy/loadwsgi.py", line 253, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/.../PasteDeploy-2.0.1-py3.8.egg/paste/deploy/loadwsgi.py", line 275, in loadobj
    context = loadcontext(
  File "/.../PasteDeploy-2.0.1-py3.8.egg/paste/deploy/loadwsgi.py", line 299, in loadcontext
    return _loaders[scheme](
[...]

PR #735, opened 2019-11-22
Add missing Paste dependency.

icemac:
Otherwise Zope won't start up when installed with `pip` and the `constraints.txt` file.

When accepted this PR should also be ported to the 4.x branch.

Relates to #734.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,8 +55,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Improve documentation for Zope's error logging services.
 
-- Add ``Paste`` as install dependency, otherwise it won't start when installed
-  with `pip` and the `constraints.txt` file.
+- Add ``Paste`` as ``extras_require`` dependency to pull in ``Paste`` when 
+  installing with `pip` and `constraints.txt` to prevent startup errors.
   (`#734 <https://github.com/zopefoundation/Zope/issues/734>`_)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,7 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Add ``Paste`` as ``extras_require`` dependency to pull in ``Paste`` when 
   installing with `pip` and `constraints.txt` to prevent startup errors.
+  This requires adding the ``[wsgi]`` extra in the egg specification.
   (`#734 <https://github.com/zopefoundation/Zope/issues/734>`_)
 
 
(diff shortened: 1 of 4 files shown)


icemac:
Hm, in #664 we decided against having `Paste` in the install dependencies. Maybe we should think over having `translogger` in the default WSGI conf. @dataflake What do you think?

dataflake:
The problem is (from what I remember) without translogger you have no logging at all. That's a very unfriendly user experience upon first install

dataflake:
Maybe there's some othr WSGI logger out there that is small enough and doesn't drag all of Paste with it so it's OK to just have that in the install_requires?

CHANGES.rst
```diff
@@ -51,6 +51,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Improve documentation for Zope's error logging services.
 
+- Add ``Paste`` as install dependency, otherwise it won't start when installed
+  with `pip` and the `constraints.txt` file.
+  (`#734 <https://github.com/zopefoundation/Zope/issues/734>`_)
```
dataflake: This is misleading. It only won't start if you used the built-in `mkwsgiinstance` to create the WSGI server `.ini` file.

dataflake:
@icemac I had a different idea that makes the `pip` install work while still making it clear that `Paste` just isn't an installation dependency. I moved it into an extra that I named `wsgi`, which is then used for the install. It's now used inside the buildout configuration and the installation instructions for a `pip` install have it as well.

CHANGES.rst
dataflake: Changed the wording

icemac:
Paste at least does not have many dependencies, only `six`, see
https://github.com/cdent/paste/blob/6d7526982e00fc49634503123fbb8f7a95a678a4/setup.py#L56

dataflake:
@icemac could you review my changes please?

CHANGES.rst
```diff
@@ -55,6 +55,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Improve documentation for Zope's error logging services.
 
+- Add ``Paste`` as ``extras_require`` dependency to pull in ``Paste`` when 
+  installing with `pip` and `constraints.txt` to prevent startup errors.
+  (`#734 <https://github.com/zopefoundation/Zope/issues/734>`_)
```
icemac: Maybe we should mention here that this requires a change in how to install Zope with `pip`

dataflake: Just added a note in there.

AngelPujaa:
can we use zope3 with Python3 or not?

PR #736, opened 2019-11-23
Fix broken deprecated import when ZServer is not installed

### LINKED ISSUE #714: deprecated/deferred import for webdav.interfaces.IFTPAccess
Originally reported at 
https://community.plone.org/t/zodbupdate-modulenotfounderror-no-module-named-webdav-interfaces/10480

This deprecated import https://github.com/zopefoundation/Zope/blob/master/src/OFS/interfaces.py#L1059 is a problem because the top level webdav module is moved to ZServer. The latest ZServer is not compatible with python 3.

The problem manifested for me when trying to use zodbupdate to migrate my ZODB from python 2 to python 3. The deprecated import here seems to preclude using plone.app.upgrade's alias_module function to simulate deprecated modules, and letting it resolve to ZServer/webdav is not an option because that package is not compatible with python 3. My solution was to just edit OFS/interfaces.py and remove that deprecated import.

dataflake:
Workaround to fix #714

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ https://github.com/zopefoundation/Zope/blob/4.0a6/CHANGES.rst
 4.1.3 (unreleased)
 ------------------
 
+- Fix broken deprecated import when ZServer is not installed
+  (`#714 <https://github.com/zopefoundation/Zope/issues/714>`_)
+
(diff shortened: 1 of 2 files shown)


dataflake:
The OP has tested the fix, so I am merging.

PR #737, opened 2019-11-24
fixed border bug of select element (chromium)

drfho:
The PR fixes an irritating rendering defect of the html select element (header menu) in latest Chromium browsers: the outline does not become invisible by border-width:0 but needs a border-radius of 1px.

zmi_cssbug

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.0a1 (unreleased)
 ------------------
 
+- The method ``unrestrictedTraverse`` raises an error when
+  the argument ``path`` is not something it can work with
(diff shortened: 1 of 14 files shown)


dataflake:
Screenshot

Is this appearance in Safari normal? This is not the result of this patch, it's been that way before, I've just always be wondering about it.

drfho:
the problem of the safari glow effect is descibed here:
https://stackoverflow.com/questions/935559/remove-safari-chrome-textinput-textarea-glow
Maybe 2 more styles wilkl help?
https://github.com/zopefoundation/Zope/pull/737/commits/4ba7b3a1013e06aefb81d69b5a565aa001b3efb6

dataflake:
It's still the same... but it's not that important. I'll release 4.1.3 next weekend, even if there is no fix for Safari I am going to integrate your changes into the `4.x` branch as well.

PR #740, opened 2019-11-29
close opened db during shutdown

mamico:
Close opened database connections during shutdown. 

That's important at least with Relstorage, where ZODB cache is saved persistently on filesystem within db closing, with a great improvement on startup.

Here where ZServer do something like this: https://github.com/zopefoundation/ZServer/blob/master/src/ZServer/Zope2/Startup/starter.py#L112

### DIFF
--- a/src/Zope2/Startup/serve.py
+++ b/src/Zope2/Startup/serve.py
@@ -18,6 +18,7 @@ import optparse
 import os
 import re
 import sys
+import Zope2
 from logging.config import fileConfig
 
 from paste.deploy import loadapp
@@ -209,6 +210,8 @@ and then use %(http_port)s in your config files.
                     msg = ''
                 self.out('Exiting%s (-v to see traceback)' % msg)
             finally:
+                for db in Zope2.opened:
+                    db.close()
                 self.unlinkPidFile()
 
         serve()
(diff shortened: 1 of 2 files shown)


src/Zope2/Startup/serve.py
```diff
@@ -18,6 +18,7 @@
 import os
 import re
 import sys
+import Zope2
```
dataflake: This import needs to go below `from App.config import getConfiguration` to make the linting tests pass. I would have done it myself but since you are working off a fork I can't.

mamico: Sorry. Done. 

But the PR has checked the option "allow edit from mainteners" (that's the default) where I read "If checked, users with write access to zopefoundation/Zope can add new commits to your master branch."
Anyway, if is a better choice, next time I'll do a branch instead of working on a fork in my account. No problem.

dataflake: In didn't even know there is such an option. I never work with forks, so my automatic reaction is "this sits in someone elses's account, I have no access". But even with that option, I would be forced to do a separate clone and then build to get your fork. That's really not necessary. If possible, create a branch in here. You will make everyone's life easier.

PR #749, opened 2019-12-23
Update main.zpt to correct Last Modified sorting after paste

kenara:
add missing manage_main after paste into folder

### DIFF
--- a/src/OFS/zpt/main.zpt
+++ b/src/OFS/zpt/main.zpt
@@ -57,9 +57,9 @@
             </th>
             <th scope="col" class="zmi-object-date text-right hidden-xs">
               <a title="Sort Ascending by Modification Date"
-                 href="?skey=_p_mtime&rkey=asc"
+                 href="manage_main?skey=_p_mtime&rkey=asc"
                  tal:attributes="title python:'Sort %s by Modification Date'%( rkey_alt.upper() );
-                                 href python:'?skey=_p_mtime&rkey=%s'%( rkey_alt );
+                                 href python:'manage_main?skey=_p_mtime&rkey=%s'%( rkey_alt );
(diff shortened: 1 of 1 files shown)


jugmac00:
Hi Kenara, thank you for the reported bug and the pull request.

A couple of things first:
In order to be able to contribute to Zope you need to sign a developer agreement:
https://www.zope.org/developer/becoming-a-committer.html

I saw you referenced Zope 4.1.2 - usually it's best to test against the most up-to-date version, which is Zope 4.1.3 at the moment (or if you provide a patch yourself, test against master).

Ok, I can reproduce your bug.

But not only `sort by last modified` does not work, but all the other `sorter links` also do not work properly.

eg `sort by file size`
https://github.com/zopefoundation/Zope/blob/d0b62c5528ea2be5f9ab1cff3efbf834cab70454/src/OFS/zpt/main.zpt#L48-L51

Also, I am not entirely sure whether fixing the problem in the template is the correct way to go. Maybe the fix should go into `manage_pasteObjects` which is called when you hit the `paste` button (or even the cut/copy methods).

I have not much experience with ZMI and also there is this thing called Christmas :-), so I have no time to dig deeper at the moment.

Maybe @drfho or @dataflake, who have a lot of experience with the ZMI, will have a look at this after Christmas time.

Thanks again for your report and your pull request.

dataflake:
This is still missing the contributor agreement in order to be merged:
https://www.zope.org/developer/becoming-a-committer.html

dataflake:
Since this PR is unacceptable due to the missing contributor agreement I have created my own fix for `master` in 655b0aed4dd1fc159e3b5fc10c882e5b96c77a1d and `4.x` in 07e047818d1998cc13e4fe39df83656ec5edecd6, so this can be closed.

PR #762, opened 2020-01-23
Pin more-itertools to less than 6.0.0. [2.13]

mauritsvanrees:
Version 6 and higher no longer work on Python 2.7, so you get a SyntaxError when running buildout.

(I am looking at Zope 2.13 because the Plone hotfix of this week should be merged in parts of Zope. But the Zope(2) package itself does not seem affected. Still, would be nice if the buildout can finish.)

### DIFF
--- a/version_ranges.cfg
+++ b/version_ranges.cfg
@@ -99,6 +99,7 @@ mechanize = < 0.3.dev
 python-gettext = < 1.3.dev
 zope.testbrowser = < 3.12.dev
 mr.developer = < 1.35.dev
+more-itertools = < 6.0.0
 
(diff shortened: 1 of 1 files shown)


jugmac00:
If you got a spare minute... I am just curious - where is the dependency on `more-itertools` coming from? I can't find it in Zope nor in the hotfix.

mauritsvanrees:
With some hacking in zc.buildout:

- it is a dependency of `zipp 2.0.1`,
- which is a dependency of `importlib-metadata 1.4.0`,
- which is a dependency of `tox 3.14.3`

PR #764, opened 2020-01-26
- harmonize the viewlet and browserpage allowed_interface declarations

### LINKED ISSUE #755: allowed_interface declarations ignore inheritance of methods from base classes.
https://github.com/zopefoundation/Zope/blob/b41806767e4dc98d17142f5597a19500b3a84915/src/Products/Five/viewlet/metaconfigure.py#L48

Interface.names() ignores the names of methods/attributes from base classes. This means that declaring an allowed_interface only declares *part* of the actual interface, breaking common Python assumptions and conventions around class inheritance, and in fact inconsistent with other parts of Zope itself. I will submit a patch shortly.

dataflake:
Fixes #755 

I never used the viewlet declaration, so I can't say much about what the patch really does. If someone with enough knowledge could take a look and comment (or merge) that would be great. I'll do the backport to 4.x at that point.


icemac:
I'll look into this PR on Alpine City Sprint starting in 14 days.

PR #769, opened 2020-01-28
Fix typo

jugmac00:


jugmac00:
That is no work - I just read the documentation as I need it!

Thanks to the ones who wrote and updated the documentation!

Issue #742, opened 2019-12-02
Future of ZServer


mtoepfl:
First of all, thank you for all the work you all did porting Zope to Python 3!

I red about the effort ripping out ZServer completely from Zope 5 #696 

Isn't that to early looking at the promising work @datakurre did in porting ZServer from Medusa to Twisted on Python 3.

We are porting our existing Zope applications over to ReactJS and Zope does a very good job in serving data over a GraphQL API implemented with python-graphene. With its object database and its event system it is possible to implement feature which are offered by "modern" application severs like https://hasura.io/ very easily. Also performance is really good!

The only thing we miss is the async stuff. We need a way to notify the client about changes on the server over websockets.

I visited the following talk @datakurre gave at Plone Conf 2019:
https://datakurre.github.io/ploneconf2019/zserver-reloaded-with-http2-and-websocket-support.pdf

I tried out the PubSub-socket over ZServer and it works like a charm in Twisted and Python 3.

My question is, which is the recommended way to go? Zope over wsgi? Will there be asgi support in near future? Is there a chance to keep async ZServer in Twisted in mainstream? Which way Plone people will go for async challenges?

I'm not sure if I remember right, but I think @datakurre mentioned it's not as easy to implement with asgi as with Twisted because of the different threading models. But I'm not very experienced in this area.

dataflake:
The decision to go with WSGI has been made a long time ago. It's not up for disposition. Please also remember that the support code inside Zope itself has a cost. It must be maintained and tested. I haven't seen anyone step up.

mtoepfl:
Thank you @dataflake for your fast answer and the new release you did. I'm sorry, I posted my issue inadvertently during writing it. I gave more information and added my concrete question by editing the issue.

dataflake:
I don't really know much about websockets so I can't help you on that.

datakurre:
@mtoepfl FWIW, while I haven't been able to try Zope 5, I assume that "my ZServer" is not too far from working on it, because the code already uses the WSGI ZPublisher maintained in Zope 5.

But WebDAV support is really gone in Zope 5 with dropped ZServer support. I've understood that it could be restored in Zope (without depending on ZServer) if someone would champion it, as discussed in https://github.com/zopefoundation/Zope/issues/649

I really don't know much about ASGI yet. About WSGI and WebSockets in general I had the understanding that any solution would be more or less tied to some specific WSGI server, so I could as well implement it into ZServer, which I understand better than the other servers.

[...]

PR #772, opened 2020-01-29
Restore WebDAV functionality

### LINKED ISSUE #744: Re-integrate WebDAV support
Related to #649 , #696 , #742 

When ZServer was moved out of Zope the webdav support module was removed as well, even though it didn't neeed to. Bringing it back will restore WebDAV functionality to Zope.

The removal was done in 24d566b475ec60a5f872d39cdb63cb730fdc0f82, PR #696

dataflake:
Fixes #744 

Reverts most of #696 as well as the the webdav part of 24d566b475ec60a5f872d39cdb63cb730fdc0f82

Related to #649, #742 and to 

This PR has been put together from several sources. The first was the move of the `webdav` package into ZServer (24d566b475ec60a5f872d39cdb63cb730fdc0f82) and then the removal of all remaining support code in #696. I have left out anything that relates to FTP only, as that is an entirely separate protocol. Some inspiration was taken from https://github.com/datakurre/ZServer/tree/py3 by @datakurre. 

With the `webdav` package back in Zope supporting the refreshed ZServer code by @datakurre   should become easier as well.


src/ZPublisher/BaseRequest.py
```diff
@@ -451,6 +457,19 @@ def traverse(self, path, response=None, validated_hook=None):
...
+                    # This is webdav support. The last object in the path
+                    # should not be acquired. Instead, a NullResource should
+                    # be given if it doesn't exist:
+                    if no_acquire_flag and \
+                       hasattr(object, 'aq_base') and \
+                       not hasattr(object, '__bobo_traverse__'):
+
+                        if (object.__parent__ is not
+                                aq_inner(object).__parent__):
+                            from webdav.NullResource import NullResource
```
icemac: Could this import be moved to the top?

src/webdav/PropertySheets.py
```diff
@@ -0,0 +1,156 @@
...
+from App.Common import rfc1123_date
+from OFS.interfaces import IWriteLock
+from OFS.PropertySheets import PropertySheet
+from OFS.PropertySheets import View
+from OFS.PropertySheets import Virtual
+from webdav.common import isDavCollection
+from webdav.common import urlbase
+
+
+def absattr(attr):
```
icemac: There is another identical implementation in `webdav/common.py`.

icemac:
Should we even backport this PR to Zope 4 to ease the transition to Zope 4 and 5?

datakurre:
@pbauer What do you think, if WebDAV without ZServer would be backported for Zope 4, would there be "final" Plone 5.x release for Python 2 with WebDAV on WSGI?

dataflake:
I have no problem backporting it to Zope 4, no. Once it's done. There's still work to be done, like making the default object creation better (who needs DTML Documents anymore?), I need to look at the locking mechanism and its representation in the ZMI, and of course some real life testing.

src/ZPublisher/BaseRequest.py
dataflake: It cannot due to circular imports. I tried.

pbauer:
I could also imaging re-adding webdav-support to a 5.2.x release if its not a breaking change. Semantically 5.3 release would be correct though but I fear that would raise expectation to have more features that we're probably not able to live up to.

src/webdav/PropertySheets.py
dataflake: You're right, I deleted this one.

dataflake:
The WebDAV reintegration PR is now ready for review. I will leave this open for a few days if anyone wants to suggest corrections or improvements.

If you go to https://github.com/zopefoundation/Zope/tree/webdav/src/webdav and scroll down you will also see testing notes from tests with the "litmus" WebDAV test tool, including tests against the first implementation in 2007. The current code is a bit better since I took the time to continue diagnosis on some of the issues @mcdonc found back then.

pbauer:
I'd love to test this with Plone but we first have to make the branch for Plone 6 to run with the current state of Zope 5. We might find the time to do that at the Alpine City Sprint in Innsbruck next week. But please don't wait for us when it is ready merge!

dataflake:
As soon as it's merged I will backport the whole thing to the `4.x` branch. 4.x needs a new release soon, anyway. I'll call that 4.2 then.

dataflake:
FYI, Zope 4.2 is now released and it contains the WebDAV support code.

zopyx:
How do you enable the WebDAV support in Zope 5.0a1? Setting webdav-source-port is enough or do you habe to make changes to the WSGI configuration?

dataflake:
@zopyx It would be sufficient to ask this question in one place and not place the identical question in three places. Please read the documentation for `webdav-source-port` at https://zope.readthedocs.io/en/latest/operation.html#zope-configuration-reference

PR #787, opened 2020-02-16
Enable WebDAV support without having to use ZServer

dataflake:
In the first round of WebDAV changes I pulled the general WebDAV support back into Zope itself. However, that still requires a user to run ZServer because only ZServer knows how to mark up the request correctly using its designated WebDAV source server.

This PR implements the same support for WSGI-based installations. Zope gains a configuration setting to designate a network port as WebDAV source port, which triggers changes to the `REQUEST` object similar to what ZServer did if a request is coming through the designated port.

Zope itself has no server capabilities anymore, so the WSGI server must be configured listen to that port (alongside the standard port used for web requests) for the processing to kick in. This is easily done using the `listen` directive for `waitress`, which can take multiple `host:port` pairs.

The request modifications take place ~~in the central VirtualHostMonster `__call__` method, which is called before the start of traversal and a sensible place to make (traversable-related) changes to the request object.~~ in the WSGI publisher.


dataflake:
After re-thinking it I have moved the hook further up the pipeline, to the WSGI publisher itself. That way it is available without having to instantiate a Virtual Host Monster and the implementation is cleaner.

src/ZPublisher/tests/test_WSGIPublisher.py
```diff
@@ -784,6 +784,48 @@ def test_set_REMOTE_USER_environ(self):
...
+        # WebDAV source port not configured
+        environ = self._makeEnviron(PATH_INFO='/test')
+        self.assertNotIn('WEBDAV_SOURCE_PORT', environ)
+        self.assertEqual(WSGIPublisher._WEBDAV_SOURCE_PORT, 0)
+        self._callFUT(environ, start_response, _publish)
+        self.assertNotIn('WEBDAV_SOURCE_PORT', environ)
+        self.assertEqual(environ['PATH_INFO'], '/test')
+
+        # Configuring the port
+        WSGIPublisher.set_webdav_source_port(9800)
```
icemac: As this statement changes a global variable there should be a try-finally-clause to reset it at the end of the test or in teardown.

src/Zope2/Startup/tests/test_starter.py
```diff
@@ -108,6 +108,23 @@ def testSetupConflictRetries(self):
...
+            instancehome <<INSTANCE_HOME>>""")
+        starter = self.get_starter(conf)
+        starter.setupPublisher()
+        self.assertEqual(WSGIPublisher._WEBDAV_SOURCE_PORT, 0)
+
+        conf = self.load_config_text("""
+            instancehome <<INSTANCE_HOME>>
+            webdav-source-port 9800""")
+        starter = self.get_starter(conf)
+        starter.setupPublisher()
```
icemac: Same here.

icemac:
> (Additionally the tests on Appveyor are broken. I have seen the same breaking when trying to use Python 3.9 locally. I hope I still have the local branch to make it a PR.)

See #791.

src/ZPublisher/tests/test_WSGIPublisher.py
dataflake: A added a cleanup step at the end of the test

src/Zope2/Startup/tests/test_starter.py
dataflake: A added a cleanup step at the end of the test

dataflake:
As mentioned in #791 I have isolated the Appveyor issue somewhat. It's certainly not a showstopper here, just a big annoyance.

My merge right now will also bring in my shorter Appveyor config, by the way.

PR #791, opened 2020-02-24
Use pip inside tox

icemac:
When trying to update to Python 3.9 I saw the same errors like in https://ci.appveyor.com/project/zopefoundation/zope/builds/30978656/job/1xumy2pa0um4cdqu:

```python
File "c:\projects\zope\src\OFS\tests\testHistory.py", line 13, in <module>
   from ZODB.FileStorage import FileStorage
File "c:\projects\zope\.tox\lint\eggs\zodb-5.5.1-py2.7.egg\ZODB\FileStorage\__init__.py", line 3, in <module>
   from ZODB.FileStorage.FileStorage import FileStorage, TransactionRecord
File "c:\projects\zope\.tox\lint\eggs\zodb-5.5.1-py2.7.egg\ZODB\FileStorage\FileStorage.py", line 29, in <module>
   from zc.lockfile import LockFile
ImportError: No module named lockfile
```

They seem to occur because `zc.buildout` is installed via `pip` and `zc.lockfile` via `buildout`.
I only saw this kind of errors when running the tests under `coverage`. Without coverage `zc` is a module having two values in `__path__` (one porting to the site-packages and one pointing to the zc.lockfile egg). With `coverage` it is a namespace-package object those `__path__` only points to site-packages, so `zc.lockfile` cannot be found.)

This PR tries to solve the problem by installing everything using `pip`. This requires some new tox environments because I did not find a way to run both the Zope tests and the tests of the dependencies which are run before together.

Any ideas or suggestions?


tox.ini
```diff
@@ -1,26 +1,72 @@
...
     coverage-report
 
 [testenv]
 coverage_run = {env:WIN_COV:coverage run}
+use_develop = True
 commands =
-    {envbindir}/buildout -c {toxinidir}/buildout.cfg buildout:directory={envdir} buildout:develop={toxinidir} install test alltests
-    {[testenv]coverage_run} {envdir}/bin/alltests {posargs:-vc}
-skip_install = true
+    {[testenv]coverage_run} -m zope.testrunner \
```
icemac: This will no longer work on appveyor on `master` it was used to disable coverage on windows.

icemac:
Hm, `coverage` does not seem to work in this setup as we get a coverage of 0.0 %.

dataflake:
I experimented a little more to first clean up the Appveyor config and then isolate the problem more. This is my version now: https://github.com/zopefoundation/Zope/blob/webdav_part_2/appveyor.yml. With no changes to `tox.ini` and `.travis.yml` all tests run though - just not Python 2.7.

The problem only exists on Python 2.7. It looked like it also happened for the `lint` environment, but that was only because it used Python 2.7 despite the configuration that seemed to specify Python 3.7. Using the same `TOXENV` name that we have - `lint-py36` - fixed that.

Looking at the Appveyor output the buildout appears to install `zc.lockfile` correctly.

icemac:
@dataflake Once the problems with Persistence are fixe for Python 3.9 we will have the same problems we saw on Appveyor in Python 3.9. But let's wait until we see the problems.

dataflake:
Another data point. To make sure there's no issue like 4.2 with an undiscovered bad version pick for Python 2.7 I tested several scenarios in a clean clone. I am now able to reproduce the issues, still only under Python 2.7, and only if the tests run under `tox`.

The tests run through without a problem if I build a clean Python 2.7 virtual environment in the toplevel folder, run `bin/buildout` and then `bin/alltests` directly.

Running `tox` in the clean Python 2.7 environment behaves the same as it does under Python 3 and the `py27` environments fail.

dataflake:
Right now it looks like the tests on 4.x, including Appveyor with Python 2.7, run fine. On master I see a different failure about page templates that don't compile:

`builtins.KeyError: 'Suite'`

However, the import errors appear to be gone. This may be solved.

dataflake:
@icemac can this be closed?

icemac:
@dataflake Thank you for the reminder: As we found a way around the problem I was trying to solve here, I am going to close this PR now.

Issue #799, opened 2020-03-05
WebDAV: AttributeError: 'RequestContainer' object has no attribute 'PrincipiaSearchSource'


zopyx:
Using cadaver as WebDAV client to "edit nonexistingfile" performs three requests

```
127.0.0.1 - admin [05/Mar/2020:12:08:22 +0200] "OPTIONS /foo/m HTTP/1.1" 200 0 "-" "cadaver/0.23.3 neon/0.30.2"
127.0.0.1 - admin [05/Mar/2020:12:08:22 +0200] "LOCK /foo/m HTTP/1.1" 201 448 "-" "cadaver/0.23.3 neon/0.30.2"
127.0.0.1 - admin [05/Mar/2020:12:08:22 +0200] "PROPFIND /foo/m HTTP/1.1" 207 297 "-" "cadaver/0.23.3 neon/0.30.2"
```

Error:

```
2020-03-05 12:08:22 ERROR [waitress:357][waitress] Exception while serving /foo/m
Traceback (most recent call last):
  File "/home/ajung/src/zope/lib/python3.7/site-packages/waitress/channel.py", line 349, in service
    task.service()
  File "/home/ajung/src/zope/lib/python3.7/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/home/ajung/src/zope/lib/python3.7/site-packages/waitress/task.py", line 439, in execute
    app_iter = self.channel.server.application(environ, start_response)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/ZPublisher/httpexceptions.py", line 30, in __call__
    return self.application(environ, start_response)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/paste/translogger.py", line 69, in __call__
    return self.application(environ, replacement_start_response)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/ZPublisher/WSGIPublisher.py", line 364, in publish_module
    response = _publish(request, new_mod_info)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/ZPublisher/WSGIPublisher.py", line 267, in publish
    bind=1)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/ZPublisher/mapply.py", line 85, in mapply
    return debug(object, args, context)
  File "/home/ajung/src/zope/lib/python3.7/site-packages/ZPublisher/WSGIPublisher.py", line 68, in call_object
    return obj(*args)
[...]

PR #802, opened 2020-03-06
Zope-801 Fixes issue where GET on a binary WebDAV resource returned

### LINKED ISSUE #801: WebDAV: GET on File objects with content-type application/octet-stream are not handled properly
- I uploaded a binary file through cadaver to Zope through WebDAV, internal content-type application/octem-stream
- a GET on this resource returns an empty result body.

The GET operation calls

https://github.com/zopefoundation/Zope/blob/master/src/webdav/Resource.py#L677-L680

The `PrincipaSearchSource()` implementation return an empty byte string for non text/* content:

https://github.com/zopefoundation/Zope/blob/master/src/OFS/Image.py#L471-L476

The `manage_davGET()` implementation is broken by calling `PrincipaSearchSource()` here.

zopyx:
Fixes #801 by using the `bytes()` result for a `GET` operation rather than `PrincipaSearchSource`.
Zope 5 passes the pyfilesystem2 and webdavfs test suite with 80 tests completely with this fix.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Bug fixes
 - Fixed encoding issue of `displayname` WebDAV property 
   (`#797 <https://github.com/zopefoundation/Zope/issues/797>`_)
 
+- Fixed issue where GET on a binary WebDAV resource returned improper
+  data (empty string instead of the binary data)
+  (`#797 <https://github.com/zopefoundation/Zope/issues/801>`_)
 
 Other changes
 +++++++++++++
--- a/src/webdav/Resource.py
+++ b/src/webdav/Resource.py
@@ -677,7 +677,12 @@ class Resource(Base, LockableItem):
     def manage_DAVget(self):
         """Gets the document source"""
         # The default implementation calls PrincipiaSearchSource
-        return self.PrincipiaSearchSource()
+        print('DAV')
+        import pdb; pdb.set_trace()
+        try:
+            return bytes(self)
+        except TypeError:
+            return self.PrincipiaSearchSource()
 
     @security.protected(webdav_access)
     def listDAVObjects(self):
--- a/src/webdav/Resource.py
+++ b/src/webdav/Resource.py
@@ -677,8 +677,6 @@ class Resource(Base, LockableItem):
     def manage_DAVget(self):
         """Gets the document source"""
         # The default implementation calls PrincipiaSearchSource
-        print('DAV')
-        import pdb; pdb.set_trace()
         try:
             return bytes(self)
         except TypeError:


jugmac00:
Any chance to get a test which fails before and now succeeds? Especially the WebDav part has comparably low test coverage.

zopyx:
> Any chance to get a test which fails before and now succeeds? Especially the WebDav part has comparably low test coverage.

I would prefer having a functional testing approach - either through Litmus or my own testsuite (the one of PyFilesystem/webdavfs). Litmus did find the issues I found.

dataflake:
Litmus is not a part of the normal test suite so I consider that insufficient. Besides, you didn't document any "before and after" test results. I fear this is being rushed in without making sure that the other simple use case like editing simple content items in cadaver works. That's why I'd like unit tests.

jensens:
This is the failing test in Plone (test worked in 2.13 and was not available in 4.0-4.1 so we skipped it, then feature was restored in 4.2+):
https://github.com/plone/plone.app.imaging/blob/master/src/plone/app/imaging/tests/test_new_scaling.py#L157-L167
I think it can be modified to work in Zope - create an OFS Image in the test and try to get it via WebDAV.

zopyx:
> This is the failing test in Plone (test worked in 2.13 and was not available in 4.0-4.1 so we skipped it, then feature was restored in 4.2+):
> https://github.com/plone/plone.app.imaging/blob/master/src/plone/app/imaging/tests/test_new_scaling.py#L157-L167
> I think it can be modified to work in Zope - create an OFS Image in the test and try to get it via WebDAV.

The WebDAV testsuite of Zope is broken by design and of little use here. Artificial tests...nothing beats functional blackbox testing here.

zopyx:
> Litmus is not a part of the normal test suite so I consider that insufficient. Besides, you didn't document any "before and after" test results. I fear this is being rushed in without making sure that the other simple use case like editing simple content items in cadaver works. That's why I'd like unit tests.

A simple test does not bring you an extra safety. Different WebDAV clients do different things for the same operation. Cadaver for example asks only for a subset of properties in PROPFIND operations while our test suite asks for all properties (exposing the bug in #799). 

Adding functional tests as parts of CI is the way to go here

dataflake:
A simple functional unit test would have shown the issue and proven the fix. This is not a matter of different clients behaving differently.

jensens:
While I agree with @zopyx that the tests are far from perfect for webdav, I am also pragmatic enough to see this is out of scope here.A simple test like the one I referenced should IMO be sufficient. I can try to add it later, because I want this one merged and backported to 4.3.

dataflake:
Closing this. c7b09a91f2653ad77fe1e7caf9981620ffafd748 has a fix and some actual tests, and it adds a more appropriate implementation for files and images.

dataflake:
... and it has a backport to 4.x in 06ee93caa3226d26e6f4b12681bc6b0269e10462

Issue #717, opened 2019-10-25
Products.PageTempates.engine.Program.cook: design weakness - "engine" mostly ignored


d-maurer:
The `cook` method only partially honours its parameter *engine*. This parameter typically defines which tales expression types are available and how they are to be evaluated. `cook` reduces this information to a boolean via `engine is getEngine()`and defines the available expression types based only on this boolean and ignoring any other information in *engine*.

This makes it impossible for applications to define additional tales expression types and for components (such as e.g. `Products.TrustedExecutables`) to provide alternative evaluations for expression types.

The `cook` above creates a `ChameleonPageTemplate`. Its tales expression evaluations are incompatible with those for Zope's standard tales implementation. Thus, an `engine` created for the latter cannot be used with a `chameleon` template. Apparently, the authors of `cook` therefore decided that it would be the best backward compatibility approach to mostly ignore *engine*. In the long run, the decision is likely not optimal as applications and components may want to define additional expression types or change their evaluation.

I suggest the following changes:
* Zope no longer implicitly switches from its standard template implementation to `chameleon` but does so only due to explicit template engine configuration.
* `cook` uses its *engine* parameter as provided; i.e. it expects that the caller has ensured that the expression evaluations there are compatible with `chameleon`.
* components (such as `Products.TrustedExecutables`) must check the template engine configuration value and ensure an appropriate *engine*; if they do not support the template engine, then they can generate a meaningful error. Applications may do so as well or require a specific template engine.

leorochael:
I mostly agree with your suggestions, except the one about not implicitly switching to chameleon.

The use of chameleon by default is a nice feature over older versions, and most templates tend to just work, so I'd rather have a config switch to bring back the old standard implementation.

dataflake:
As I said in #728 I can't see a valid use case for wanting to use the old engine, which is not supported by anyone anymore. At least I haven't seen anyone stepping up to keep it alive. The code hangs around because no one has disentangled and deleted it (see #104 ).

jugmac00:
@dataflake Thanks - I was not aware of that issue.

d-maurer:
Fixed by #807, #812, #810.

PR #805, opened 2020-03-14
Improve on issue 717

d-maurer:
This addresses #717 and improved on the current state (sufficiently to allow `Products.TrustedExecutables` to support `chameleon`).

Unfortunately, it is not a complete solution. The reason: the `chameleon` integration is a mess. `chameleon`  requires a specialized TALES implementation, incompatible with a standard `zope` TALES implementation. The integration point is `zope.pagetemplate` but the integration itself is controlled by `Zope`. This PR causes `Zope`'s use of page templates to be aware of a potential template engine switch and use a corresponding TALES implementation. However, there are packages (e.g. `z3c.form`) which directly use `zope.pagetemplate` templates; they are unaware of a template engine switch and provide an inappropriate TALES implementation when a switch to `chameleon` has happened. This PR hacks in this situation: if the provided TALES implementation does not come from `Zope` (and therefore should be template engine switch aware), it is ignored and replaced by a trusted `chameleon` compatible TALES implementation which retained the previous behaviour in this case.

Maybe, an adapter "TALES implementation --> `chameleon` compatible TALES implementation" should be used instead of the above outlined fixed logic? This would allow packages directly using `zope.pagetemplate` to register their own adapters to achieve `chameleon` compatibility.

Ideally, however, `chameleon` should be able to work with a standard TALES implementation and not require something special.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,15 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.3.1 (unreleased)
 ------------------
 
+- Improve *engine* handling of
+  ``Products.PageTempates.engine.Program.cook``
+  to support TALES extensions/adaptations (such as those by
+  ``Products.TrustedExecutables``).
+  However, *engine* is still not honoured for templates
+  used directly from ``zope.pagetemplate`` (rather than ``Zope``);
+  for them, the previous behaviour is unchanged.
+  (`#717 <https://github.com/zopefoundation/Zope/issues/717>`_).
+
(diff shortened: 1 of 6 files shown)


d-maurer:
I have removed my review requests. I want first explore a different approach: adapt any standard zope TALES implementation to a `chameleon` TALES implementation. This may lose some special `chameleon` optimizations (especially for "python" expressions) but otherwise drastically increase robustness and simplify the integration.

d-maurer:
I close this PR in favor of #807 which is simpler, more localized, conceptionally sounder.

PR #807, opened 2020-03-18
Consistently use "zope.tales" as TALES implementation

d-maurer:
Formerly,  `Zope` components (e.g. `Zope` itself, `Products.CMFCore`, `Products.DCWorkflow`, ...) typically used `zope.tales` as TALES implementation. Recently, `Zope` itself uses as second TALES implementation `chameleon.tales` -- for PageTemplates with `chameleon` template engine. The use of two incompatible TALES implementations made hacks necessary as the one described in #717 (expression engine mostly ignored) which breaks applications/components requiring TALES expression adaptations/extensions (such as `Products.TrustedExecutables`).

This PR replaces the use of `chameleon.tales` by `zope.tales` even for PageTemplates with the `chameleon` engine. `Zope` and its components/applications can again consistently use `zope.tales` as its sole TALES implementation.

Fixes #717 and #806.

PR #810, opened 2020-03-27
Forward port of #807 (Consistently use "zope.tales" as TALES implementation)

d-maurer:
Port from `4.x` to `master`


d-maurer:
The failing py3.9 build comes from the `setup.py` of `zope.hookable` importing `Feature` (no longer supported by `setuptools`).

dataflake:
In my local environment using the latest tarball from python.org (3.9.0a5) the buildout runs through without a problem when running `bin/tox`. There are a lot of failing tests, a lot of the page template-related. But the output looks like this is some deeper issue, something that affects slice operations.

Have you tried running the tests locally with `tox`?

d-maurer:
> In my local environment using the latest tarball from python.org (3.9.0a5) the buildout runs through without a problem when running `bin/tox`. There are a lot of failing tests, a lot of them page template-related. But the output looks like this is some deeper issue, something that affects slice operations.

A `chameleon` related comment suggests important 3.9 changes regarding the `ast` library module (specifically: the comment says that the 3.9 `ast` has lost the `PARAM` node type). `ast` is fundamental for `chameleon` and `RestrictedPython` (and used superficially in `Products.PageTemplates.engine`).

> Have you tried running the tests locally with `tox`?

No. I have local installations of Python 2.7 and Python 3.6 (and now Python 3.9a5) and I run tests there with `ztest` (a small wrapper around `zope.testrunner`). For the other Python versions, I rely on the "Travis" tests.

d-maurer:
> But the output looks like this is some deeper issue, something that affects slice operations.

I have analyzed the failing test `AccessControl.tests.testZopeGuards.TestActualPython.testPython`. The problem effectively comes from `RestrictedPython.transformer.RestrictingNodeTransfomer.transform_slice` which distinguishes its operation based on the slice node type. The problem comes from 3.9 now using a different node types as slice than before - one, that `transform_slice` does not yet know how to handle.

In 3.6:
```
>>> from ast import parse, dump
>>> dump(parse("e[y]"))
"Module(body=[Expr(value=Subscript(value=Name(id='e', ctx=Load()), slice=Index(value=Name(id='y', ctx=Load())), ctx=Load()))])"
```
In Python 3.9:
```
>>> from ast import parse, dump
>>> dump(parse("e[y]"))
"Module(body=[Expr(value=Subscript(value=Name(id='e', ctx=Load()), slice=Name(id='y', ctx=Load()), ctx=Load()))], type_ignores=[])"
```
I.e. while formerly, `Index` was used as slice type, now it is `Name`.

d-maurer:
The `AccessControl` tests pass again with the following change to `RestrictedPython.transformer`:
```
--- transformer.py~	2020-03-29 11:05:42.451979637 +0200
+++ transformer.py	2020-03-29 12:02:38.407181025 +0200
@@ -355,6 +355,9 @@
         if isinstance(slice_, ast.Index):
             return slice_.value
 
+        elif isinstance(slice_, ast.Name):
+            return slice_
+
         elif isinstance(slice_, ast.Slice):
             # Create a python slice object.
             args = []
```

dataflake:
Thank you for the analysis, I have created zopefoundation/RestrictedPython#193 and copied your comments.

I have changed the Travis configuration to allow test failures on Python 3.9. This will stay in place until there are no more Python 3.9 issues due to important dependencies.

Other than that this looks like it is ready to go in.

d-maurer:
Jens Vagelpohl wrote at 2020-3-29 03:11 -0700:
>Thank you for the analysis, I have created zopefoundation/RestrictedPython#193 and copied your comments.
>
>I have changed the Travis configuration to allow test failures on Python 3.9. This will stay in place until there are no more Python 3.9 issues due to important dependencies.
>
> Other than that this looks like it is ready to go in.

Maybe, we wait for Maurits to check #812 against Jenkins
(i.e. that the `Archetypes` problems are fixed)
and then put in the changes from #812 as well.

PR #814, opened 2020-04-03
Use checkout of RestrictedPython.

icemac:
This hopefully fixed Python 3.9.


icemac:
Python 3.9 still fails in some template tests. I cannot run the tests on Python 3.9 locally even though I have 3.9.0a5 installed: When calling `tox -epy39` I get lots of errors that the Zope distribution cannot be found. I end up with over 100 test failures instead of the 6 ones TravisCI finds.

Does anyone have an idea to fix this locally without enforcing #791?

dataflake:
I see no such issues in my local setup. `tox -epy39` ran as expected and produced the same failures you see on Travis.

icemac:
I found a way to fix my local issue: Run `bin/tox -epy39` after installing `virtualenv` in the local `venv`. Now I get the same errors as TravisCI. I'll try to look into this issue later.

dataflake:
Just did a cross-check with a fresh clone and everything is fine, apart from those six failures. This is how I create my sandbox:

```
git clone ...
cd Zope
python3.7 -m venv .
bin/pip install -U pip setuptools zc.buildout virtualenv wheel
bin/buildout
bin/tox
```

icemac:
@dataflake My setup is quite close to yours. Thank you.

icemac:
I'd like to merge this PR so the tests on Python 3.9 can get green. (With this PR they fail with `AssertionError: Template errors: ['Compilation failed', 'builtins.SyntaxError: invalid syntax` in a single test.)

As Zope 5 final ist to be released before the final Python 3.9 release (September vs. October 2020), I see no need to have a release of `RestrictedPython` supporting Python 3.9 before the Zope 5 release.

PR #830, opened 2020-05-16
Improve solidity of the ``debugError`` method

### LINKED ISSUE #829: The debugError function is too brittle
The actual code
https://github.com/zopefoundation/Zope/blob/5bfcac7fc0c5a620326955571fc69f20e106a884/src/ZPublisher/HTTPResponse.py#L780-L784
breaks when `entry` is not a string.

For example I add an error when `entry`turned out to be a tuple:
```
(Pdb) l
823         # If a resource is forbidden, why reveal that it exists?
824         forbiddenError = notFoundError
825  
826         def debugError(self, entry):
827             import pdb;pdb.set_trace()
828  ->         raise NotFound(self._error_html(
829                 "Debugging Notice",
830                 ("Zope has encountered a problem publishing your object.<p>"
831                  "\n" + entry + "</p>")))
832  
833         def badRequestError(self, name):
(Pdb) entry
('The object at http://localhost:8080/Plone/... is not publishable.',)
```

ale-rt:
Fixes #829

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.0a3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Improve solidity of the ``debugError`` method.
+  (`#829 <https://github.com/zopefoundation/Zope/issues/829>`_)
 
 
 5.0a2 (2020-04-24)
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -780,8 +780,11 @@ class HTTPResponse(HTTPBaseResponse):
     def debugError(self, entry):
         raise NotFound(self._error_html(
             "Debugging Notice",
-            ("Zope has encountered a problem publishing your object.<p>"
-             "\n" + entry + "</p>")))
+            (
+                "Zope has encountered a problem publishing your object. "
+                "<p>%r</p>" % entry
+            )
+        ))
 
     def badRequestError(self, name):
         self.setStatus(400)
--- a/src/ZPublisher/tests/testHTTPResponse.py
+++ b/src/ZPublisher/tests/testHTTPResponse.py
@@ -873,7 +873,7 @@ class HTTPResponseTests(unittest.TestCase):
         except NotFound as raised:
             self.assertEqual(response.status, 200)
             self.assertTrue("Zope has encountered a problem publishing "
-                            "your object.<p>\ntesting</p>" in str(raised))
+                            "your object. <p>'testing'</p>" in str(raised))
         else:
             self.fail("Didn't raise NotFound")
 


dataflake:
@jugmac00 Did you actually look at https://github.com/zopefoundation/Zope/issues/829 and the changes here? The change was made specifically so the code can also handle the tuple you talk about. And the test code was changed.

jugmac00:
Maybe I miss something, but the test was run with a string input ("testing") before and afterwards ( or in other words - the test cases / fixtures have not changed).

https://github.com/zopefoundation/Zope/blob/e952da13dd9562225f3328c833c67afd0450ff66/src/ZPublisher/tests/testHTTPResponse.py#L869-L878

There was no update for the test for the case when debugError is called with a tuple, as it is done in production.

So, now it is possible, that somebody changes the production code, breaks the "contract" that tuple should be handled, and the test still passes.

dataflake:
The test fixture has changed because the code was changed to omit a newline character, and `repr` for a string will add the quotes. Did you overlook the part below?

```
@@ -873,7 +873,7 @@ def test_debugError(self):
        except NotFound as raised:
            self.assertEqual(response.status, 200)
            self.assertTrue("Zope has encountered a problem publishing "
-                            "your object.<p>\ntesting</p>" in str(raised))
+                            "your object. <p>'testing'</p>" in str(raised))
        else:
            self.fail("Didn't raise NotFound")
```

dataflake:
P.S.: There doesn't have to be a specific additional test case that uses a tuple, by the way. `%r` (`__repr__`) will always give you a string representation. The previous implementation using `+` would break on anything that is not a string.

jugmac00:
When I want a method to support str and tuple, I usually write at least tests for both cases (or with pytest one test with parametrization) and some with invalid input.

Once again, @ale-rt 's reason to change the production code was to support tuples as inputs. But the test only uses a string input.

Somebody can change the production code as follows - which is totally made up, I know:
```
    def debugError(self, entry):
        if isinstance(entry, tuple):
            return "The current test cases do not care whether tuples are supported."
        raise NotFound(self._error_html(
            "Debugging Notice",
            (
                "Zope has encountered a problem publishing your object. "
                "<p>%r</p>" % entry
            )
        ))
```

And all tests still pass! So, the test gives zero security and regressions can happen.

That all said - you were faster and you have a different opinion and I am totally fine with that.

Bear with me - I am the only Python developer in my company and I have very strong opinions and nobody is there to object :-D

Cheers!

dataflake:
If you wanted to add a test that uses a tuple then all that test would show is that ‘repr’ on a tuple returns a string. I think we can agree that we do not need to write tests that test basic built-in functionality.

ale-rt:
Sorry for being late to the party :)

I thought @jugmac00 has a point and so I wrote some tests https://github.com/zopefoundation/Zope/pull/832.
Turns out I discovered that the patch is flawed!

I overlooked the fact that `"%r" % obj` is different from `"%s" % repr(obj)`, e.g.:
```
>>> "%r" % (1)
'1'
>>> "%r" % (1, 2)
TypeError: not all arguments converted during string formatting
>>> "%s" % repr((1, ))
'(1,)'
>>> "%s" % repr((1, 2))
'(1, 2)'
```
So my patched worked by chance with a 1 dimensional tuple :)

Also using format works in a more expected way:
```
>>> "{0!r}".format((1,))
'(1,)'
>>> "{0!r}".format((1, 2))
'(1, 2)'
```

I am keen to change the patch like:
```
"Zope has encountered a problem publishing your object. <p>{0!r}</p>".format(entry)
```

What do you think?

P.s.: waiting for this to be the norm:
```
>>> entry = 1
>>> f"{entry}"
'1'
>>> entry = (1,)
>>> f"{entry}"
'(1,)'
>>> entry = (1,2)
>>> f"{entry}"
'(1, 2)'
```

jugmac00:
> "Zope has encountered a problem publishing your object. <p>{0!r}</p>".format(entry)

Actually, I think this is hard to read.

There is an easier way:
https://stackoverflow.com/questions/1455602/printing-tuple-with-string-formatting-in-python

d-maurer:
Jürgen Gmach wrote at 2020-5-16 10:10 -0700:
>> "Zope has encountered a problem publishing your object. <p>{0!r}</p>".format(entry)
>
>Actually, I think this is hard to read.

Why?

The only not very often used feature in the above expression is the
"!r" conversion which means "convert via `repr`".


Of course, "<p>{0!r}</p>".format(entry)" is
equivalent to "<p>%s</p>" % repr(entry)". But "%" formatting has
some caveats (e.g. special treatment of tuples) which apprently
lead to the move to the `format` nethod.
Once you have understood the features of its pattern language,
its use is quite natural.

jugmac00:
I just think that the solution from Alex which I linked above, is easier to read
```
>>> thetuple = (1, 2, 3)
>>> print "this is a tuple: %s" % (thetuple,)
this is a tuple: (1, 2, 3)
```

But anyway, I have no strong preferences.

When I recall the discussion during this week's Zope sprint correctly, we possibly drop Python 3.5 support in October and then we can use f-strings anyway. At least for Zope 5.

PR #832, opened 2020-05-16
829 fix debug error take 2

ale-rt:
Contains a proof that the patch merged in #830 works with one element tuples but not with two element tuples


ale-rt:
This was just a test

PR #839, opened 2020-05-28
Improve information for errors during template execution

### LINKED ISSUE #837: PageTemplates: suboptimal exception information with chameleon engine
When a `chameleon` template hits an exception, it wraps the exception into a `chameleon.exc.ExceptionFormatter` which is responsible for the presentation. This uses `chameleon.utils.format_kwargs` to obtain a representation of the template's context.

`format_kwargs` does not simply call `repr` on the values but uses type specific presentation logic. The fallback case represents a value *v* as `<"%s %s at %s" % (type(v).__name__, getattr(v, '__name__`, "-"), hex(abs(id(v))))`. This often gives in a Zope context `<ImplicitAcquisitionWrapper  at 0x....>` which is not very informative.

The basic problem is in `chameleon`. Currently, it does not provide a way to register additional representations for types with special importance for a context, such as `ImplicitAcquisitionWrapper` for Zope.

d-maurer:
This PR uses (upcoming) `chameleon>=3.7.2` configuration to improve information for errors during template execution. It fixes #837.

The PR essentially restores corresponding behavior known from `zope.pagetemplate` (actually `zope.tales.tales.TALESTracebackSupplement`). Malthe reports in "https://github.com/malthe/chameleon/issues/309#issuecomment-634469772" that early `chameleon` versions behaved similarly but in some cases, the information was found to be too detailed, such that now the `chameleon` default drastically limits the amount. The PR configures `chameleon` in a way to go back to its previous behavior. I think this is acceptable as we lived for decades with it without complaint (to my knowledge).

The code change becomes effective only with `chameleon>=3.7.2` (not yet released). With an older `chameleon` version, the behavior remains unchanged. For this reason, the PR currently does not have tests verifying the new behavior.

The PR also fixes an unrelated spelling error (`RepeatDict` -> `RepeatDictWrapper`) which apparently had no significant effect.


setup.py
```diff
@@ -73,7 +73,7 @@ def _read_file(filename):
         'AccessControl >= 4.2',
         'Acquisition',
         'BTrees',
-        'Chameleon >= 3.7.0',
+        'Chameleon >= 3.7.2',
```
dataflake, line 76: This is a minor concern, but I would only update the pin in setup.py if it prevents real errors because pins in setup.py are a big blunt hammer. This PR "only"  solves a convenience issue.

d-maurer: I have reset `setup.py` to its former state. Depending on how `Zope` is installed, the added test may fail (if `chameleon < 3.7.2` is used).

d-maurer:
@dataflake Would you like to take another look?
If I hear no objection, I will merge tomorrow morning.

dataflake:
Totally fine for merging right now from my point of view.

PR #840, opened 2020-05-28
Roadmap

icemac:
As discussed on the Zope May Sprint I sketched a rough roadmap for Zope.

What du you think about it?


icemac:
On my notes I additionally had:

* drop support for Python 3.5 in Zope 5 (as its sunset is in September 2020) instead use f-strings
* maybe Python 3.9 support has to be delayed until Zope 5.1

dataflake:
> On my notes I additionally had:
> 
>     * drop support for Python 3.5 in Zope 5 (as its sunset is in September 2020) instead use f-strings
> 
>     * maybe Python 3.9 support has to be delayed until Zope 5.1

IMHO Zope 5 is a good place for a "clean slate" as far as supported Python versions is concerned. I wouldn't mind dropping Python 3.5 unless there's some provable critical demand for it.

Do you think Python 3.9 is too hard to support in time for 5.0?

docs/roadmap.rst
```diff
@@ -0,0 +1,35 @@
...
+* It supports Python 2.7, 3.5 up to 3.8.
+* It is currently actively developed.
+* After the final release of Zope 5 it will be moved into bugfix mode until the end of 2021.
+* Until the end of 2022 it will get security fixes.
+  (This is when Plone plans to end the security support for Plone releases based on Zope 4.)
+  
+Zope 5
+------
+
+* This the current development version which already has some alpha releases.
```
jugmac00, line 28: This **is** the ...

icemac:
@dataflake wrote:
> IMHO Zope 5 is a good place for a "clean slate" as far as supported Python versions is concerned. I wouldn't mind dropping Python 3.5 unless there's some provable critical demand for it.

Agreed, I am putting it on the list for Zope 5.

> Do you think Python 3.9 is too hard to support in time for 5.0?

I am not sure. It is hard to say before the final release is out. Previous versions even had changes we had to incorporate within the beta versions. I'd rather not depend on Python 3.9 support for Zope 5.0 final. (If we get it in time – no problem.)

icemac:
See #841 as the issue to remove Python 3.5 support.

dataflake:
I condensed the document to make it easier to spot the important bits and will merge it now so we have it up in public on RTD. That way more people will see and maybe comment. We can always go back and adust.

PR #843, opened 2020-06-04
Update the management bar with the current version

rbanffy:
It looks more appropriate. Maybe we could read the version on rendering, but it's not that terrible to edit this on every major release.

I'm just happy my favorite application server aged so well. 
 
image

### DIFF
--- a/src/App/dtml/manage_navbar.dtml
+++ b/src/App/dtml/manage_navbar.dtml
@@ -11,7 +11,7 @@
   </div>
   <a class="navbar-brand" title="To Zope Root"
      href="<dtml-var "REQUEST.getVirtualRoot()">/manage" target="_parent">
-    <span class="product">ZOPE 4</span>
+    <span class="product">ZOPE 5</span>
   </a>
   <ul class="navbar-nav flex-row ml-sm-auto d-flex">
 


dataflake:
How about the actual full version number instead?

rbanffy:
I think the full version number would be too much information for a management interface. Would be a good idea for development branches, perhaps. 

I also like that this patch is a single least-significant bit flip on a single character. It's the minimum possible commit that actually improves a product.

dataflake:
I have created https://github.com/zopefoundation/Zope/pull/844 which does the same thing without resorting to a hardcoded value.

rbanffy:
Awww... That's too bad - yours is much better than mine. :-(

I guess I won't break my record for most insignificant open source contribution (currently another bit-flip, at http://bazaar.launchpad.net/~maas-committers/maas/trunk/revision/3825).

dataflake:
Don't feel bad - the idea was great and I uncovered a flaw in version parsing in Zope itself that I fixed at the same time.

rbanffy:
Easy for you to say ;-)

Well... Let's close this. Your's is obviously a superior solution, even though mine was much more concise.

PR #847, opened 2020-06-05
Fix ``default`` keyword handling in page templates

### LINKED ISSUE #846: PageTemplate handling of `default` broken starting with release 4.4
Symptom: Attempting to use the `default` keyword in a TAL expression to render the default contents of a tag will not render that content, but something like `<object object at 0x105810e30>`. This happens regardless of how `default` is used:

```
<p tal:define="foo string:" tal:content="python: foo or default">Defaul content</p>
```
 or

```
<p tal:content="context/unknownvariable|default">Defaul content</p>
```

Tagging Dieter because 4.4 contained major PageTemplate-related changes from him.

dataflake:
Fixes #846


dataflake:
@d-maurer This PR adds a simple test that passes on Zope < 4.3.1.

As I am trying to debug this I see very strange behavior: With the `CHAMELEON_CACHE` environment variable set to a valid folder path the test passes. As soon as I unset `CHAMELEON_CACHE` the test fails again.

dataflake:
Next step: When enabling `CHAMELEON_DEBUG` or `CHAMELEON_CACHE` the Python code chameleon generates is definitely different.

Without caching or debugging enabled `chameleon.loader.MemoryLoader` is used. In the resulting Python file I see the following at the top:

```
from Products.PageTemplates.engine import _compile_zt_expr as __compile_zt_expr
from Products.PageTemplates.engine import _c_context_2_z_context as __c_context_2_z_context
from sys import exc_info as _exc_info

_static_4489251152 = __c_context_2_z_context
_static_4489251280 = __compile_zt_expr
```

With caching or debugging, `chameleon.loader.ModuleLoader` is used. It apparently passes in entirely different contexts:

```
from Products.PageTemplates.expression import BoboAwareZopeTraverse as _BoboAwareZopeTraverse
from sys import exc_info as _exc_info
from zope.location.interfaces import LocationError as _LocationError
from zExceptions.unauthorized import Unauthorized as _Unauthorized
from zExceptions import NotFound as _NotFound

_static_4530880656 = _BoboAwareZopeTraverse()
```

d-maurer:
Jens Vagelpohl wrote at 2020-6-5 07:20 -0700:
>@dataflake requested your review on: zopefoundation/Zope#847 Fix ``default`` keyword handling in page templates.

I will look into the issue in detail tomorrow.

d-maurer:
Jens Vagelpohl wrote at 2020-6-5 08:17 -0700:
>Next step: When enabling `CHAMELEON_DEBUG` or `CHAMELEON_CACHE` the Python code chameleon generates is definitely different.
>
>Without caching or debugging enabled `chameleon.loader.MemoryLoader` is used. In the resulting Python file I see the following at the top:
>
>```
>from Products.PageTemplates.engine import _compile_zt_expr as __compile_zt_expr
>from Products.PageTemplates.engine import _c_context_2_z_context as __c_context_2_z_context
>from sys import exc_info as _exc_info
>
>_static_4489251152 = __c_context_2_z_context
>_static_4489251280 = __compile_zt_expr
>```
>
>With caching or debugging, `chameleon.loader.ModuleLoader` is used. It apparently passes in entirely different contexts:
>
>```
>from Products.PageTemplates.expression import BoboAwareZopeTraverse as _BoboAwareZopeTraverse
>from sys import exc_info as _exc_info
>from zope.location.interfaces import LocationError as _LocationError
>from zExceptions.unauthorized import Unauthorized as _Unauthorized
>from zExceptions import NotFound as _NotFound
>
>_static_4530880656 = _BoboAwareZopeTraverse()

Are you sure that this is code newly generated (and not old
cached code)? The code above looks like native `chameleon` code with
`chameleon.tales` as TALES engine -- generated before my
switch to `zope.tales`.
You could empty the cache folder to ensure you get newly generated
code.


`chameleon`'s caching code is fragile. It caches based on a hash
value. The hash value is built based on a hash of the template
source and a (small, apparently abritrary selected) sequence of builtin names.

To force cache invalidation between Zope versions,
I added a builtin containing the Zope version in its name.
I am not sure that this really ensures cache invalidation
between Zope versions (it is difficult to test:
without hacks, during a test, the Zope version remains constant).
Maybe, I made a mistake here and stale cached data can still be used?

dataflake:
I have deleted the contents of the cache folder for the run where I had the cache enabled. It should be fresh. I ran this several times.

d-maurer:
Jens Vagelpohl wrote at 2020-6-5 12:19 -0700:
>I have deleted the contents of the cache folder for the run where I had the cache enabled. It should be fresh. I ran this several times.

I will look tomorrow.

dataflake:
Your latest change fixes both the tests as well as a manual test with a page template in a running Zope instance.

d-maurer:
As explained in a comment to #846, the problem arises because each `chameleon` template has its own private representation of the TALES `default` object: a local constant `'__default'`.

Commit "91980b6cb" tries to work around this (conceptual) `chameleon` flaw by inspecting the constants of the `chameleon`'s `_render` function (it represents the template) for `default` representation candidates. If it finds a single one, it uses this object as representation of the TALES `default` value.

This approach assumes that Python maps all occurrencies of the string constant `'__default'` inside the `_render` function to a single object. While I do not know that Python guarantees this, `chameleon`, too, seems to rely on this assumption.

Maybe, @malthe has a better idea to resolve this issue.

dataflake:
While the immediate issue is solved by your patch, do you have any idea why cache/debug flags create different code representations than without cache/debug?

d-maurer:
Jens Vagelpohl wrote at 2020-6-6 00:47 -0700:
>While the immediate issue is solved by your patch, do you have any idea why cache/debug flags create different code representations than without cache/debug?

I can imagine that debug flags might cause different code to be generated
(depending on the purpose of the debug flags).

On the other hand there should be no code differences
in the caching versus non-caching case - when the cache does not
contain the corresponding item. Stepping through `chameleon` code,
I have not yet hit indications which indicate such differences:
code generation and caching seem to be well separated.

malthe:
With constant folding, do you mean [string interning](https://docs.python.org/2/library/functions.html#intern) – ?

d-maurer:
Malthe Borch wrote at 2020-6-6 02:09 -0700:
>With constant folding, do you mean [string interning](https://docs.python.org/2/library/functions.html#intern) – ?


No.

With "constant folding" I meant that several occurrencies of
a constant are mapped to the same object (such that one can compare
them with "is" rather than "==").
(Usually, "constant folding" means more: that (some) constant expressions
are propagated and evaluated at compile (rather than run) time.)


When I looked at the `chameleon` generated code (in the `_compile` method),
I had the impression that (e.g.) a `tal:content=<expr>` is compiled to
something like:

	  expression = ... evaluate <expr>
	  if '__default' is expression:
	    ...
          else:
	    ...
	  
This assumes that the occurrence of '__default' in the 'is' expression
refers to the same object as one (potentially) used during the evaluation
of "expr".

Python specifies that if such constants are immutable, then those
occurrencies may or may not refer to the same object. It declares
that all occurrencies of `None`, `True` and `False` always
refer to the same object. It explains
further that CPython (currently) does the mapping for (some) small
integers.

Apparently, many Python variants map occurrencies of
the same string constant also in the same function body (this is necessary
for the above use of "is" to be safe). I believe however, that
Python does not guarantee this.


In my view, the TALES `default` object should be a singleton, unique
across all templates (at least of the same kind). This would
allow a template "A" to call (i.e. "render") template "B" passing
`default` as an argument: `default` could be (almost) handled
like any other value.

`zope.pagetemplate` passes `default` to templates via the "context"
(and the typical "context" is set up to map `default` to
`zope.tales.tales._default`). I do not know what happens if
a template tries to override `default` (of course, it should not do this).

malthe:
@d-maurer the reason I mention string interning is that you can get a string that you can reliably do object identity testing with cheaply using `intern(<string>)`.

It might be that all that's really required is that those `__default` strings need to be interned as part of module initialization – i.e. that it must be compiled as `intern("__default")` or whatever it is.

d-maurer:
Malthe Borch wrote at 2020-6-6 23:33 -0700:
>@d-maurer the reason I mention string interning is that you can get a string that you can reliably do object identity testing with cheaply using `intern(<string>)`.

>


I do not say that an `intern` based solution would be impossible.
It is just not the one I favour.

My favorite solution is to get a global object
defined (globally) by `chameleon` (in any way it likes)
which can be used by any TALES implementation
interacting with the `chameleon` template engine to get
the semantics of the TALES `default` object.

The integration could check for the existence of this object
to determine whether the `chameleon` version is sufficiently recent
or whether the workaround is still required --
thus, no need for tight version synchronization.

d-maurer:
Jens Vagelpohl wrote at 2020-6-6 00:47 -0700:
>While the immediate issue is solved by your patch, do you have any idea why cache/debug flags create different code representations than without cache/debug?

An idea which might explain differences between template use
in Python and Zope:
The `chameleon` integration is still controlled via a utility.
When you start Zope, this utility is registered (and the
`chameleon` template engine is used); when you use a `PageTemplate`
directly in Python, the default `zope.pagetemplate` engine is
used (unless you have registered the utility explicitly).

dataflake:
The test that shows the difference is part of this PR. As you can see in `src/Products/PageTemplates/tests/testHTMLTests.py` the `HTMLTests` class explicitly calls `useChameleonEngine` in its `setUp` method. So the registration should be unaffected if all I do is set or unset environment variables that chameleon reads between different test runs.

dataflake:
I have merged Dieter's workaround because we need some fix and there's no indication if @malthe wants to take this up on the Chameleon side.

malthe:
@dataflake FYI there is a fix out https://github.com/malthe/chameleon/pull/313.

dataflake:
@malthe That patch does not solve the issue. I still get stuff like `&lt;object object at 0x10afaa820&gt;` rendered instead of the actual default content.

d-maurer:
Malthe Borch wrote at 2020-6-17 07:25 -0700:
>@dataflake FYI there is a fix out https://github.com/malthe/chameleon/pull/313.

As discussed, this requires that the Zope integration must
either use the implementation detail "default implemented by
`intern('__default')`) or look into the generated template "render"
function (as does the current hack). Both approaches are not
optimal.

Why not putting "default" (however you choose to implement it) into
the "Scope" (aka `econtext`) at the start of template rendering?
TALES will use this `default` - and when the template puts it there
itself, no need for the integration to know how it is implemented.

d-maurer:
Jens Vagelpohl wrote at 2020-6-17 07:44 -0700:
>@malthe That patch does not solve the issue. I still get stuff like `&lt;object object at 0x10afaa820&gt;` rendered instead of the actual default content.

The integration would need an adaptation to the
changed `chameleon` behavior.

Currently, it overrides the definition of `__default` in
the prolog to `from zope.tales.tales import _default as __default`
(making @malthe change ineffective).

An adaptation would keep the prolog unchanged
and put the definition "default = intern('__default')" into the
rendering context.


This could be done but would use the implementation
detail that `chameleon` implements the TALES `default`
via `intern('__default')` -- possible but not optimal.

dataflake:
So we're still at a standstill it seems. I read a lot of "could", "would" or "possible" but no better solution.

d-maurer:
Jens Vagelpohl wrote at 2020-6-17 16:15 +0000:
>So we're still at a standstill it seems. I read a lot of "could", "would" or "possible" but no better solution.

The primary decision is:

  How `chameleon` should make possible the coordination regarding
  the TALES `default` value between TALES and templates.

Once this decision is made and a corresponding `chameleon`
version available, the integration into Zope can adapt to
that decision. The change will likely be simple.


Likely, it is possible to set up "travis" to work with
a `chameleon` branch **BUT** I do not yet know how.
With my current knowledge, I would need a `chameleon` release.

Apart from this testing issue, I could adapt
the integration to "https://github.com/malthe/chameleon/pull/313"
within less than a hour.

malthe:
I have released Chameleon 3.7.3 with the above-mentioned fix.

d-maurer:
> I will release Chameleon 3.7.4 with the above-mentioned fix.

**NOTE** The current  4.x will **NOT** work with this change: it will require a prior `chameleon` version.

dataflake:
That does not matter since everything is pinned.

PR #848, opened 2020-06-08
Better version recognition with regex from packaging package

dataflake:


src/App/version_txt.py
```diff
@@ -23,6 +23,36 @@
 ZopeVersion = collections.namedtuple(
     "ZopeVersion",
     ["major", "minor", "micro", "status", "release"])
+VERSION_PATTERN = r"""
+    v?
```
gforcada, line 27: minor nitpick, as the merge request title suggest, this regex comes from another package, right? If so, I have the tendency to document where I copied it from and which version of that package I used, just in case it evolves and one wants to check what are the differences with the newer versions (in case it ends up growing independently here)

dataflake: I understand what you mean but there is no desire to "track" what's in the other package. This is just something that works. Only if it ever fails would there be a reason to revisit.

gforcada: Sure, no worries :slightly_smiling_face:

PR #850, opened 2020-06-16
4.x: New versions

icemac:
Update to new bugfix versions.


icemac:
Thank you.

dataflake:
Today I noticed that you only updated third-dot dependencies where I would usually update to the latest releases unless there are obvious incompatibilities, like missing Python 2 support on our `4.x` branch.

Your approach appears to assume that dependency package maintainers also have a strict delineation between second-dot and third-dot releases like we are trying, but I am not sure you can rely on that.

I am working off the premise that it's generally a better idea to use the latest compatible dependency release so we're not missing out on fixes or improvements.

Wht do you think?

dataflake:
By the way, after Zope 4.4.4 I have gone ahead and upgraded to the latest release that still supports Python 2, no matter if it was just a bugfix or a feature update. I looked at all of them specifically and noted that some had made a feature release for adding support for Python 3.8, so that's a release we definitely want to have seeing how we claim to support Python 3.8.

icemac:
@dataflake Thank you for looking into the dependencies. I only did as little as possible to get at least bug fixes. But you are right, there are some important feature releases we should use as Zope 4 is currently the stable Zope version. (I was looking to much into what was done for Zope 2.13)

icemac:
Maybe we need another approach to make sure we update to the newest versions which are still Python 2 compatible. `z3c.checkversions` has a blacklist feature but it only allows to list certain versions but not something like `< 4.0`.

dataflake:
I personally don't mind doing the (sometimes) manual vetting and leaving comments in the versions files to explain where we need to hold back upgrades. Just in case you think it's too much work.

icemac:
Alternatively maybe https://github.com/zopefoundation/z3c.checkversions/issues/17 could lead to a solution.

dataflake:
Again, I personally don't need a solution and won't spend time on it. If you think it's worthwhile, go ahead.

PR #857, opened 2020-06-30
Switch to use f-strings.

### LINKED ISSUE #841: Drop support for Python 3.5 and use f-strings
Python 3.5 will have its sunset in September 2020 (5 years after its initial final release, see https://www.python.org/dev/peps/pep-0478/).

According to #840 we plan to release Zope 5 in September 2020, there should be no need to support Python 3.5.

After dropping Python 3.5 support we are able to use f-strings in Zope.
Using [pyupgrade](https://pypi.org/project/pyupgrade/) most of the strings can be converted automatically.

Does anyone have a problem with this approach? Please raise your hands.

icemac:
I called `pyupgrade --py3-only --py3-plus --py36-plus --py37-plus` on
the Python files.
While looking through the changes, I did some adoptions to sometimes
use f-strings in places where the tool did not use them.

Additionally `pyupgrade` removed some Python 2 compatiblity leftovers.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Configuration file for the Sphinx documentation builder.
 #
--- a/docs/zdgbook/examples/PollImplementation.py
+++ b/docs/zdgbook/examples/PollImplementation.py
@@ -1,7 +1,7 @@
 from Poll import Poll
 
 
-class PollImplementation(object):
+class PollImplementation:
     """
     A multiple choice poll, implements the Poll interface.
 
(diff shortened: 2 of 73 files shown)


jugmac00:
@icemac 

> I called pyupgrade --py3-only --py3-plus --py36-plus --py37-plus on the Python files.

Does that mean Zope 5 does not support Python 3.6 any more?

dataflake:
No it doesn't. I am guessing that was either a typo or just done unintentionally. But either way, if you look at https://pypi.org/project/pyupgrade/ and search for any mention of `--py37-plus` then all you will find is something having to do with  one kind of `__future__` import.

icemac:
I've never used `generator_stop` which is removed by `--py37-plus` and it is not in the diff. But you are right, I should not have used this option in the first place.

PR #861, opened 2020-07-03
Restore "attrs" template variable support

### LINKED ISSUE #860: Products.Pagetemplates: several predefined variables lost
With the switch to `zope.tales` several predefined variables documented under "https://zope.readthedocs.io/en/latest/zopebook/AppendixC.html#built-in-names" are no longer supported: `attrs`, `root`, `container`, `CONTEXTS`.

This was originally reported in "https://github.com/plone/Products.CMFPlone/issues/3135".

d-maurer:
This fixes #860 as far as ``chameleon`` allows it.

As "https://github.com/malthe/chameleon/issues/323" demonstrates, `chameleon` currently supports `attrs` only in `tal:content` but not in `tal:define` nor in `tal:replace`. This PR allows the use of the predefined page template variable `attrs` at locations where `chameleon` supports it; if used at other places, `attrs` has the value `None`. The code should continue to work after the `chameleon` problem is fixed - but some cleanup would be possible after that.

The PR also removes the mentioning of `CONTEXTS` as predefined page template variable from the Zope Book as no template engine (neither `zope.pagetemplate`nor `chameleon`) implements it. It adds tests for the remaining predefined variables.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,12 +10,11 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
-- support the ``attrs`` predefined template variable again (as
+- Support the ``attrs`` predefined template variable again (as
   far as ``chameleon`` allows it)
   (`#860 <https://github.com/zopefoundation/Zope/issues/860>`_).
 
-- in the "Zope Book" remove ``CONTEXTS`` from the list of predefined
-  template variables (because it was never implemented).
+- Improve documentation of ``CONTEXTS`` in the "Zope Book".
   
 - Update dependencies to the latest releases that still support Python 2.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,13 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- support the ``attrs`` predefined template variable again (as
+  far as ``chameleon`` allows it)
+  (`#860 <https://github.com/zopefoundation/Zope/issues/860>`_).
+
+- in the "Zope Book" remove ``CONTENTS`` from the list of predefined
+  template variables (because it was never implemented).
+  
 - Update dependencies to the latest releases that still support Python 2.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
   far as ``chameleon`` allows it)
   (`#860 <https://github.com/zopefoundation/Zope/issues/860>`_).
 
-- in the "Zope Book" remove ``CONTENTS`` from the list of predefined
+- in the "Zope Book" remove ``CONTEXTS`` from the list of predefined
   template variables (because it was never implemented).
   
 - Update dependencies to the latest releases that still support Python 2.
--- a/docs/zopebook/AppendixC.rst
+++ b/docs/zopebook/AppendixC.rst
@@ -698,6 +698,13 @@ Note that the (popular) ``chameleon`` template engine implements ``attrs``
 and ``default`` not as standard variables but in a special way.
 Trying to change their value may have undefined effects.
 
+Note that beside variables you can use ``CONTEXTS``
+as initial element in a path expression. Its value is a mapping
+from predefined variable names to their value. This can be used to
+access the predefined variable when it is hidden by a user defined
+definition for its name. Again, `attrs` is special; it is not covered
+by `CONTEXTS`.
+
 
 TALES Exists expressions
 ========================
--- a/docs/zopebook/AppendixC.rst
+++ b/docs/zopebook/AppendixC.rst
@@ -674,10 +674,6 @@ These are the names always available to TALES expressions in Zope:
 - *attrs*- - a dictionary containing the initial values of the attributes of
   the current statement tag.
 
-- *CONTEXTS*- - the list of standard names (this list). This can be used to
-  access a built-in variable that has been hidden by a local or global variable
-  with the same name.
-
 - *root*- - the system's top-most object: the Zope root folder.
 
 - *context*- - the object to which the template is being applied.
--- a/docs/zopebook/AppendixC.rst
+++ b/docs/zopebook/AppendixC.rst
@@ -694,6 +694,11 @@ Note the names `root`, `context`, `container`, `template`, `request`, `user`, an
 `modules` are optional names supported by Zope, but are not required by the
 TALES standard.
 
+Note that the (popular) ``chameleon`` template engine implements ``attrs``
+and ``default`` not as standard variables but in a special way.
+Trying to change their value may have undefined effects.
+
+
 TALES Exists expressions
 ========================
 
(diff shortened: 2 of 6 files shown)


d-maurer:
Jens Vagelpohl wrote at 2020-7-3 00:52 -0700:
>@dataflake commented on this pull request.
>
>Would it be worth waiting with a final fix until/if Malthe fixes Chameleon?

I do not know.

Currently, `attrs` does not work at all. With this fix,
it works where it works with `chameleon` (currently in `tal:content`).
Unless `chameleon` changes in a strange way, the fix should not
hurt even after `chameleon` is fixed (it essentially adds
"try: attrs except NameError: attrs = None") - when `chameleon` is
fixed `attrs` will always be defined, thus the `try except` becomes
unnecessary but does not hurt. After the fix, the `test_attrs` test
can also be simplified - but this, too, is cosmetics only.

Maybe, we should ask @ale-rt (he reported the initial problem)
and @malthe for their opinion.

dataflake:
How does https://github.com/malthe/chameleon/pull/324 fit into the picture? Would it allow you to simplify your changes here?

d-maurer:
Jens Vagelpohl wrote at 2020-7-3 01:18 -0700:
>How does https://github.com/malthe/chameleon/pull/324 fit into the picture? Would it allow you to simplify your changes here?

Looks as if it would solve the `chameleon` problem -- and
the work arounds could go away.

But I have no more time today. In will check on this in the next
few days.

CHANGES.rst
```diff
@@ -10,6 +10,13 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- support the ``attrs`` predefined template variable again (as
+  far as ``chameleon`` allows it)
+  (`#860 <https://github.com/zopefoundation/Zope/issues/860>`_).
+
+- in the "Zope Book" remove ``CONTENTS`` from the list of predefined
```
vincentfretin, line 17: CONTEXTS you mean?

ale-rt:
I can confirm this is going to fix my issue :) Thanks a lot!

d-maurer:
Vincent Fretin wrote at 2020-7-3 01:28 -0700:
>@vincentfretin commented on this pull request.
>
>
>
>> @@ -10,6 +10,13 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
> 4.4.5 (unreleased)
> ------------------
>
>+- support the ``attrs`` predefined template variable again (as
>+  far as ``chameleon`` allows it)
>+  (`#860 <https://github.com/zopefoundation/Zope/issues/860>`_).
>+
>+- in the "Zope Book" remove ``CONTENTS`` from the list of predefined
>
>CONTEXTS you mean?

Indeed. Thank you! I will fix the typo.

d-maurer:
Dieter Maurer wrote at 2020-7-3 10:27 +0200:
>Jens Vagelpohl wrote at 2020-7-3 01:18 -0700:
>>How does https://github.com/malthe/chameleon/pull/324 fit into the picture? Would it allow you to simplify your changes here?
>
>Looks as if it would solve the `chameleon` problem -- and
>the work arounds could go away.

I am meanwhile convinced that we should keep the work arounds --
to avoid a high risk of breakage.

The conviction comes from the experience that one of the tests
failed with the latest `chameleon` patch.
The test failure means that there are still features
(`tal:on-error` in the concrete case) for which
`chameleon` does not yet provide `attrs` support.
Without the work arounds, a template will fail
if it uses any feature without `attrs` support by `chameleon`,
whether or not the template actually uses `attrs` in this feature.
With the work arounds, the template will only fail if it
actually uses `attrs` in an unsupported feature
(like with `chameleon` itself).

Our test suite is quite good **BUT** currently it does
not test `chameleon` only features (examples are
e.g. `$name` and `${expr}` substitutions outside `tal` attributes
and `tal:switch`). Because of the incompleteness of our
test suite, it may not detect all cases where the work arounds
are vital.

Thus, keep the work arounds unless @malthe can guarantee
that any `chameleon` feature has `attrs` support.

d-maurer:
Jens Vagelpohl wrote at 2020-7-4 05:36 -0700:
>@dataflake approved this pull request.

Please still wait with the merging.
I have a local spelling fix ("CONTENTS" -> "CONTEXTS"
in `CHANGES.rst`) not yet pushed to `github`.

dataflake:
I want you to do the merge when you think it's ready, I was just signaling my agreement with your previous comment to leave the workarounds from this PR in place.

malthe:
I think these issues should be fixed in Chameleon; for example, `attrs` should work where you expect it to, etc.

d-maurer:
Malthe Borch wrote at 2020-7-4 06:56 -0700:
>I think these issues should be fixed in Chameleon; for example, `attrs` should work where you expect it to, etc.

I agree with you.

I want to leave the work around just for safety reasons: we
do not have sufficient tests to cover all potential `chameleon` features.
With the work around, `attrs` works for a feature whenever `chameleon`
supports it for that feature; if this is not the case, `attrs`
cannot be used for this feature but otherwise nothing bad happens.
Without the work around, the template
breaks when it uses a feature for which `chameleon` does not support `attrs`
even if the template does not use `attrs` at all.
The work around is cheap (essentially
`try attrs except NameError: attrs = None`). Should breakage occur,
it is dramatic (total template failure even if the template does nothing
with `attrs`): thus, we can cheaply avoid potentially big damage.

d-maurer:
@dataflake I have fixed the spelling error, improved the comments related to `chameleon` work arounds and added a note to the Zope Book about the special implementation of `attrs` and `default` and its consequence. Perhaps you want to take a second look - especially on the new note to the Zope Book.

The travis tests currently fail with a problem apparently related to `isort`:
```
...
lint run-test: commands[0] | - isort --check-only --recursive /home/travis/build/zopefoundation/Zope/src setup.py

usage: isort [-h] [--src SRC_PATHS] [-a ADD_IMPORTS] [--ac] [--af] ...
```
It is highly unlikely that I caused the problem - and I have no idea how to solve it. Could you take a look?

d-maurer:
I was wrong: `zope.tales` **DOES** implement `CONTEXTS` but not as a regular "variable" (as the Zope Book suggested) but as a special case of the initial name in a (sub)path expression. This implies that `CONTEXTS` cannot be used in a Python expression but can be used as leading component in a path expression.The value of `CONTEXTS` is not a list (as specified in the Zope Book) but a mapping (as it should be to fulfill its design purpose). I doubt, however, that it will include the special "variable" `attrs`.

I will add a test for `CONTEXTS` and see what it really gives.

d-maurer:
> The travis tests currently fail with a problem apparently related to `isort`:
> 
> ```
> ...
> lint run-test: commands[0] | - isort --check-only --recursive /home/travis/build/zopefoundation/Zope/src setup.py
> 
> usage: isort [-h] [--src SRC_PATHS] [-a ADD_IMPORTS] [--ac] [--af] ...
> ```
> 
> It is highly unlikely that I caused the problem - and I have no idea how to solve it. Could you take a look?

Apparently, it tries to use a wrong `isort`.

dataflake:
`isort` had a new major release that breaks too many things right now. I pinned the older version.

dataflake:
FYI: I have forward-ported this to Zope `master` in https://github.com/zopefoundation/Zope/commit/eeb675c42f5e77a77e512745ba299bc42b6d5af4

Issue #862, opened 2020-07-03
Products.Pagetemplates: switch/case does not work anymore for Zope > 4.3


ale-rt:
Tested in Plone buildout.coredev.

This snippet:
```
        <div tal:define="key string:foo"
             tal:switch="key"
        >
          <tal:bar case="string:bar">bar</tal:bar>
          <tal:default case="default">baz</tal:default>
        </div>
```
Used to render:
```
<div>baz</div>
```
Now it does not render anything because `default` is not set to the value of key.

Note if I keep chameleon 3.8 and switch back to Zope 4.3 the behavior is the expected one

ale-rt:
As a workaround I can change the default with `<tal:default case="key">baz</tal:default>`

d-maurer:
Alessandro Pisa wrote at 2020-7-3 03:58 -0700:
>Tested in Plone buildout.coredev.
>
>This snippet:
>```
>        <div tal:define="key string:foo"
>             tal:switch="key"
>        >
>          <tal:bar case="string:bar">bar</tal:bar>
>          <tal:default case="default">baz</tal:default>
>        </div>
>```
>Used to render:
>```
><div>baz</div>
>```
>Now it does not render anything because `default` is not set to the value of key.

@malthe
I assume this is a `chameleon` issue:
recent `chameleon` versions have changed the `default` handling
from a local (ever changing) default marker towards a unique object.
Likely, this change was not applied to `tal:switch`.

ale-rt:
Note: if I keep chameleon 3.8 and switch back to Zope 4.3 the behavior is the expected one. So I am suspect more a change in Zope pagetemplate handling.

d-maurer:
Alessandro Pisa wrote at 2020-7-3 15:33 -0700:
>Note: if I keep chameleon 3.8 and switch back to Zope 4.3 the behavior is the expected one. So I am suspect more a change in Zope pagetemplate handling.

Yes and no:

The yes part:
`chameleon` used to work with local (ever changing) definitions
of the TALES `default` object. This means that with it, you did not have
a single `default` object but each template (and likely even each element
with a TAL/METAL attribute in a template) had its own representation
of `default`. This worked (in simple cases)
because the `chameleon.tales` TALES implementation has partial
access to the local variables of the calling rendering function --
among them the current `default` definition.
`zope.tales` is not designed to access local variables of the caller.
Therefore, it expects `default` to be a singleton, a unique global object
that can be used like any other value and does not need special treatment.
[...]

PR #865, opened 2020-07-06
add test for `tal:switch` - to verify #862

d-maurer:
This PR adds a test for `tal:switch` in relation to #862 .

Note that the test requires a new (upcoming) `chameleon` version to succeed. Do not merge until this version is used by `Zope`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- Add ``tal:switch`` test
+
 - Update dependencies to the latest releases that still support Python 2.
 
 
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -20,6 +20,12 @@ parts =
     requirements
 sources-dir = develop
 auto-checkout =
+    Chameleon
+
+
+[sources]
+Chameleon = git https://github.com/malthe/chameleon.git branch=patch/fix-static-attrs
+
 
 [testenv]
 PYTHONHASHSEED = random
--- a/src/Products/PageTemplates/tests/input/switch.html
+++ b/src/Products/PageTemplates/tests/input/switch.html
@@ -1,0 +1,12 @@
+<html>
+<head></head>
+<body>
+<div tal:switch="python:1"><tal:case
+  case="python:0">case 0</tal:case><tal:case
+  case="python:1">case 1</tal:case><tal:case
+  case="default">case default</tal:case></div>
+<div tal:switch="python:2"><tal:case
+  case="python:0">case 0</tal:case><tal:case
+  case="python:1">case 1</tal:case><tal:case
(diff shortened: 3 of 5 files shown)


dataflake:
@d-maurer Simply add the Chameleon checkout to `buildout.cfg` so running the buildout will fetch the correct branch and the test run, and then merge that into `4.x`. It is perfectly acceptable to have checkouts on the branch. The release manager is resposible for checking the buildout configuration and replacing it with a release version pin during the relase.

d-maurer:
Jens Vagelpohl wrote at 2020-7-5 23:06 -0700:
>@d-maurer Simply add the Chameleon checkout to `buildout.cfg` so running the buildout will fetch the correct branch and the test run, and then merge that into `4.x`. It is perfectly acceptable to have checkouts on the branch. The release manager is resposible for checking the buildout configuration and replacing it with a release version pin during the relase.

Thank you for the tip!

It will (initially) work in a PR development branch - but may fail in the
integrating 4.x branch -- when 2 branches want different checkouts.

As "github" (apparently) deletes development branches on merge,
it may also fail in a PR development branch.
For example: the current PR depends on the
`chameleon` branch `patch/fix-static-attrs`.
When @malthe merges it into `master`, it may no longer be available
for a checkout.
When I use the commit id (instead of the branch name),
then it becomes quite difficult for an integrator to determine
what is required in the integrating context.

dataflake:
In this case there's just one Chameleon branch where Malthe is fixing issues (`patch/fix-static-attrs`) so there won't be any conflict yet.

Deleting development branches when a PR is merged is a repository-specific GitHub setting. I always delete those branches, they don't serve a purpose anymore and just clutter up the branch list when you need to look for a specific branch.

If Malthe's branch is gone it's a 2 second change to either switch the buildout source configuration here to use the `master` branch or to delete the checkout configuration altogether if a release has happened.

I just added the Chameleon checkout to the buildout configuration and will merge it all to 4.x now so this PR and any issue connected to it can be closed.

malthe:
You could also just refer to the commit rather than the development branch.

On Github, the commit is attached to an internal namespace as part of the completed pull request so the commit object would not go away even on garbage collection.

dataflake:
I didn't want to tie it to a specific commit because the branch is a moving target and you seem to be adding more fixes.

malthe:
I have released Chameleon 3.8.1 with these changes:

3.8.1 (2020-07-06)
------------------

- Added code optimization to reduce sequential appends of static text.

- The `default` symbol in dynamic attributes is now
  symbolic. Previously, it was assigned the string value of the
  default attribute text. A similar change has been made for
  switch/case expressions.

- The built-in `attrs` dictionary of static element attributes now
  correctly works with `tal:define`, etc.

- Fix slice code generation compatibility issue on Python 3.9.

dataflake:
Thank you Malthe, I have updated the version pins for Zope `4.x` and `master`

Issue #863, opened 2020-07-03
Products.PageTemplates: should (some) builtins become legal path expressions?


d-maurer:
In "https://github.com/plone/Products.CMFPlone/issues/3135#issuecomment-653429483"
@vincentfretin asks whether ` tal:omit-tag="True"` can again become legal.

With `Zope`, the default TALES expression type is `path` and according to the Zope Book any path expression starts with a "variable" defined as either a predefined variable (such as `nothing`, `default`, `context`, ...) or a variable defined by the template. According to this specification, `True` is not a valid path expression -- and is rejected.

The former `chameleon.tales` based TALES implementation was less strict: a path expression did not need to start with a variable, some builtins ( I do not yet know the precise set - but it included `True` and `False`) and internals (that included `default`) were allowed as well.
And what is allowed and fairly natural (like the `tal:omit-tag=True`) has a good chance to have been used somewhere. And those places break now with the stricter requirements on path expressions.

Should we relax the requirements on path expressions (and thus make expressions like the above legal)?

Should we  support as leading element of a path expression beside a variable
only a small set of builtins (including `True` and `False`) or any legal builtin (i.e. any Python builtin for a trusted engine and any `AccessControl.safe_builtins` for a untrusted engine)?

dataflake:
If it's not too much work then anything that makes "pure" Chameleon and our version more similar is a win as long as security isn't affected negatively.

Issue #864, opened 2020-07-04
Products.PageTemplates: should we partially reinstate TALES engine guessing?


d-maurer:
"https://github.com/plone/Products.CMFPlone/issues/3141#issuecomment-6537251" reveals a problem introduced by Zope 4.4 fully honoring the return value of a template's `pt_getEngine`.

By design, the TALES implementation to be used by a template is determined by the return value of its `pt_getEngine` method. Until `Zope 4.4`, the `chameleon` integration mostly ignored this return value: it was just used to guess whether the template likely wants a trusted or untrusted TALES implementation. Other details were ignored and the integration used a TALES engine of its own just making the distinction "trusted/untrusted". With Zope 4.4, the return value of `pt_getEngine` is fully honored.

Plone integrates packages originally designed for "Zope3", e,g. `z3c.form`. Those packages likely derive their templates (it they use templates) directly from `zope.pagetemplate` and not from `Products.PageTemplates`. Those templates have a `Zope` unaware `pt_getEngine`. Before Zope 4.4, its return value was ignored by the `chameleon` integration and replaced by a trusted `Zope` aware engine; with Zope 4.4, those templates effectively loose `Zope` awareness (e.g. path expressions no longer use `boboAwareZopeTraverse` but instead "simple traversing").

The clean solution would be to require integrators of foreign packages into the Zope world to make  templates `Zope` aware. But we could also introduce a hack - maybe along the following lines. We know that a template whose `pt_getEngine` returns a "base TALES engine" (one typically used by `zope.pagetemplate`) cannot fully work inside `Zope`. In those cases, we replace it by a `Zope` trusted engine. This would reinstate the behavior before Zope 4.4 for those cases.

dataflake:
If those "unaware" foreign packages can be detected easily at runtime and any "normal" handling is unaffected, sure.

d-maurer:
Jens Vagelpohl wrote at 2020-7-4 01:42 -0700:
>If those "unaware" foreign packages can be detected easily at runtime and any "normal" handling is unaffected, sure.

Of course, this is impossible. Any approach which
does not fully honor the engine returned by a template's
`pt_getEngine` may do something against the intentions
of the template's author.

As the word "guessing" indicates, such an approach would necessarily
use an heuristic and consequently can guess wrongly.

But we could make the guessing better than that of `Zope` before 4.4.


The `chameleon` integration in `Zope` before 4.4 guessed as follows:

   use the integration's untrusted TALES engine if
[...]

PR #866, opened 2020-07-06
allow (some) builtins as first element in a path expression

d-maurer:
This PR is one possible implementation for "https://github.com/zopefoundation/Zope/issues/863".

It interprets "some builtins" as:
 * in an untrusted context: the builtins from `AccessControl.safe_builtins`
 * in a trusted context: all Python builtins and those from `AccessControl.safe_builtins`

Thus, the set of allowed builtins is likely larger than that natively supported by `chameleon`. I believe that `chameleon` currently does not expose its set for external access; this would be necessary to ensure that the same set of builtins is supported.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,13 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- Allow (some) builtins as first element of a (TALES) path expression:
+  in an untrusted context, the builtins from
+  ``AccessControl.safe_builtins`` are allowed;
+  in a trusted context, all Python builtins are allowed in addition
+  (and take precedence)
+  (`zope.tales#23 <https://github.com/zopefoundation/zope.tales/issues/23>`_).
+
 - Add ``tal:switch`` test
 
 - Support the ``attrs`` predefined template variable again (as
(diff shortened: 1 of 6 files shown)


d-maurer:
Some builtins will be rejected because their name is not supported (e.g. starts with `_`).

dataflake:
What are the security implications of this patch? Is it just making all those names available in path expressions that were already available in python expressions, or does it add a lot of additional access that wasn't available at all before?

d-maurer:
> What are the security implications of this patch? Is it just making all those names available in path expressions that were already available in python expressions, or does it add a lot of additional access that wasn't available at all before?

I expect that now only builtins can be accessed as path expressions which formerly were
already accessible in Python expressions.

Behind this expectation are the following assumptions:

 * an unrestricted Python expression can access any Python builtin

 * a restricted Python expression can access any builtin from `AccessControl.safe_builtins` provided its name does not start with `_` (those are considered internal) (the restricted Python `ast` transformation rejects names starting with `_`).

I am quite sure that the second assumption is correct: as far as I know a restricted Python expression is first parsed into `ast` normally; the `ast` is then processed by the restricted Python transform - this transforms/checks as necessary (in particular, names starting with `_` are rejected); the result is evaluated with a copy of `AccessControl.safe_builtins` as "global dict" (and thus can access its builtins).

The path expression infrastructure enforces that the first element in a path expression has a legal variable name, especially does not start with `_`.

dataflake:
Thanks for the explanation. I believe https://github.com/zopefoundation/Zope/issues/864 won't be affected by this patch because there's nothing "new" being added to whatever the different types of TAL expressions were already able to reach. Just trying to stay on top of the "big picture".

Issue #859, opened 2020-07-01
Add `pyupgrade` to linter setup


icemac:
It makes sense to add `pyupgrade` with the current options as in #857 to our linter setup.

This prevents the `return` of old syntax with new pull requests.

_Originally posted by @jugmac00 in https://github.com/zopefoundation/Zope/issues/841#issuecomment-651639713_

dataflake:
This is not straightforward. `pyupgrade` requires to specify input files, it does not work on folders recursively, so you must construct some unwieldy command line that finds the files first and then feeds them to `pyupgrade`. The alternative that's supposed to work around that, `pyupgrade-directories`, insists on writing all file paths it finds to stdout, there is no flag to suppress that, so the output becomes impossible to visually parse.

jugmac00:
When we run `pyupgrade` only manually from time to time, two things will happen:
- old syntax is used again
- maybe not all files are catched, as above mentioned it is not easy to catch em all

Actually, both things already happened.

In the initial upgrade run `utils.py` was forgotten, and not even 24 hours ago, several lines of code with old syntax was introduced again.

Let me state that clearly - I do not mind. But... if you want new syntax, why don't we enforce it?

For the problem with catching all files, there is a modern, and easy solution. By the same maintainer as `tox`, `flake8`, `pyupgrade` and many more tools.

It is called `pre-commit`  ( https://pre-commit.com/ ). FWIW - I use it for quite some time, and it is really comfortable.

`pre-commit` is a linter runner. It does all the heavy lifting of finding files (see comment above), and it also pins the versions (but also makes it super easy to update the versions: `pre-commit autoupdate`).

You can use it in two ways - either run `pre-commit` as a tox env on demand, or install it locally, and everytime you `git commit` it runs, and makes things like `fix flake8` commits a thing of the past, as you can only commit if there are no linter errors (and yes, you can override that).

I created a proof of concept pr - see #867 

Attention - this looks like much, but I created two commits - one for the setup and one to fix the introduced old syntax.

Setup
https://github.com/zopefoundation/Zope/pull/867/commits/630db55166a2c1f5e0f0d313e841f4525f1d8b7d

Syntax fixes
https://github.com/zopefoundation/Zope/pull/867/commits/849d2b9a6ae76bee427610d511ccc6f5daedb50c


[...]

PR #867, opened 2020-07-07
Pre commit

jugmac00:
see #859


icemac:
It would have been nice to be able to see how the errors are reported on TravisCI.

jugmac00:
Hi @icemac - I'll comment in the issue #859 so the discussion does not spread into too many places.

jugmac00:
I close this PR as it was replaced by #868

PR #868, opened 2020-07-08
Add pyupgrade via pre-commit

jugmac00:
`pre-commit` is a linter runner, which does all the heavy lifting of finding the correct files to lint, recursively.

Also, `pre-commit` both able to pin package versions, and to autoupdate all at once if desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ matrix:
           env: TOXENV=py38
         - python: "3.9-dev"
           env: TOXENV=py39
-        - python: "3.8"
-          env: TOXENV=pre-commit
     allow_failures:
         - python: "3.9-dev"
           env: TOXENV=py39
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ matrix:
           env: TOXENV=py38
         - python: "3.9-dev"
           env: TOXENV=py39
+        - python: "3.8"
+          env: TOXENV=pre-commit
     allow_failures:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -13,6 +13,8 @@ matrix:
           env: TOXENV=py38
         - python: "3.9-dev"
           env: TOXENV=py39
+        - python: "3.8"
+          env: TOXENV=pre-commit
```
icemac, line 17: Maybe this job should be allowed to fail.

dataflake: Allowing it to fail makes this completely pointless. Step back for a second and think about who the audience is, what we gain and what price we pay.

If you add this and allow failure then no one except for core developers will bother to look at the Travis results because Travis says "it's all OK". At the same time we lose because any added Travis job makes the tests run even longer. We gain nothing and lose out on loger Travis runs.

icemac: Okay, you got me: We should either have it as normal job in Travis or not at all.

jugmac00: I removed the Travis run and added a description for the `tox` environment which you can list with `tox -av` or `tox -lv`.

jugmac00:
If there is anything left or some oversight - I will have a look later this evening.

jugmac00:
Rebased on current master - and, once again, applied `pyupgrade` on current code base.

Should be ready to merge when tests pass - will have another look in an hour.

PR #869, opened 2020-07-08
`IZopeAwareEngine` to address #864

d-maurer:
This PR addresses #864 .

It defines a new interface `Products.PageTemplates.interfaces.IZopeAwareEngine` which conceptionally is a marker indicating that a TALES engine has a `Zope` compatible `PathExpr`.

The PR marks the "Engine" classes defined by `Products.PageTemplates` with `IZopeAwareEngine` even though Zope awareness is not a property of the class but of the individual instances and their `PathExpr`. The motivation here is that components which derive from or instantiate those classes are likely providing a Zope compatible `PathExpr`.

`IZopeAwareEngine` can be used as the `provides` of an adapter registration in order to adapt a non `IZopeAwareEngine`. An adapter registered in this way is used only for engines which do not provide `IZopeAwareEngine` by themselves (currently, adaptation only takes place for template rendering via `chameleon` -- this may change in the future). Therefore, the code fragment below can restore Zope 4.3 behavior regarding Zope awareness for non `Zope` TALES engines:
```
from zope.component import provideAdapter
from Products.PageTemplates.Expressions import getTrustedEngine
from Products.PageTemplates.interfaces import IZopeAwareEngine

def adapt(engine):
    return getTrustedZopeEngine()

provideAdapter(adapt, (None,), IZopeAwareEngine)
```
Note: likely, one would put the adapter registration into ZCML; this way, it could be "unconfigured" if necessary.

Note: the `adapt` above reflects what Zope 4.3 effectively did.A more careful adaptation would look at the adapted *engine* to determine whether a trusted or untrusted Zope aware engine should be used and return the appropriate engine.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,16 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- New interface ``Products.PageTemplates.interfaces.IZopeAwareEngine``.
+  It can be used as the "provides" of an adapter registration
+  to adapt a non ``Zope`` tales engine to an engine to be used
+  by ``Zope`` page templates
+  (`#864 <https://github.com/zopefoundation/Zope/issues/864>`_).
+  Currently, the adaptation is used only when the
+  template is rendered with ``chameleon``;
+  with ``zope.pagetemplate``, the engine is used
+  as is - this may change in the future.
+
 - Allow (some) builtins as first element of a (TALES) path expression:
   in an untrusted context, the builtins from
   ``AccessControl.safe_builtins`` are allowed;
--- a/src/Products/PageTemplates/Expressions.py
+++ b/src/Products/PageTemplates/Expressions.py
@@ -18,6 +18,8 @@ for Python expressions, string literals, and paths.
 
 import logging
 
+from Products.PageTemplates import ZRPythonExpr
+from Products.PageTemplates.interfaces import IUnicodeEncodingConflictResolver
 from six import binary_type
 from six import text_type
 
@@ -25,8 +27,6 @@ import OFS.interfaces
 from AccessControl import safe_builtins
 from Acquisition import aq_base
 from MultiMapping import MultiMapping
-from Products.PageTemplates import ZRPythonExpr
-from Products.PageTemplates.interfaces import IUnicodeEncodingConflictResolver
 from zExceptions import NotFound
 from zExceptions import Unauthorized
 from zope.component import queryUtility
@@ -48,6 +48,8 @@ from zope.tales.tales import ErrorInfo as BaseErrorInfo
 from zope.tales.tales import Iterator
 from zope.traversing.adapters import traversePathElement
 from zope.traversing.interfaces import ITraversable
+
+from .interfaces import IZopeAwareEngine
 
 
 SecureModuleImporter = ZRPythonExpr._SecureModuleImporter()
@@ -325,6 +327,12 @@ class ErrorInfo(BaseErrorInfo):
     __allow_access_to_unprotected_subobjects__ = True
 
 
+# Whether an engine is Zope aware does not depend on the class
+# but how it is configured - especially, that is uses a Zope aware
+# `PathExpr` implementation.
+# Nevertheless, we mark the class as "Zope aware" for simplicity
+# assuming that users of the class use a proper `PathExpr`
+@implementer(IZopeAwareEngine)
 class ZopeEngine(Z3Engine):
 
     _create_context = ZopeContext
--- a/src/Products/PageTemplates/Expressions.py
+++ b/src/Products/PageTemplates/Expressions.py
@@ -18,8 +18,6 @@ for Python expressions, string literals, and paths.
 
 import logging
 
-from Products.PageTemplates import ZRPythonExpr
-from Products.PageTemplates.interfaces import IUnicodeEncodingConflictResolver
 from six import binary_type
 from six import text_type
 
@@ -49,6 +47,8 @@ from zope.tales.tales import Iterator
 from zope.traversing.adapters import traversePathElement
 from zope.traversing.interfaces import ITraversable
 
+from . import ZRPythonExpr
+from .interfaces import IUnicodeEncodingConflictResolver
 from .interfaces import IZopeAwareEngine
 
 
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -20,12 +20,12 @@ from chameleon.exc import ExpressionError
 from chameleon.tal import RepeatDict
 from chameleon.tales import DEFAULT_MARKER  # only in chameleon 3.8.0 and up
 from chameleon.zpt.template import Macros
-from z3c.pt.pagetemplate import PageTemplate as ChameleonPageTemplate
 
 from AccessControl.class_init import InitializeClass
 from AccessControl.SecurityInfo import ClassSecurityInfo
 from App.version_txt import getZopeVersion
 from MultiMapping import MultiMapping
+from z3c.pt.pagetemplate import PageTemplate as ChameleonPageTemplate
 from zope.interface import implementer
 from zope.interface import provider
 from zope.pagetemplate.engine import ZopeBaseEngine
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -20,12 +20,12 @@ from chameleon.exc import ExpressionError
 from chameleon.tal import RepeatDict
 from chameleon.tales import DEFAULT_MARKER  # only in chameleon 3.8.0 and up
 from chameleon.zpt.template import Macros
+from z3c.pt.pagetemplate import PageTemplate as ChameleonPageTemplate
 
 from AccessControl.class_init import InitializeClass
 from AccessControl.SecurityInfo import ClassSecurityInfo
 from App.version_txt import getZopeVersion
 from MultiMapping import MultiMapping
-from z3c.pt.pagetemplate import PageTemplate as ChameleonPageTemplate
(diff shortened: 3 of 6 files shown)


d-maurer:
Help needed: the failing tests come from an unhappy `isort`. The problem: I have sorted those files with my local `isort` -- apparently, the "travis" `isort` and my local `isort` do not agree on the right import order. I have tried to find a reason for those differences - but failed.

dataflake:
I'll look at it

dataflake:
It looks like you were not using the version of `isort` used by our `tox` configuration. Our `isort` is currently pinned at `<5` because the recent major upgrade is causing issues and I did not have the time to look at those in detail, so `isort` is pinned in `tox.ini` for now.

To use the correct `isort` either run `tox -elint` or (after one `tox` run) the `isort` script at `.tox/lint/bin/isort`.

dataflake:
Note: I think this is (so far) the last urgent issue/PR and I could do a 4.x release. Or do you see anything else that should be fixed?

d-maurer:
Jens Vagelpohl wrote at 2020-7-8 03:15 -0700:
>It looks like you were not using the version of `isort` used by our `tox` configuration. Our `isort` is currently pinned at `<5` because the recent major upgrade is causing issues and I did not have the time to look at those in detail, so `isort` is pinned in `tox.ini` for now.

I am using `isort=4.3.21`.


>To use the correct `isort` either run `tox -elint` or (after one `tox` run) the `isort` script at `.tox/lint/bin/isort`.

I will try this. Thank you for your help!

dataflake:
I'm surprised - I thought you were probably using `isort` 5 or higher. Your version matches what `tox` uses. But anyway, issue solved for now.

Like I said, merge when you're ready and if you can't think of anything I should wait for I'll plan on a 4.x release on short notice.

d-maurer:
Jens Vagelpohl wrote at 2020-7-8 03:15 -0700:
> ...
>To use the correct `isort` either run `tox -elint` or (after one `tox` run) the `isort` script at `.tox/lint/bin/isort`.

Apparently, it is not a question of the `isort` version:

When I run `tox -elint`, I get (precisely) the errors reported
by "travis". Thus, I can locally check.

When I run `.tox/lint/bin/isort --check-only --recursive src`, I get errors in
almost all files - the same errors as with the system `isort`.
By some magic `tox` must have given `isort` different parameters
than when `isort` is run directly.

I am not interested in the precise import order.
Thus, I look for a way to arrange this order in a way that
"travis" will be happy. In `tox.ini` I have seen an `apply-isort`
which may be applicable. Can you tell me how I activate it?

jugmac00:
You can run any `tox` envs by 
```
tox -e name-of-env
```

So you can run the `apply-isort` one via `tox -e apply-isort`.

I am not sure why we have isort configured in two tox envs - I will have a look later.

dataflake:
All I can say is that the way I described running `isort` works for me locally and Travis is green as well. Have you pulled the latest changes into your branch?

The `apply-isort` tox environment seems to be a convenience thing added at some point, I don't use it. It's not run during normal tox runs, you must invoke it manually.

d-maurer:
@dataflake 
> ... if you can't think of anything I should wait for I'll plan on a 4.x release on short notice.

I remember that the original TALES specification mentioned a `path` function. This function can be used in Python expressions to evaluate a path expression, e.g. `python:path("`*path*`")` is equivalent to the path expression *path*. I have not yet seen a test for this feature and it might be possible that it fails. I would like to look into this tomorrow.

dataflake:
Great idea, I'll hold off with the release then.

d-maurer:
Jens Vagelpohl wrote at 2020-7-8 15:08 +0000:
>All I can say is that the way I described running `isort` works for me locally and Travis is green as well. Have you pulled the latest changes into your branch?

No. That's why I still get the errors from `tox -e lint`.
I am sure, they will disappear when I pull the current state
(because you have fixed the `isort` problems).

My questions had the purpose to solve `isort` problems on
my own in the future.

dataflake:
On the `4.x` and `master` branches I have added a command line switch to the `isort` invocation from `tox[lint]` that will print out a diff showing you exactly where `isort` wants the imports.

d-maurer:
Jens Vagelpohl wrote at 2020-7-8 08:16 -0700:
>Great idea, I'll hold off with the release then.

@dateflake
There already was a "path function" test:
`test_persistenttemplate.TestPersistent.test_path_function`.

No objection from me for a release.

However, you may consider including
"https://github.com/zopefoundation/Zope/pull/870".
This is cosmetics only: it avoids that each test from
`testHTMLTests.HTMLTests` is executed 3 times by the test runner.

d-maurer:
Jürgen Gmach wrote at 2020-7-8 08:04 -0700:
>You can run any `tox` envs by
>```
>tox -e name-of-env
>```
>
>So you can run the `apply-isort` one via `tox -e apply-isort`.
>
>I am not sure why we have isort configured in two tox envs - I will have a look later.

Thank you for the tip!

However, `tox -e isort-apply` changes the import order in hundred of
files - and those are then rejected by `tox -e lint`.
Something is really strange with the `isort` integration.

dataflake:
Like I said, I never looked at or used that `isort-apply` tox environment. It may very well be bit-rotted. I'll take a look.

dataflake:
P.S.: I am just noticing that it does not apply the `isort` version pin. There's your reason right there.

mauritsvanrees:
I have merged https://github.com/plone/Products.CMFPlone/pull/3149, which means Plone 5.2.2 will now register an `IZopeAwareEngine` adapter.  Seems to work as intended, and fix a reported bug.

If wanted, I could make a PR to do this in Zope instead, probably in `Products.PageTemplates`. Then I would revert my Plone change.

Meanwhile, a Zope 4.5.1 would be useful, with the latest template-related fixes, and maybe with such a PR if wanted. That should solve the last big hurdle for a Plone 5.2.2 release (except that I want to add one more non-Zope related change since today...)

d-maurer:
Maurits van Rees wrote at 2020-7-21 11:57 -0700:
> ...
>If wanted, I could make a PR to do this in Zope instead, probably in `Products.PageTemplates`. Then I would revert my Plone change.

I would prefer to have the adapter registration **not** in `Zope`:
`Zope` itself has nothing to do with templates using a strange
engine; therefore, it should not be responsible for guessing
the right adaptation.

In my view, this guessing is the responsiblity of the integrator:
he can know which engines are used in his application
and decide what adaptations are necessary.

>Meanwhile, a Zope 4.5.1 would be useful, with the latest template-related fixes, and maybe with such a PR if wanted. That should solve the last big hurdle for a Plone 5.2.2 release (except that I want to add one more non-Zope related change since today...)

Please wait a bit. I just made a PR to fix an "unsolicited translations" bug.

mauritsvanrees:
The above remaining problem has been fixed and merged a week ago, so I think we are good to go.
Can we have a Zope 4.5.1 please?

dataflake:
ZODB 5.6 is not compatible with current ZEO releases and there are still outstanding page template issues.

d-maurer:
Maurits van Rees wrote at 2020-7-30 12:52 -0700:
>The above remaining problem has been fixed and merged a week ago, so I think we are good to go.
>Can we have a Zope 4.5.1 please?

If possible, we should also have a new release of `zope.sendmail` --
as a prerequisite to finalize and merge a `Products.MailHost` PR.

d-maurer:
Jens Vagelpohl wrote at 2020-7-30 14:12 -0700:
>ZODB 5.6 is not compatible with current ZEO releases and there are still outstanding page template issues.

I know only of one important page template issue (unsolicited translations).
It is now well understood and the translation function can
work around it. `chameleon` would need a new configuration option
(controlling what should be considered an "i18n message") to
solve it cleanly - but I have the feeling that we will not get
it in the near future.

Besides that I know only of minor issues regarding suboptimal
error reports -- to be fixed in `chameleon`.

icemac:
> If possible, we should also have a new release of `zope.sendmail` --

I think this is because of https://github.com/zopefoundation/zope.sendmail/pull/29. After it got an entry in the change log I'd be happy to cut a release.

icemac:
https://pypi.org/project/zope.sendmail/5.1/ just released.

dataflake:
> I know only of one important page template issue (unsolicited translations). It is now well understood and the translation function can work around it. `chameleon` would need a new configuration option (controlling what should be considered an "i18n message") to solve it cleanly - but I have the feeling that we will not get it in the near future. Besides that I know only of minor issues regarding suboptimal error reports -- to be fixed in `chameleon`.

What's the status of that translation issue, is it fixed on our side?

d-maurer:
Jens Vagelpohl wrote at 2020-8-6 06:38 -0700:
>> I know only of one important page template issue (unsolicited translations). It is now well understood and the translation function can work around it. `chameleon` would need a new configuration option (controlling what should be considered an "i18n message") to solve it cleanly - but I have the feeling that we will not get it in the near future. Besides that I know only of minor issues regarding suboptimal error reports -- to be fixed in `chameleon`.
>
>What's the status of that translation issue, is it fixed on our side?

As far as we can.

The bag caching (causing an internal `chameleon` data type arriving
at application code and the translation function) is fixed (in 4.x).

The potential problems caused by `chameleon`'s wide notion
of an i18n message cannot be fixed by `Zope` (unless `chameleon`
provides a new configuration option). It must be handled in
the application specific translation function: that must look
at the type of the msgid and the domain and do the right thing
based on those.

mauritsvanrees:
With a new ZEO release finally here, I think we are good for a new Zope 4.x release.

dataflake:
@mauritsvanrees Sorry, I had lost track of the various tasks that blocked 4.5.1. I'm doing that today.

dataflake:
Zope 4.5.1 is now public.

mauritsvanrees:
Thanks! Much appreciated!

PR #871, opened 2020-07-09
make "Unicode Conflict Resolution" available for `chameleon` rendering

d-maurer:
This should fix https://github.com/plone/Products.CMFPlone/issues/3145

`Products.PageTemplates.Expressions.ZopeContext` has provisions for so called "Unicode Conflict Resolution", i.e. logic to convert bytes to unicode. This was never supported for `chameleon` (which uses its own independent logic for this case).

Due to a bug, it became not available with the combination of `zope.tales` and `chameleon`. This bug is fixed by this PR and now "Unicode Conflict Resolution" is available for `chameleon` rendering as well.

The "Unicode Conflict Resolution" relies on extensions by `ZopePageTemplate`, it does not work for a raw `PageTemplate`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.4.5 (unreleased)
 ------------------
 
+- Make "Unicode Conflict Resolution" available for templates
+  rendered with ``chameleon``
+  (`Products.CMFPlone#3145
+  <https://github.com/plone/Products.CMFPlone/issues/3145>`_).
+
 - New interface ``Products.PageTemplates.interfaces.IZopeAwareEngine``.
   It can be used as the "provides" of an adapter registration
   to adapt a non ``Zope`` tales engine to an engine to be used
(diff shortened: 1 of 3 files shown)


d-maurer:
Jens Vagelpohl wrote at 2020-7-9 03:43 -0700:
>@dataflake approved this pull request.

I still have two more improvements:

 1. currently, "unicode conflict resolution" works only for
    `ZopePageTemplate` (and not a raw `PageTemplate`).

    The cause of this is targeting the
    `PreferredCharsetUnicodeConflictResolver`. As sketched earlier,
    it looks at the request to find the HTTP header
    `ACCEPT_CHARSETS`. In order to access the request,
    we must have a `ZopePageTemplate`.

    But there are alternative resolvers which can work with
    a raw `PageTemplate` -- and the "PreferredCharset"-resolver
    can be made resilient to the case that it cannot access the request.

    Thus, I would like to ensure that "unicode conflict resolution"
    works with any template and no longer requires a `ZopePageTemplate`.

 2. "unicode conflict resolution" was lost because the
    "chameleon/zope" context wrapper was not faithful with
    respect to the incoming zope context.

    I changed it in a way that "unicode conflict resolution" is
    now available -- but the wrapper is still not faithful.

    I would like to improve its faithfulness.

dataflake:
Those are all great ideas. I don't mind releases in quick succession.

PR #873, opened 2020-07-09
make "chameleon zope context wrapping" more faithfull

d-maurer:
This PR improves the faithfulness of the "chameleon zope context wrapping".

`chameleon` evaluates TALES expressions in a do called `Scope` -- a mapping from (variable) names to their current values.

`zope.tales` TALES expressions are functions returning a value when called with a "context". The "context" is expected to know about the current name to value mapping and to know the TALES engine. The precise details of the "context" are determined by the TALES engine.

The "chameleon zope context wrapper" integrates `chameleon`'s `Scope` and `zope.tales`' "context" into one object. Its task is to manage variable settings in the `Scope` and everything else in the "context".

Formerly, the wrapper delegated data to the correct subobjects; but it took much of its implementation directly from `ZopeContext`, the class of the "context" used by `Products.PageTemplates`. This is wrong in the case that the "context" is not an instance of `ZopeContext`. This PR lets the wrapper delegate unknown method calls to the "context" rebindung their `__self__` to ensure that those methods access the correct "variable -> value" mapping.

### DIFF
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -187,6 +187,9 @@ class Name2KeyError(Mapping):
     def __iter__(self):
         return iter(self.data)
 
+    def __len__(self):
+        return len(self.data)
+
 
 class _C2ZContextWrapperBase(object):
     """Behaves like "zope" context with vars from "chameleon" context.
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -12,6 +12,7 @@ in the future.
 import ast
 import logging
 import re
+from weakref import ref
 
 from chameleon.astutil import Static
 from chameleon.astutil import Symbol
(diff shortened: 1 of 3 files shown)


d-maurer:
I want to change the implementation. The current approach will not work when a "context" uses `super`.

src/Products/PageTemplates/engine.py
```diff
@@ -215,7 +214,26 @@ def beginScope(self, *args, **kw):
...
+    def __setattr__(self, attr, v):
+        setattr(self.__z_context, attr, v)
+
+    def __delattr__(self, attr):
+        delattr(self.__z_context, attr)
+
+    # auxiliary
+    def _setattr(self, attr, v):
+        # currently used only for *attr* with ``__`` prefix
+        super(_C2ZContextWrapper, self).__setattr__(
```
dataflake, line 235: You can shorten `super(_C2ZContextWrapper, self)`  to just `super()` and get rid of the line break

d-maurer: This works in Python 3 but not in Python 2:
`TypeError: super() takes at least 1 argument (0 given)`

dataflake: You're right, sorry for the noise

PR #878, opened 2020-07-21
avoid unsolicited translations

### LINKED ISSUE #876: tal:content and tal:replace are sometimes treated as to be translated
I originally posted this in https://github.com/malthe/chameleon/issues/326 but it was closed there as `wontfix` with the hint
 
> To work around this issue, make sure expression is coerced to a string or unicode in _compile_zt_expr.
> That said, the approach taken is quite different from how Chameleon was first imagined. The idea with expression compilers is that they generate code at compile time, not at render time (even if it is cached). I would imagine that quite a bit of performance is lost due to the extra function calls and scope changes.

We have seen some weird behavior when switching to Zope4 that seems to be related to the Chameleon cache.

We use a translation service that effectively delegates translation to a Python Script in the context of the page template in question. But for the sake of providing an MWE, I stripped it down to something that simply returns a fixed text whenever something is to be translated. You can find it [here](https://github.com/perfact/Products.PerFactTranslationDomain/tree/dummy-translator).

If using a page template like

    <p i18n:translate="">original text</p>

This correctly replaces the content of the `p` tag by `translated`.

However, sometimes it also replaces the content of

    <p tal:content="string:original text">something</p>

and the same with `tal:replace`. By trying to nail down the problem, I found that this happens if the Chameleon cache file for the given page template has been created by the currently running Zope instance. So if I create a new page template, the behavior can be observed. If I restart Zope, it is gone. If I stop Zope, clear the Chameleon cache and start it again, it can again be observed.

Trying to debug the issue, I found myself inside the Chameleon cache file in a code block like

        def __quote(target, quote, quote_entity, default, default_marker):
            __tt = type(target)
            if ((__tt is int) or (__tt is float) or (__tt is int)):
                target = str(target)
            else:
                if (__tt is bytes):
                    target = decode(target)
                else:
                    if (__tt is not str):
                        try:
                            target = target.__html__
                        except:
                            __converted = convert(target)
                            target = (str(target) if (target is __converted) else __converted)
                        else:
                            return target()
                if (target is not None):
                    ...

If we are in the restarted run (where `tal:replace` does not get translated), `target` is of type `str` (with content `original text`) and we land in the last branch of the evaluation, where the string is being escaped.

However, if the Chameleon cache was cleared before starting Zope, `target` is of type `<class 'chameleon.tokenize.Token'>`. Since such a `Token` does not have the `__html__` attribute, `convert(target)` is called, which somehow ends up calling the translation service.

There is one thing most probably broken here and one thing that I maybe simply am not using correctly: One is that the type of `target` changes depending on if there was already a Chameleon cache present when Zope was started, leading to different behaviors. The second is that the `econtext` given to the `render` function contains the fields

    {'__convert': functools.partial(<function PageTemplate.render.<locals>.translate at 0x7fb5b35e5af0>, target_language=None),
     '__decode': <function PageTemplate.render.<locals>.decode at 0x7fb5b35e5ca0>,
     '__on_error_handler': None,
     '__translate': <function PageTemplate.render.<locals>.translate at 0x7fb5b35e5af0>,
    ...

Not sure if `convert` should point to a `translate` function if there is a specific `__translate` field also which seems to be the correct one for calling tags with `i18n:translate`.

### Steps to reproduce

    python3 -m venv Zope4
    cd Zope4
    bin/pip install -r https://raw.githubusercontent.com/zopefoundation/Zope/4.5/requirements-full.txt
    bin/pip install https://github.com/perfact/Products.PerFactTranslationDomain@dummy-translator
    bin/mkwsgiinstance -d zope -u admin:admin
    bin/runwsgi zope/etc/zope.ini

Open a browser at `http://localhost:8080/manage_main` and provide the credentials `admin:admin`. Now add a page template `testpt` with the content

    <p tal:content="string:replaced text">original text</p>

and let it render (`http://localhost:8080/testpt`)

### Expected behavior

We should obtain a paragraph with the content `replaced text`.

### Actual behavior

On the first try, we obtain a paragraph with the content `translated`. If restarting Zope, we again obtain the expected behavior. If removing everything in `zope/var/cache/` and restarting Zope, we again obtain `translated`.

### Versions

As seen above, I took the newly released Zope 4.5 as starting point. It contains Chameleon 3.8.1

d-maurer:
This PR fixes #876 

When a template is compiled by `chameleon` expressions are represented in the following execution by `chameleon.tokenize.Token`s, when the template code comes from the `chameleon` cache, expressions are represented in the execution by simple text. The `zope.tales` integration has been partially type preserving: in some cases a `Token` input resulted in a `Token` output. Under Python 3, `chameleon` interprets a `Token` result wrongly as "to be translated". The result has been unsolicited translations when a template was compiled by the current process and correct translations when the template came from the cache.

This PR works around this behavior by explicitly converting `Token` into `str` under Python 3.

### DIFF
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -14,6 +14,7 @@ import logging
 import re
 from weakref import ref
 
+import six
 from chameleon.astutil import Static
 from chameleon.astutil import Symbol
 from chameleon.codegen import template
@@ -135,6 +136,13 @@ def _compile_zt_expr(type, expression, engine=None, econtext=None):
     """
     if engine is None:
         engine = econtext["__zt_engine__"]
+    # *expression* is a ``chameleon.tokenize.Token`` when
+    # the template is compiled but "text" when the template code
+    # comes from the ``chameleon`` cache.
+    # Under PY3, ``chameleon`` wrongly translates ``Token``;
+    # convert to ``str`` to avoid this
+    if six.PY3:
+        expression = str(expression)
     key = id(engine), type, expression
     # cache lookup does not need to be protected by locking
     #  (but we could potentially prevent unnecessary computations)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.1 (unreleased)
 ------------------
 
+- Avoid unsolicited translations
+  (`#876 <https://github.com/zopefoundation/Zope/issues/876>`_)
 
 - Make "chameleon-zope context wrapping" more faithful.
   (`#873 <https://github.com/zopefoundation/Zope/pull/873/files>`_)
--- a/src/Products/PageTemplates/tests/testTranslation.py
+++ b/src/Products/PageTemplates/tests/testTranslation.py
@@ -14,7 +14,10 @@
 from unittest import TestCase
 
 from z3c.pt import pagetemplate
+from zope.component import provideAdapter
 from zope.component.testing import PlacelessSetup
+from zope.i18nmessageid import ZopeMessageFactory as _
+from zope.traversing.adapters import DefaultTraversable
 
 from ..PageTemplate import PageTemplate
 from .util import useChameleonEngine
@@ -41,6 +44,7 @@ class TranslationTests(PlacelessSetup, TestCase):
 
     def setUp(self):
         super(TranslationTests, self).setUp()
+        provideAdapter(DefaultTraversable, (None,))
         useChameleonEngine()
 
     def test_translation_body(self):
@@ -57,6 +61,11 @@ class TranslationTests(PlacelessSetup, TestCase):
         t = PageTemplate()
         t.write("""<p i18n:translate="" tal:replace="string:x">text</p>""")
         self.assertEqual(t(), "translated")
+
+    def test_translation_msgid(self):
+        t = PageTemplate()
+        t.write("""<p tal:replace="options/msgid">text</p>""")
+        self.assertEqual(t(msgid=_(u"x")), "translated")
 
     def test_no_translation_body(self):
         t = PageTemplate()
--- a/src/Products/PageTemplates/tests/testTranslation.py
+++ b/src/Products/PageTemplates/tests/testTranslation.py
@@ -1,0 +1,74 @@
+##############################################################################
+#
+# Copyright (c) 2020 Zope Foundation and Contributors.
+#
+# This software is subject to the provisions of the Zope Public License,
+# Version 2.1 (ZPL).  A copy of the ZPL should accompany this distribution.
+# THIS SOFTWARE IS PROVIDED "AS IS" AND ANY AND ALL EXPRESS OR IMPLIED
+# WARRANTIES ARE DISCLAIMED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF TITLE, MERCHANTABILITY, AGAINST INFRINGEMENT, AND FITNESS
+# FOR A PARTICULAR PURPOSE
+#
+##############################################################################
+
+from unittest import TestCase
+
+from z3c.pt import pagetemplate
+from zope.component.testing import PlacelessSetup
+
+from ..PageTemplate import PageTemplate
+from .util import useChameleonEngine
+
+
+def translate(*args, **kw):
+    return "translated"
+
+
+class TranslationLayer(object):
+    """set up our translation function."""
+    @classmethod
+    def setUp(cls):
+        cls._saved_translate = pagetemplate.fast_translate
+        pagetemplate.fast_translate = translate
+
+    @classmethod
+    def tearDown(cls):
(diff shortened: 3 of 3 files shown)


viktordick:
The code looks fine, but unfortunately it did not fix the issue. In our full environment, I am still getting an `Unauthorized` error (which is due to our translation function calling `strip()` on the message, which is not a string but still a `Token` I guess). I will have a more detailed look later.

d-maurer:
Viktor Dick wrote at 2020-7-21 22:23 -0700:
>The code looks fine, but unfortunately it did not fix the issue. In our full environment, I am still getting an `Unauthorized` error (which is due to our translation function calling `strip()` on the message, which is not a string but still a `Token` I guess). I will have a more detailed look later.

Are you using Python 2 or Python 3?
The PR does nothing for Python 2 (because for Python 2 all my
tests had shown correct behavior).

viktordick:
I am using Python 3. The main reason why we are switching to Zope 4 is to get rid of Python 2.

d-maurer:
Viktor Dick wrote at 2020-7-21 23:46 -0700:
>I am using Python 3.

Then the wrong translation can no longer come from expressions:
as you can see from the code, for Python 3, *expression*
is unconditionally converted to `str`.

In principle, `chameleon` may use `Token` also for text which
does not come from an expression but directly from the template,
e.g. from a construct such as `<p i18n:translate="">text</p>`.

d-maurer:
Dieter Maurer wrote at 2020-7-22 10:27 +0200:
>Viktor Dick wrote at 2020-7-21 23:46 -0700:
>>I am using Python 3.
>
>Then the wrong translation can no longer come from expressions:
>as you can see from the code, for Python 3, *expression*
>is unconditionally converted to `str`.
>
>In principle, `chameleon` may use `Token` also for text which
>does not come from an expression but directly from the template,
>e.g. from a construct such as `<p i18n:translate="">text</p>`.

I just tried: it does not use `Token` but `str` -- at least sometimes.

viktordick:
@d-maurer Since we now know that the translation function needs to take care of rejecting objects it is not prepared to translate when using chameleon anyway, I guess this PR is no longer necessary, is it?

Although the added tests are probably a good addition if we change the `translate` function to something like

    def translate(msgid, **kw):
        if isinstance(msgid, str):
            return 'translated'
        return msgid

d-maurer:
Viktor Dick wrote at 2020-7-22 12:15 -0700:
>@d-maurer Since we now know that the translation function needs to take care of rejecting objects it is not prepared to translate when using chameleon anyway, I guess this PR is no longer necessary, is it?

I would keep it.

Your observations have shown that sometimes an internal
`chameleon` type can arrive at a place where in fact
a string is expected. In your case, this caused unwanted translations
but in other cases it could cause other problems.
While the `chameleon` `Token` behaves quite like `str` (for Python 3)
or `unicode` (for Python 2), there are differences and
non-`chameleon` code (not necessarily related to translation)
may not be prepared for them.

The PR handles only the Python 3 case - but Python 2 support will
be dropped for Zope 5 anyway. Thus, I would keep the PR as is
and get active when problems for Python 2 are reported.

d-maurer:
Viktor Dick wrote at 2020-7-22 12:49 -0700:
>Thanks for the great help on this journey.

Thank you for the great preliminary detective work: problems which
apparently disappear after a restart are not easy to analyse.

src/Products/PageTemplates/engine.py
```diff
@@ -135,6 +136,13 @@ def _compile_zt_expr(type, expression, engine=None, econtext=None):
     """
     if engine is None:
         engine = econtext["__zt_engine__"]
+    # *expression* is a ``chameleon.tokenize.Token`` when
+    # the template is compiled but "text" when the template code
+    # comes from the ``chameleon`` cache.
+    # Under PY3, ``chameleon`` wrongly translates ``Token``;
```
malthe, line 142: I do think this comment is rather misleading given the nature of the integration into Chameleon's expression translation machinery. A rather substantial explanation of the issue has been provided to explain why a `Token` is provided in the initial compilation pass, and not during rendering.

d-maurer:
Malthe Borch wrote at 2020-7-22 13:27 -0700:
>@malthe commented on this pull request.
>> @@ -135,6 +136,13 @@ def _compile_zt_expr(type, expression, engine=None, econtext=None):
>     """
>     if engine is None:
>         engine = econtext["__zt_engine__"]
>+    # *expression* is a ``chameleon.tokenize.Token`` when
>+    # the template is compiled but "text" when the template code
>+    # comes from the ``chameleon`` cache.
>+    # Under PY3, ``chameleon`` wrongly translates ``Token``;
>
>I do think this comment is rather misleading given the nature of the integration into Chameleon's expression translation machinery. A rather substantial explanation of the issue has been provided to explain why a `Token` is provided in the initial compilation pass, and not during rendering.

The comment reflects external observations (`Token` on initial compilation,
`str` after restart) -- not necessarily what happens internally.
Likely, I am starting just now to understand what happens internally.

The "wrongly translates `Token`" is from before I read
the `chameleon` documentation which clearly describes
its extremely wide notion of "i18n message" - which among others includes
`Token`. Thus, the "wrongly" is wrong.

Feel free to improve the comment.

PR #879, opened 2020-07-23
improve unsolicited translation avoidance

### LINKED ISSUE #876: tal:content and tal:replace are sometimes treated as to be translated

d-maurer:
Until @malthe 's comment in "https://github.com/zopefoundation/Zope/pull/878#issuecomment-662724293", I did not really understand the cause of #876 (and as a consequence blamed `chameleon` wrongfully to be part of the problem).

This is an improved PR to fix #876 based on the insight gained by the comment above.

### DIFF
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -250,8 +250,8 @@ class MappedExpr(object):
         # the ``_compile_zt_expr`` below causes this to be cached
         # which can lead under Python 3 to unsolicited translations
         # (details "https://github.com/zopefoundation/Zope/issues/876")
-        # call ``_compile_zt_expr`` with a "plain" string under PY3
-        expr = str(expression) if six.PY3 else expression
+        # call ``_compile_zt_expr`` with a "plain" string
+        expr = str(expression) if six.PY3 else unicode(expression)
         self.expression = expression
         # compile to be able to report errors
         compiler_error = zt_engine.getCompilerError()
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -136,13 +136,6 @@ def _compile_zt_expr(type, expression, engine=None, econtext=None):
     """
     if engine is None:
         engine = econtext["__zt_engine__"]
-    # *expression* is a ``chameleon.tokenize.Token`` when
-    # the template is compiled but "text" when the template code
-    # comes from the ``chameleon`` cache.
-    # Under PY3, ``chameleon`` wrongly translates ``Token``;
-    # convert to ``str`` to avoid this
-    if six.PY3:
-        expression = str(expression)
     key = id(engine), type, expression
     # cache lookup does not need to be protected by locking
     #  (but we could potentially prevent unnecessary computations)
@@ -253,6 +246,12 @@ class MappedExpr(object):
     """map expression: ``zope.tales`` --> ``chameleon.tales``."""
     def __init__(self, type, expression, zt_engine):
         self.type = type
+        # at this place, *expression* is a `chameleon.tokenize.Token`
+        # the ``_compile_zt_expr`` below causes this to be cached
+        # which can lead under Python 3 to unsolicited translations
+        # (details "https://github.com/zopefoundation/Zope/issues/876")
+        if six.PY3:
+            expression = str(expression)
         self.expression = expression
         # compile to be able to report errors
         compiler_error = zt_engine.getCompilerError()
--- a/src/Products/PageTemplates/engine.py
+++ b/src/Products/PageTemplates/engine.py
@@ -246,11 +246,13 @@ class MappedExpr(object):
     """map expression: ``zope.tales`` --> ``chameleon.tales``."""
     def __init__(self, type, expression, zt_engine):
         self.type = type
-        # at this place, *expression* is a `chameleon.tokenize.Token`
-        # the ``_compile_zt_expr`` below causes this to be cached
+        # At this place, *expression* is a `chameleon.tokenize.Token`
+        # (a subtype of `str` for PY3 and of `unicode` for PY2).
+        # The ``_compile_zt_expr`` below causes this to be cached
         # which can lead under Python 3 to unsolicited translations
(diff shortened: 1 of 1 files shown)


malthe:
This looks correct to me.

dataflake:
I fixed the linting error by defining `unicode` if on Python 3, but is it really safe to just call `unicode` on the expression without specifying a character set and thus assuming it is always ASCII text?

d-maurer:
> I fixed the linting error by defining `unicode` if on Python 3

Thank you.

, but is it really safe to just call `unicode` on the expression without specifying a character set and thus assuming it is always ASCII text?

It should be safe. At this place `expression` should be a `chameleon.tokenize.Token`. For Python 3, this class derives from `str` and for Python 2 from `unicode`. This means that `expression` contains already (unicode) text (and never an encoded binary string) -- we just replace the subtype `Token` by the supertype (`str` or `unicode).

jugmac00:
> It should be safe. At this place `expression` should be a `chameleon.tokenize.Token`. For Python 3, this class derives from `str` and for Python 2 from `unicode`. This means that `expression` contains already (unicode) text (and never an encoded binary string) -- we just replace the subtype `Token` by the supertype (`str` or `unicode`).

Maybe it is a good idea to add a comment with this explanation?

d-maurer:
Jürgen Gmach wrote at 2020-7-23 02:02 -0700:
>> It should be safe. At this place `expression` should be a `chameleon.tokenize.Token`. For Python 3, this class derives from `str` and for Python 2 from `unicode`. This means that `expression` contains already (unicode) text (and never an encoded binary string) -- we just replace the subtype `Token` by the supertype (`str` or `unicode`).
>
>Maybe it is a good idea to add a comment with this explanation?

The comment before the "cast" now reads:
```
        # At this place, *expression* is a `chameleon.tokenize.Token`
        # (a subtype of `str` for PY3 and of `unicode` for PY2).
        # The ``_compile_zt_expr`` below causes this to be cached
        # which can lead under Python 3 to unsolicited translations
        # (details "https://github.com/zopefoundation/Zope/issues/876")
        # To avoid this, call ``_compile_zt_expr`` with
        # *expression* cast to the `Token` base type.
```

PR #882, opened 2020-07-30
Updated master versions based on 4.x versions.

mauritsvanrees:
I copied them over, including not yet merged zodbupdate 1.5, and edited to prevent unwanted changes.
I did this because I noticed that ZODB was still at 5.5.1 instead of 5.6.0.
All updated versions are newer versions.
(The edits to constraints and full requirements are the automatic changes after running bin/buildout.)


dataflake:
Not so fast. I will not accept an update of the ZODB version pin until there is a new ZEO release that is actually compatible with ZODB 5.6.

dataflake:
> Not so fast. I will not accept an update of the ZODB version pin until there is a new ZEO release that is actually compatible with ZODB 5.6.

I misunderstood the patch, sorry. Zope 4.x has that ZODB 5.6 version pin already. This is good for merging.

The incompatibilities with the current ZEO release were discovered after making that version pin on the `4.x` branch, so I will hold off on making a Zope 4 release until ZEO 5.2.2 is released. I cannot do the release myself because only @tseaver and @jimfulton are PyPI maintainers for it, I have asked Tres to do it.

PR #885, opened 2020-08-21
Better session documentation and link

dwt:
Update seasoning documentation hint and link the chapter on sessions.

Not sure this is the best way to link to the zope book - guidance on the right way to link this would be very welcome.


dataflake:
I cleaned up the formatting and expanded the explanation a little.

PR #889, opened 2020-09-17
Zope 4: use DocumentTemplate branch 3.x.

mauritsvanrees:
The master branch targets Zope 5.
It fails with python 2.7:

```
File "/Users/maurits/community/plone-coredev/5.2/src/DocumentTemplate/src/DocumentTemplate/DT_Var.py", line 156, in <module>
    import urllib.parse
ImportError: No module named parse
```


dataflake (approved):
Good catch

PR #893, opened 2020-09-28
Fix export of files with non-latin-1 compatible names

### LINKED ISSUE #890: Export / download of files with non-latin-1 compatible names is broken
**Reproduction**
- upload a file via ZMI with the name "ıqrm.png" (please note, the first letter is a turkish i)
- try to export it
- waitress error

```
2020-09-24 14:57:24,528 ERROR   [waitress:358][waitress-0] Exception while serving /bliss/db/manage_exportObject
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/channel.py", line 350, in service
    task.service()
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/task.py", line 171, in service
    self.execute()
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/task.py", line 479, in execute
    self.write(chunk)
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/task.py", line 311, in write
    rh = self.build_response_header()
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/task.py", line 284, in build_response_header
    return tobytes(res)
  File "/home/jugmac00/Projects/bliss_deployment/work/_/home/jugmac00/.batou-shared-eggs/waitress-1.4.4-py3.7.egg/waitress/compat.py", line 69, in tobytes
    return bytes(s, "latin-1")
UnicodeEncodeError: 'latin-1' codec can't encode character '\u0131' in position 54: ordinal not in range(256)
```

Actually, this was not my use case. I have a custom download (and a file view) in my Zope app, and noticed that the following code is broken for non latin-1 names:
```
    def download(self, REQUEST):
        """
        download the attachment
         @REQUEST : request
        """
        with open(self.completeFileName(), "rb") as f:
            ret = f.read()
        REQUEST.RESPONSE.setHeader('content-type', self.getMimeType())
        REQUEST.RESPONSE.setHeader('content-length', str(len(ret)))
        REQUEST.RESPONSE.setHeader('content-disposition', 'attachment; filename="%s"' % self.printFileName())
        return ret
```

I then grepped the Zope source for a possible solution and found above bug.

**possible solutions**
- `zope.file` offers some solution - but when I understand that correctly, it changes the file name by double en- and decoding (which seems wrong to me, but possibly I do not understand the trick)
- `cherrypy` and afaik `Django` handle this differently

**zope.file**
https://github.com/zopefoundation/zope.file/blob/b952af17c23f7702b95a93b28b6458b38560874b/src/zope/file/download.py#L65-L70

**cherrypy**
https://github.com/cherrypy/cherrypy/pull/1851/files#diff-1b04e50f6724343e4321295525e2bd58R32

**django-sendfile**
https://github.com/johnsensible/django-sendfile/blob/7a000e0e7ae1732bcea222c7872bfda6a97c495c/sendfile/__init__.py#L69-L86

In my Zope app I implemented `cherrypy`s solution and it works like a charm.

As there is also a bug in Zope's `manage_exportObject` I'd like to have a function in Zope itself, given a filename and the type (inline or attachment), builds a correct header.

I'd volunteer to implement a solution during next Monday's Zope sprint.

@d-maurer @icemac - Maybe there is something I overlook? I hardly can't imagine nobody ever wanted to offer a file download for non-latin-1 names before?

P.S.: Maybe also both the comment of the `setHeader` method and the method itself should be updated, ie either warn or prohibit to set non-latin-1 values.

jugmac00:
With Zope 4 unicode characters for ids are allowed.

The `zexp` export was not updated yet.

This has been fixed now.

closes #890

modified:   CHANGES.rst
modified:   src/OFS/ObjectManager.py
modified:   src/Testing/ZopeTestCase/testZODBCompat.py
new file:   src/ZPublisher/http_header_utils.py
new file:   src/ZPublisher/tests/test_http_header_utils.py


jugmac00:
I copied the `make_request_response`  from the WebDAV module. I think this is too good to be hidden and also spread over several test files.

I'd like to have more readily available, maybe in the `Testing` package?

src/ZPublisher/tests/test_http_header_utils.py
```diff
@@ -0,0 +1,37 @@
+##############################################################################
+#
+# Copyright (c) 2017 Zope Foundation and Contributors.
```
jugmac00, line 3: This would be a good thing to add to the newly created `contributing` file. What do you have to do when you create a new file/ module?

dataflake:
I am very unhappy that this PR copies third party files wholesale into Zope. I believe this is not legal if you look at the contributor agreement. You are required to assign rights to the code to the Zope Foundation, and you cannot do that since you did not write this code originally.

src/ZPublisher/tests/test_http_header_utils.py
icemac:
```suggestion
# Copyright (c) 2020 Zope Foundation and Contributors.
```

dataflake:
@jugmac00 Please do the following, as copying third party code into our code base is illegal under the contributor agreement (regardless of its license):

- remove the two files you copied from CherryPy
- write your own function and add that to an existing module where it makes sense, you're free to use the CherryPy code as inspiration

src/ZPublisher/http_header_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    # See: https://en.wikipedia.org/wiki/Unicode_equivalence.
+    ascii_name = (
+        unicodedata.normalize('NFKC', file_name).
+        encode('ascii', errors='ignore').decode()
+    )
+    header = '{}; filename="{}"'.format(disposition, ascii_name)
+    if ascii_name != file_name:
+        quoted_name = urllib.parse.quote(file_name)
+        header += '; filename*=UTF-8\'\'{}'.format(quoted_name)
+    return header
```
icemac, line 61: I agree with @dataflake. The first section of the contributor agreement requires you to only check-in code you created yourself:

> When you write or make changes to software programs, you often automatically obtain a copyright in what you wrote. The way to assign to the Plone Foundation the copyright in software that you wrote or changes that you made to Plone Foundation software is to send us a signed Assignment Agreement. This makes the Plone Foundation the legal copyright holder so that we can register the copyright in the software that you wrote or changes that you made. It is assumed that you wrote the software or changes yourself; if other people wrote or changed parts, we will likely need Assignment Agreements from them as well.

Sorry for my wrong advise beforehand.

icemac (changes requested):
Sorry for my wrong advise but vendoring code is not allowed by the contributor agreement.

jugmac00:
Thanks for the review!

As I knew there would be objections against the vendoring, I talked to @icemac beforehands and he was ok with the changes, especially with vendoring the code. He explicitly stated that "this is legal as both Zope and CherryPy use BSD based licenses" (update: I just saw @icemac's comment :+1: )

Shift happens :-)

There are several examples of vendored code in Zope, e.g. https://github.com/zopefoundation/Zope/commit/72ae9a1bed26db928e7f74fa063c8de15a755544 or the vendoring of `font-awesome`.

Why would this one be special?

While I think the proposed solution is the nicest one, I certainly thought about a custom implementation.

I would suggest to try to encode the header using `latin-1` and in case of an `UnicodeEncodeError` exception apply the necessary pattern (adding `;filename* ....`).

dataflake:
I am sure there are other places where code was added illegally. That doesn't excuse this case, though. The only exception can be made if the original code author also signs a document that assigns rights to the code to the ZF/PF.

jugmac00:
Before I update the pull request.. are you ok with a solution as following (pseudo code):
```
try:
    file_name.encode("latin-1")
except UnicodeEncodeError:
   # create special header with the filename + filename* syntax
   return special header
else:
   return simple header
```

dataflake:
You're now requiring me to spend time to get into the technical details of this issue, which I don't have. Why don't you just use the CherryPy code as inspiration and do it like they did?

d-maurer:
Jürgen Gmach wrote at 2020-9-28 04:56 -0700:
> ...
>I would suggest to try to encode the header using `latin-1` and in case of an `UnicodeEncodeError` exception apply the necessary pattern (adding `;filename* ....`).

The latest HTTP 1.1 specifiction (RFC 7230 (-7237)) no longer defines
`latin-1` as the default charset used by HTTP header fields
(though all current browsers will likely still support this
as it was the default charset defined by the previous HTTP 1.1 spec).
Instead it advocates ASCII only headers.

The current way to have non ASCII characters in HTTP headers
seems to be specified by RFC 8187 -- it is this specification
the `cherrypy` code follows.

I would follow these latest specifications:
  encode using "USASCII"; if this succeeds, define "filename"

  if this fails, define `filename*` according to RFC 8187
  and define `filename` with `.encode('USASCII', error=epol)`
  (with *epol* either `'ignore'` or `'replace'`).
  The definition of `filename` is for browsers that do not
  yet support RFC 8187 (and its predecessors).

Of course, you could always follow the `if this fails` branch
(the only effect is that you would get a not strictly necessary
`filename*` parameter).


As a side note: it would be good if the character encoding
handling for HTTP headers were at a central place (`HTTPResponse`
seems a good place from my point of view).
I do not think it a good idea to have those issues handled at
various places in "application" (viewed from `ZPublisher`) code.

jugmac00:
Thanks, @d-maurer for your feedback!

I just pushed the proposed changes (ie use `us-ascii` instead of `latin-1`).

I am not sure about your last paragraph:

> As a side note: it would be good if the character encoding
handling for HTTP headers were at a central place (`HTTPResponse`
seems a good place from my point of view).
I do not think it a good idea to have those issues handled at
various places in "application" (viewed from `ZPublisher`) code.

When thinking where to put the new `make_content_disposition` function, I first thought of `HTTPResponse`, but this module is already pretty big. Then I saw `utils.py` - which is a non-descriptive name at all. I then decided to put the function in `http_headers_utils.py` - just besides the `HTTPResponse` module.

Do you want me to move the `make_content_disposition` function into the `HTTPResponse` module? If so, I just move it. No problem.

And thanks again for your helpful feedback!

dataflake:
Could you please move the function into a better place, like the `HTTPResponse.py` module that already contains some header utility code at the top? There's no reason to add new files just for this. Thanks!

d-maurer:
Jürgen Gmach wrote at 2020-9-28 23:36 -0700:
>Thanks, @d-maurer for your feedback!
>
>I just pushed the proposed changes (ie use `us-ascii` instead of `latin-1`).
>
>I am not sure about your last paragraph:
>
>> As a side note: it would be good if the character encoding
>handling for HTTP headers were at a central place (`HTTPResponse`
>seems a good place from my point of view).
>I do not think it a good idea to have those issues handled at
>various places in "application" (viewed from `ZPublisher`) code.
> ...
>Do you want me to move the `make_content_disposition` function into the `HTTPResponse` module? If so, I just move it. No problem.

I rarely "want" people do something -- view it more as a suggestion.

My suggestion actually is much more general:
the problem you have observed (non-ASCII in HTTP response headers)
can affect other headers and not only `Content-Disposition`.
This may indicate that it should be handled generally in
the header processing of `HTTPResponse` rather than with a special
function (you need to know and apply at the relevant places).
Thus, my suggestion actually was: do not introduce a new function
to create a `Content-Disposition` header but handle the case
automatically during header processing in `HTTPResponse`.

Unfortunately, RFC 8187 (and especially the `*` name suffix convention)
was not designed as a general purpose approach. Instead, its use
requires explicit specification. This makes things more complicated.
But, it should be safe to add a `*` suffixed parameter when `HTTPResponse`
processes a non-latin-1 response header
per default and allow a header registry to override the default.

`Content-Disposition` is specified by RFC 6266 - and it explicitely
allows "latin-1" values. Thus we have two contradicting standards:
RFC 7230 discourages the use of non ASCII values;
RFC 6266 explicitely allows "latin-1" values -- a fact, I was not
aware of yesterday. Thus, potentially, your original use of `latin-1`
may have been better than my proposed use if `US-ASCII`.

d-maurer:
Jürgen Gmach wrote at 2020-9-28 23:36 -0700:
> ...

Note: we may have a symmetric problem on import/upload. When I remember right,
the id (if not specified) of an uploaded `File` is determined from
the `filename` parameter of the `Content-Dispostion` header.
This may go wrong when the file name contains non-ASCII/non-latin-1
characters.

jugmac00:
@d-maurer When facing this problem the first time, I had a look at e.g. `HTTPResponse.setHeader` and similar, but I saw no straightforward to handle combined header data there.

I understand my proposed fix as 
- it is getting better than before
- as a Zope app developer I do not have to duplicate the header creation and unicode handling, but I can do a `from ZPublisher.HTTPResponse import make_content_disposition` and use it in my app

I do not see anybody having both the time and the expertise to implement your (obviously better) solution.

P.S.: Meanwhile, I have vendored `CherryPy`s solution into my app (no license issues there) - so I am not in a hurry, but I certainly would like to see a solution from the framework side.

d-maurer:
Jürgen Gmach wrote at 2020-9-29 00:59 -0700:
>@d-maurer When facing this problem the first time, I had a look at e.g. `HTTPResponse.setHeader` and similar, but I saw no straightforward to handle combined header data there.

As `HTTPResponse` has several methods for header handling it
is likely best to extend where the headers are serialized
(which likely happens in the `__str__` method).

I assume that you associate "combined header data" with
the various "parameter"s of `Content-Disposition` (and some other headers).
Python's `email` package contains functions to parse those composite headers
(used e.g. for the `Content-Type` header parsing).


> I do not see anybody having both the time and the expertise to implement your (obviously better) solution.

As I wrote, it was only a suggestion, not something I expect from you.

dataflake:
@jugmac00 Thanks for reimplementing and moving the function. While Dieter's suggestion about implementing this header handling in a more central place and using it on other headers is a good idea I suggest to make that a different issue if and when someone has time for it. I'll merge this one.

PR #897, opened 2020-10-02
4.x: Update to zope.interface 5.1.1 which fixes a memory leak.

icemac:


dataflake (approved):

dataflake:
@icemac I have finished the pin to 5.1.2 for `master` and `4.x` and published it on `gh-pages`

icemac:
@dataflake Thank you, I often forget that there are two more files with version pins.

PR #899, opened 2020-10-02
Restore the ZMI `Debug Information` control panel page

### LINKED ISSUE #898: Restore debug support via "Control_Panel"?
## Question/Enhancement
I am currently investigating a memory leak in Zope 2. Its "Debug information" in `Control_Panel` is valuable help (even though I had to replace `App.ApplicationManager.DebugManager.refcount` to avoid `six` problems). This functionality has been removed from Zope 4.

Should this functionality (as well as cache analysis and open connection overview) be restored
in stock Zope?

The alternative is that I make corresponding analysis tools available via a package `dm.zope.debug` -- usable for anyone who find them useful.

Zope 4.x, Zope 5

dataflake:
Fixes #898 

This PR adds a "Debug Information" page to the control panel. This Debug Information page restores those parts of the old Zope 2  DebugInfo page that are not already presented on other control panel tabs.

The trickiest part was finding a place to add the request debugging information for running requests that shows up at the bottom of the page where the database connection information is. I have added it inside the WSGI publisher code. This used to be done by the `App.ZApplication.ZApplicationWrapper.__bobo_traverse__`, but that method is only called by the old ZServer publisher.

The ZMI bits could use some prettification, but functionally it's all there.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.2 (unreleased)
 ------------------
 
+- Restore the ZMI `Debug Information` control panel page
+  (`#898 <https://github.com/zopefoundation/Zope/issues/898>`_)
+
 - Update dependencies to the latest releases that still support Python 2.
 
 - Update to ``zope.interface > 5.1.0`` to fix a memory leak.
--- a/src/App/ApplicationManager.py
+++ b/src/App/ApplicationManager.py
@@ -14,18 +14,22 @@
 import os
 import sys
 import time
+import types
 
+from six.moves._thread import get_ident
 from six.moves.urllib import parse
 
 from AccessControl.class_init import InitializeClass
 from AccessControl.requestmethod import requestmethod
 from Acquisition import Implicit
+from App.CacheManager import CacheManager
 from App.config import getConfiguration
 from App.DavLockManager import DavLockManager
 from App.Management import Tabs
 from App.special_dtml import DTMLFile
 from App.Undo import UndoSupport
 from App.version_txt import version_txt
+from DateTime.DateTime import DateTime
 from OFS.Traversable import Traversable
 from Persistence import Persistent
 from Products.PageTemplates.PageTemplateFile import PageTemplateFile
@@ -59,6 +63,8 @@ class DatabaseChooser(Tabs, Traversable, Implicit):
         {'label': 'Control Panel', 'action': '../manage_main'},
         {'label': 'Databases', 'action': 'manage_main'},
         {'label': 'Configuration', 'action': '../Configuration/manage_main'},
+        {'label': 'DAV Locks', 'action': '../DAVLocks/manage_main'},
+        {'label': 'Debug Information', 'action': '../DebugInfo/manage_main'},
     )
     MANAGE_TABS_NO_BANNER = True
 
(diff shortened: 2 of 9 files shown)


d-maurer (approved):
Good!
It has been a very good idea to move the `sys.path` information out of "Debug Information".

I have not changed the `t` parameter of the method `refcount` (even though this might be necessary to analyze memory leaks in "modern" Zope applications). I would like to await concrete experience from my current memory leak analysis and its trial definition `(ExtensionClass, type, ClassType)`: it could be possible that the refcount information for types implemented in C does not give a good estimate for the number of type instances.

d-maurer:
> ...
> it could be possible that the refcount information for types implemented in C does not give a good estimate for the number of type instances.

This is indeed the case (both Python 2 and Python 3):
```pycon
>>> from sys import getrefcount as rc
>>> l=[str(x) for x in range(1000)]
>>> rc(str)
73
```
I applies also for classes derived from elementary Python data types:
```pycon
>>> class C(dict): pass
... 
>>> l=[C() for x in range(1000)]
>>> rc(l)
2
```
For old style classes (Python 2) and "normal" classes, the class refcount gives a good estimation for the number of instantiated objects:
```pycon
>>> class C(object): pass
... 
>>> l=[C() for x in range(1000)]
>>> rc(C)
1005
>>> class O: pass
... 
>>> l=[O() for x in range(1000)]
>>> rc(O)
1002
```

icemac:
I would have liked to restart the failing job on TravisCI as it seems to be a timing issue, but I am currently (no longer?) be allowed to do so.

icemac:
I've never actually used this feature so I am removing me from the reviewers list.

dataflake:
> I would have liked to restart the failing job on TravisCI as it seems to be a timing issue, but I am currently (no longer?) be allowed to do so.

Nothing changed at Travis as far as I know. Did you log into Travis from that job status page? You won't see the buttons to restart a job if you're not logged in.

icemac:
@dataflake wrote:
> Nothing changed at Travis as far as I know. Did you log into Travis from that job status page? You won't see the buttons to restart a job if you're not logged in.

I am logged in at https://travis-ci.org via Github. In Settings I see the Zope project but if I try to go to its settings I get „Your permissions are insufficient to access this repository's settings“. Re-Login did not help. Maybe they screwed up their database a bit.

Is there a reason why we are still at travis-ci.org instead of travis-ci.com?

d-maurer:
I have now practical experience with `App.ApplicationManager.DebugManager.refcount` collecting not only refcount information for `ExtensionClass` instances but also for normal classes (i.e. instances of `six.class_types`). That extended information is useful for memory usage analysis if one is aware that the collected refcounts are not for all classes a good estimate for the number of class instances (this applies in particular for elementary Python data types and their derived classes). As a consequence, I have added `six.class_types` to the refcount `t` parameter.

dataflake:
> I am logged in at https://travis-ci.org via Github. In Settings I see the Zope project but if I try to go to its settings I get „Your permissions are insufficient to access this repository's settings“. Re-Login did not help. Maybe they screwed up their database a bit.

I can log in and view those settings fine. I then looked if I am member of any GitHub team under zopefoundation/Zope that you're not and added you to "Contributors" - maybe that makes a difference.

> Is there a reason why we are still at travis-ci.org instead of travis-ci.com?

Not that I know. I didn't set this up. Feel free to move it.

icemac:
> I can log in and view those settings fine. I then looked if I am member of any GitHub team under zopefoundation/Zope that you're not and added you to "Contributors" - maybe that makes a difference.

Now I again have the ability to restart Travis jobs. Thank you, although I am not sure if adding me to Contributors caused the change.

dataflake:
Neither am I 🙄

PR #905, opened 2020-10-06
Http response header encoding

d-maurer:
HTTP requires that response (and request) header fields use `ISO-8859-1` (or a subset) as charset for their values. `waitress` builds on this and raises an exception when a field value cannot be represented in this charset.

This PR provides support for RFC 2047 ("encode_words") and RFC 2231 ("encode params"; in its HTTP specialization by RFC 5987/8187) for response header field values not covered by `ISO-8859-1` via new `ZPublisher.HTTPResponse.header_encoding_registry` (of class `HeaderEncodingRegistry`). `HTTPResponse.listHeaders` consults this registry in its `listHeaders` method to determine how header field values outside `ISO-8859-1` should be handled. By default, it uses `encode_params` as encoding for `Content-Type` and `Content-Disposition` and does not encode other header fields (which may result in exceptions in the web server framework). An application can register encodings for additional fields.

`encode_params` is usable for structured headers to support non-ISO-8859-1 values in parameter values. Currently, it cannot be for used headers that support multiple (comma separated) parameter sets.

`encode_words` can be used for unstructured headers.

The cookie header fields are under complete control of `HTTPResponse`'s cookie handling and not affected by `header_encoding_registry`.

### DIFF
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -800,7 +800,7 @@ class HTTPResponse(HTTPBaseResponse):
     def _error_html(self, title, body):
         return ("""<!DOCTYPE html><html>
   <head><title>Site Error</title><meta charset="utf-8" /></head>
-  <body bgcolor='#FFFFFF'>
+  <body bgcolor="#FFFFFF">
   <h2>Site Error</h2>
   <p>An error was encountered while publishing this resource.
   </p>
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -18,6 +18,9 @@ import struct
 import sys
 import time
 import zlib
+from email.header import Header
+from email.message import _parseparam
+from email.utils import encode_rfc2231
 from io import BytesIO
 
 from six import PY2
@@ -25,6 +28,7 @@ from six import PY3
 from six import binary_type
 from six import class_types
 from six import reraise
+from six import string_types
 from six import text_type
 from six.moves.urllib.parse import quote
 from six.moves.urllib.parse import urlparse
@@ -109,7 +113,9 @@ _CRLF = re.compile(r'[\r\n]')
 
 
 def _scrubHeader(name, value):
-    return ''.join(_CRLF.split(str(name))), ''.join(_CRLF.split(str(value)))
+    if not isinstance(value, string_types):
+        value = str(value)
+    return ''.join(_CRLF.split(str(name))), ''.join(_CRLF.split(value))
 
 
 _NOW = None  # overwrite for testing
@@ -709,20 +715,24 @@ class HTTPBaseResponse(BaseResponse):
         """ Return a list of (key, value) pairs for our headers.
 
         o Do appropriate case normalization.
+
+        o Encode header values via `header_encoding_registry`
         """
 
         result = [
             ('X-Powered-By', 'Zope (www.zope.org), Python (www.python.org)')
         ]
 
+        encode = header_encoding_registry.encode
         for key, value in self.headers.items():
             if key.lower() == key:
                 # only change non-literal header names
                 key = '-'.join([x.capitalize() for x in key.split('-')])
-            result.append((key, value))
+            result.append((key, encode(key, value)))
 
         result.extend(self._cookie_list())
-        result.extend(self.accumulated_headers)
+        for key, value in self.accumulated_headers:
+            result.append((key, encode(key, value)))
         return result
 
     def _unauthorized(self):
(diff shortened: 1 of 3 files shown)


src/ZPublisher/HTTPResponse.py
```diff
@@ -790,7 +800,7 @@ def _traceback(self, t, v, tb, as_html=1):
     def _error_html(self, title, body):
         return ("""<!DOCTYPE html><html>
   <head><title>Site Error</title><meta charset="utf-8" /></head>
-  <body bgcolor="#FFFFFF">
+  <body bgcolor='#FFFFFF'>
```
dataflake, line 803: Why did you change the quotes? That should not be necessary. For consistency's sake please leave the double quotes.

src/ZPublisher/HTTPResponse.py
```diff
@@ -1111,3 +1121,129 @@ def __bytes__(self):
...
+    This encodes parameters as specified by RFC 5987 using
+    fixed `UTF-8` encoding (as required by RFC 8187).
+    However, all parameters with non latin-1 values are
+    automatically transformed and a `*` suffixed parameter is added
+    (RFC 8187 allows this only for parameters explicitly specified
+    to have this behavior).
+
+    Many HTTP headers use `,` separated lists. For simplicity,
+    such headers are not supported (we would need to recognize
+    `,` inside quoted strings as special).1G
```
dataflake, line 1221: The `1G` at the end of the line is a typo, right?

dataflake (changes requested):
Just a couple small changes.

src/ZPublisher/HTTPResponse.py
d-maurer: I did this because the double quotes confused the XEmacs syntax highlighting (an XEmacs bug) and rendered the code unreadable for me (following code was represented as string and vice versa).

The change is **inside** a string (destined as HTML fragment). Is quote consistency in the generated HTML really a major concern for you?

dataflake: I have a string dislike when it comes to single quotes in HTML tags. But it looks like you have a good reason, so never mind. Try a better editor 😉

d-maurer: I have reverted the corresponding commit (hoping that I will not need to work a lot in the code).

src/ZPublisher/HTTPResponse.py
d-maurer: Fixed

dataflake (approved):

dataflake:
Could you do the forward-port to `master` as well?

d-maurer:
Jens Vagelpohl wrote at 2020-10-6 02:43 -0700:
>Could you do the forward-port to `master` as well?

Yes. May take a few days.

dataflake:
I have done the port to master, Michael wants to move forard with a Zope 5 release. See 2adcb9513acd54b0436f1a36c900ca11f8c69207

mauritsvanrees:
Yesterday I added Zope to the checkouts in the Plone core development buildout. It went fine for Python 3, but with Python 2 there are [test failures](https://jenkins.plone.org/job/plone-5.2-python-2.7/1162/#showFailuresLink) in `plone.restapi`. I tracked it down to the changes from this PR.
It might be that the tests in `plone.restapi` need to be changed.
Example failure is in [this line](https://github.com/plone/plone.restapi/blob/6.15.0/src/plone/restapi/tests/test_services_users.py#L248) where we uninformatively get status 500 instead 201. When I check, the response contains this internal server error:

```
(Pdb) print(response.text)
Internal Server Error

  File "/Users/maurits/shared-eggs/cp27m/waitress-1.4.4-py2.7.egg/waitress/channel.py", line 350, in service
    task.service()
  File "/Users/maurits/shared-eggs/cp27m/waitress-1.4.4-py2.7.egg/waitress/task.py", line 171, in service
    self.execute()
  File "/Users/maurits/shared-eggs/cp27m/waitress-1.4.4-py2.7.egg/waitress/task.py", line 441, in execute
    app_iter = self.channel.server.application(environ, start_response)
  File "/Users/maurits/shared-eggs/cp27m/WebTest-2.0.35-py2.7.egg/webtest/http.py", line 83, in wrapper
    return self.test_app(environ, start_response)
  File "/Users/maurits/community/plone-coredev/5.2/src/Zope/src/ZPublisher/httpexceptions.py", line 30, in __call__
    return self.application(environ, start_response)
  File "/Users/maurits/community/plone-coredev/5.2/src/Zope/src/ZPublisher/WSGIPublisher.py", line 378, in publish_module
    start_response(status, headers)
  File "/Users/maurits/shared-eggs/cp27m/waitress-1.4.4-py2.7.egg/waitress/task.py", line 414, in start_response
    "Header value %r is not a string in %r" % (v, (k, v))
AssertionError: Header value u'http://localhost:65521/plone/@users/howard' is not a string in ('Location', u'http://localhost:65521/plone/@users/howard')
```

So the Location header is unicode instead of string. Looking at all headers:

```
(Pdb) pp headers
[('Server', 'waitress'),
 ('Date', 'Thu, 12 Nov 2020 10:28:57 GMT'),
 ('X-Powered-By', 'Zope (www.zope.org), Python (www.python.org)'),
 ('Content-Length', '290'),
 ('Content-Language', 'en'),
 ('Expires', 'Sat, 1 Jan 2000 00:00:00 GMT'),
 ('Location', u'http://localhost:49490/plone/@users/howard'),
 ('X-Frame-Options', 'SAMEORIGIN'),
 ('Content-Type', 'application/json')]
```

@d-maurer does this tell you anything on how to go about fixing this?

BTW sorry @dataflake: I should have checked that all core Plone tests were fine with a Zope checkout, before asking for a new Zope release.

d-maurer:
Maurits van Rees wrote at 2020-11-12 02:33 -0800:
>Yesterday I added Zope to the checkouts in the Plone core development buildout. It went fine for Python 3, but with Python 2 there are [test failures](https://jenkins.plone.org/job/plone-5.2-python-2.7/1162/#showFailuresLink) in `plone.restapi`. I tracked it down to the changes from this PR.
>It might be that the tests in `plone.restapi` need to be changed.
>Example failure is in [this line](https://github.com/plone/plone.restapi/blob/6.15.0/src/plone/restapi/tests/test_services_users.py#L248) where we uninformatively get status 500 instead 201. When I check, the response contains this internal server error:
>
> ...
>    start_response(status, headers)
>  File "/Users/maurits/shared-eggs/cp27m/waitress-1.4.4-py2.7.egg/waitress/task.py", line 414, in start_response
>    "Header value %r is not a string in %r" % (v, (k, v))
>AssertionError: Header value u'http://localhost:65521/plone/@users/howard' is not a string in ('Location', u'http://localhost:65521/plone/@users/howard')
>```
>
>So the Location header is unicode instead of string. Looking at all headers:
>
>```
>(Pdb) pp headers
>[('Server', 'waitress'),
> ('Date', 'Thu, 12 Nov 2020 10:28:57 GMT'),
> ('X-Powered-By', 'Zope (www.zope.org), Python (www.python.org)'),
> ('Content-Length', '290'),
> ('Content-Language', 'en'),
> ('Expires', 'Sat, 1 Jan 2000 00:00:00 GMT'),
> ('Location', u'http://localhost:49490/plone/@users/howard'),
> ('X-Frame-Options', 'SAMEORIGIN'),
> ('Content-Type', 'application/json')]
>```
>
>@d-maurer does this tell you anything on how to go about fixing this?

I will look into this tomorrow.

In principle, the PR should have affected only header values
not encodable in ISO-8859-1 and only `Content-Type` and
`Content-Disposition` but not `Location`.
Should it really be responsible, this would be a surprise.

mauritsvanrees:
I see what causes it. It is the [`_scrubHeader`](https://github.com/zopefoundation/Zope/pull/905/files#diff-91938ac6f7bfff26f990573e8b4132dabefb6e5599c162707f4836800f95f5eaR115-R118) part of this PR. Originally, this always called `str(value)`, but now it only does this when `value` is not in `six.string_types`. So if the value is Python 2 unicode, after this PR you get unicode in the header, which then fails in `waitress`.

The unicode comes from [these lines in `plone.restapi`](https://github.com/plone/plone.restapi/blob/6.15.0/src/plone/restapi/services/users/add.py#L218-L220) where the Location gets set to `portal.absolute_url() + "/@users/" + username` and `username` is unicode.

So the solution seems to be to change `plone.restapi` to ensure it sets only string headers.

It *could* be that more third party packages are hit by this. If so, the code could be improved: encode remaining unicode headers with utf-8. But for now, I will look into changing `plone.restapi`.
Anyway, previously if the value was unicode with non-ascii characters, you would have gotten a `UnicodeDecodeError`, which no longer happens.

d-maurer:
Maurits van Rees wrote at 2020-11-12 04:15 -0800:
>I see what causes it. It is the [`_scrubHeader`](https://github.com/zopefoundation/Zope/pull/905/files#diff-91938ac6f7bfff26f990573e8b4132dabefb6e5599c162707f4836800f95f5eaR115-R118) part of this PR. Originally, this always called `str(value)`, but now it only does this when `value` is not in `six.string_types`. So if the value is Python 2 unicode, after this PR you get unicode in the header, which then fails in `waitress`.

Thank you for the analysis.

> ...
>It *could* be that more third party packages are hit by this. If so, the code could be improved: encode remaining unicode headers with utf-8. But for now, I will look into changing `plone.restapi`.

I will look for a change in `Zope`.
However, I will not use `utf-8`: HTTP/1.1 specifies that headers
use ISO-8859-1 as encoding.

PR #907, opened 2020-10-07
Replace (in "OFS") deprecated direct "id" access by "getId" calls

### LINKED ISSUE #903: (Export/)Import broken for "OFS.SimpleItem.Item_w__name__" instances
## BUG/PROBLEM REPORT
If an instance of `OFS.SimpleItem.Item_w__name__` (e.g. an `OFS.Image.File` instance) is exported, a subsequent import fails with `BadRequest: ('Empty or invalid id specified', '')`.

The reason: for `Item_w__name__` instances, `id` must not be used (it is always `''`); for those instances `getId()` must be used to access the object's id.
`OFS.ObjectManager.ObjectManager._importObjectFromFile` violates this restriction.

current Zope 4.x (likely Zope 5.x as well)

d-maurer:
Direct `id` access is deprecated (for `OFS.SimpleItem.Item` instances). However, at some places (e.g. in export/import) it was still used in `OFS`. This has caused #903 .

The PR replaces most occurrences of direct `id` access by calls to `getId` (provided the object has an unacquired `getId`). The direct `id` accesses in `OFS.Traversable.Traversable.getPhysicalPath` are retained (this might give wrong results for (likely rare) objects with both `id` and `__name__` with diverging values).

Fixes #903.


dataflake (approved):

icemac:
Has there still something to be done before this PR can be merged?

PR #927, opened 2020-11-02
Exclude ${} from use in TALES path expressions

### LINKED ISSUE #925: Template problem with nested curly braces and chameleon expressions.
This is with Zope 4.5.1.
See [community.plone.org](https://community.plone.org/t/plone-5-2-1-5-2-2-page-template-issues/12793/16?u=mauritsvanrees) where Marcel Liebischer reports this. I can confirm it, and see the same in the current core Plone 5.2 development buildout, which I think has some newer versions for template related packages.

Let me copy the original report here:

----------

I ran into another problem with page templates and Plone 5.2.2.

It seems there are problems with nested curly braces and chameleon expressions. This could easily happen if you try to build a data-attribute with some JSON data in a page template.

Steps to reproduce:

    <span data-test='{"foo": "${view/somemethod}"}'/>

-> does not work. Throws "zope.location.interfaces.LocationError .. (view object, 'somemethod)'"

    <span data-test='"foo": "${view/somemethod}"'/>

-> works fine (but it's not valid JSON)

    <span data-test='{"foo": "${python: view.somemethod()}"'/>

-> works fine

The issues does not exist in Plone 5.2.0.

d-maurer:
Fixes #925 

The `chameleon` interpolation has special characters `${}`. When those characters are usable in TALES path expressions, `chameleon`'s interpolation heuristics (use the longest possible interpolation string the compilation of which does not cause a syntax error) has a high risk to break. One such example is #925 

This PR (essentially) forbids the use of `${}` in path expressions to reduce the failure risk for the interpolation heuristics (actually, it forbids those characters only in `SubPathExpr`s with a standard path; they can still be used in other types, e.g. a trailing `python` or `string` expression).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,10 +10,16 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.2 (unreleased)
 ------------------
 
+Backward incompatible changes
++++++++++++++++++++++++++++++
+
 - Exclude characters special for ``chameleon``'s interpolation syntax
-  (i.e. ``${}``) from use in path expressions to reduce the failure risk
+  (i.e. ``${}``) from use in TALES path expressions to reduce the failure risk
   for the ``chameleon`` interpolation heuristics
   (`#925 <https://github.com/zopefoundation/Zope/issues/925>`_)
+
+Fixes
++++++
 
 - Fix ``length`` for page template repeat variables
   (`#913 <https://github.com/zopefoundation/Zope/issues/913>`_)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.2 (unreleased)
 ------------------
 
+- Exclude characters special for ``chameleon``'s interpolation syntax
+  (i.e. ``${}``) from use in path expressions to reduce the failure risk
+  for the ``chameleon`` interpolation heuristics
+  (`#925 <https://github.com/zopefoundation/Zope/issues/925>`_)
+
 - Fix ``length`` for page template repeat variables
   (`#913 <https://github.com/zopefoundation/Zope/issues/913>`_)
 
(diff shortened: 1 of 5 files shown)


d-maurer:
The py3.7 test failure in the Travis PR build was caused by a `Connection reset by peer` -- i.e. a temporary error. It has nothing to do with this PR.

jugmac00:
> The py3.7 test failure in the Travis PR build was caused by a `Connection reset by peer` -- i.e. a temporary error. It has nothing to do with this PR.

Whenever there are temporary glitches, you can hit the `restart build` button in the upper right corner:
https://travis-ci.com/github/zopefoundation/Zope/jobs/423726752

That said - I am not sure who has the permissions to do that.

I just restarted the build and it is green now.

dataflake (approved):

icemac:
I think this change is necessary (I've already been bitten by the problem), but it is backwards incompatible, so the change log could be a bit more explicit about this being a breaking change (e. g. by using an additional headline „Backwards incompatible changes“).

d-maurer:
Michael Howitz wrote at 2020-11-2 23:32 -0800:
>I think this change is necessary (I've already been bitten by the problem), but it is backwards incompatible, so the change log could be a bit more explicit about this being a breaking change (e. g. by using an additional headline „Backwards incompatible changes“).

I do not think that the backward incompatibility is of significant
practical importance: it essentially affects mapping access
via path expressions with keys containing one of the excluded characters.
`$` has been excluded already before and `{}` are rarely used
in keys. In addition, from Zope 4.0 until Zope 4.4 (exclusive)
those characters were already illegal in path expressions.

Do you nevertheless think that the incompatiblity should be flagged out?

dataflake:
@d-maurer I would flag it up if Zope 4.0-4.4 had not flagged it up already. Just so people have some awareness.

icemac:
> Do you nevertheless think that the incompatiblity should be flagged out?

Okay, you convinced me. Maybe I ran into this problem during the migration to Zope 4.

PR #930, opened 2020-11-13
Ensure native string response headers

d-maurer:
Apparently (--> "https://github.com/zopefoundation/Zope/pull/905#issuecomment-726018560"), `waitress` insists on native strings as values for HTTP response headers and rejects `unicode` values on Python 2.

Looking into this problem, I noticed that under Python 3 `bytes` header values are not handled correctly (this time by `ZPublisher.HTTPResponse._scrubHeader`).

This PR ensures proper use of native strings as HTTP response header values converting via `ISO-8859-1` if necessary.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,14 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 
 - Update dependencies to the latest releases that still support Python 2.
 
+Fixes
++++++
+
+- Convert ``bytes`` (Python 3) and ``unicode`` (Python 2) values for
+  HTTP response headers into native strings using the HTTP/1.1
+  stipulated ``ISO-8859-1`` encoding. This makes ``waitress`` happy
+  which insists on native strings for those values.
+
 
 4.5.2 (2020-11-12)
 ------------------
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -113,6 +113,10 @@ _CRLF = re.compile(r'[\r\n]')
 
 
 def _scrubHeader(name, value):
+    if PY3 and isinstance(value, bytes):
+        # handle ``bytes`` values correctly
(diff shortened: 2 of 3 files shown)


dataflake (approved):

mauritsvanrees (approved):
LGTM. This works both with and without my `plone.restapi` fixes from yesterday, both in Python 2 and 3.

dataflake:
@d-maurer Is there anything here that would be useful on `master`?

d-maurer:
Jens Vagelpohl wrote at 2020-11-14 10:16 +0000:
>@d-maurer Is there anything here that would be useful on `master`?

Maybe, the bit concerning a `bytes` value as HTTP header.
Not sure, however, whether it is of practival importance:
Python 3 makes handling of `bytes` that tedious that likely
nobody will try to use `bytes` as header value.

mauritsvanrees:
This fix is very much needed. You cannot even view the homepage of a Plone site without this, because some Plone code sets the `Content-Language` header to unicode in Python 2. See https://github.com/plone/buildout.coredev/pull/688
What is crazy, is that this is not caught by any test in Plone. There are lots of browser tests, but they don't go through `waitress`, which is the part that raises the error...
In the linked PR, I add Zope to the Plone checkouts, let's see if the tests pass.

jugmac00:
Thanks for the update, @mauritsvanrees!

I also noted a whole new category of problems, not covered by any of my 80% test suite for my Zope app, as the errors only show up when the pages are served by waitress.

If you ever find a proper solution for testing (eg running waitress), please keep us posted.

I was thinking about typing.... but Python's type system is not strong enough to catch all problems - left aside there is no aspiration to type Zope at all.

mauritsvanrees:
@jugmac00 Do you mean you noticed new problems starting with the Zope release 4.5.1 from last week? Or are these older problems?

Note: with the Zope 4.x checkout, the Plone tests are green again.

jugmac00:
@mauritsvanrees These were problems since the Python 2 -> 3 migration, which only surfaced recently.

PR #933, opened 2020-11-19
Fix longstanding issue with incorrect identification of Python memory…

darryldixon:
… address strings

This is used so that generic object __repr__ strings are not returned to a client browser. The code to perform this function has existed for a long time and been buggy the entire time because the regex used is incorrect. The existing (wrong) behaviour can be observed by putting in the name of any generic view against a running zope instance, eg:

http://localhost:8080/resource

returns:

<Zope2.App.traversing.resource object at 0x7f81ed0c1d90>

But what the code in HTTPResponse.py *tries* to do is identify these and return 'Not Found' instead.

### DIFF
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -69,7 +69,7 @@ status_codes['resourcelockederror'] = 423
 
 start_of_header_search = re.compile('(<head[^>]*>)', re.IGNORECASE).search
 base_re_search = re.compile('(<base.*?>)', re.I).search
-bogus_str_search = re.compile(b" [a-fA-F0-9]+>$").search
+bogus_str_search = re.compile(b" 0x[a-fA-F0-9]+>$").search
 charset_re_str = (r'(?:application|text)/[-+0-9a-z]+\s*;\s*'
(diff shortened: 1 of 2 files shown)


jugmac00:
Hi @darryldixon 
I am not able to review this PR, but it would certainly make it easier to review for others when the test suite passes.
https://travis-ci.com/github/zopefoundation/Zope/jobs/443289619
Thanks you!

darryldixon:
Hi @jugmac00 I updated the relevant test to correctly represent what it is supposed to be testing.

icemac:
I tried the change. It changes http://localhost:8080/resource from

before

to using the error page:

after

Thank you for your contribution.

According to the contributing policies of the zopefoundation organization you need to sign a contributor agreement before any non-trivial change before it can be merged. For details please consult the [Contributing guidelines for zopefoundation projects](https://github.com/zopefoundation/.github/blob/master/CONTRIBUTING.md).

dataflake:
@icemac IMHO this is not a non-trivial change. It really is trivial. No one would ever be able to derive some kind of copyright on a contribution that adds `0x` in two places. This is OK for merging.

icemac:
Thank you for looking into this PR. 😀

PR #935, opened 2020-11-20
Port of #933 to 4.x

dataflake:
I created a PR for this backport because it required an additional fix in response handling that I need some eyes on. The fix only affects Python 2 where the `setBody` method receives a body of type `io.BytesIO` sometimes, and the code would misinterpret that as one of the objects to raise the `NotFound`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix rendering of not found resources.
+  (`#933 <https://github.com/zopefoundation/Zope/pull/933>`_)
 
 
 4.5.3 (2020-11-16)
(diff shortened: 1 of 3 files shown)


d-maurer (changes requested):
I think `BytesIO` should be handled elsewhere: The "normal" place to handle IO responses (and other advanced responses such as iterator responses) is `WSGIResponse.setBody`. This uses `isinstance(body, IOBase)` to recognize this case. Due to the definition of `IOBase` (near the start of `HTTPResponse`) it does not correctly recognize `BytesIO`.

In my view, we should fix the `IOBase` definition. Instead of
```python
if sys.version_info >= (3, ):
    from io import IOBase
else:
    IOBase = file  # NOQA
```
I propose
```python
from io import IOBase
if PY2:
    IOBase = (file, IOBase)
```

icemac:
Dieter's proposal sounds interesting but I do not feel competent for this PR.

dataflake:
@d-maurer Thank you for looking at this. Your suggestion makes a lot more sense. I only changed one aspect: I want to avoid shadowing the imported name `IOBase` with that tuple you suggested, so I created a separate variable.

d-maurer (approved):

PR #941, opened 2021-02-02
Prevent DeprecationWarning

icemac:
`zope.datetime` has the following functions where we have duplicates in `App.Common`: 

* `iso8601_date`
* `rfc850_date`
* `rfc1123_date`

Since Python 3.8 they produce a deprecation waring when using them with a `DateTime` object:

`DeprecationWarning: an integer is required (got type DateTime).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.`

`zope.datetime 4.3` fixed this warning.
I decided to reduce duplicate code by using the zope.datetime functions here, too.


dataflake (approved):
👍👍👍

icemac:
@dataflake Do you think this PR should be ported to 4.x, too?

dataflake:
I would do it. It drops duplicate code and the earlier people know about the moved imports the better.

icemac:

icemac:
Ported to Zope 4 in https://github.com/zopefoundation/Zope/commit/2c736f0d59086e31312eba7ebd313fbc13c35b5b.

PR #942, opened 2021-02-03
Config with zope product

icemac:


src/Zope2/utilities/tests/test_zconsole.py
```diff
@@ -92,6 +92,6 @@ def test_runscript(self):
         # so we have the raw string in there.
         expected = (
             "42\n"
-            r"['run', {!r}, {!r}, {!r}, 'baz']"
-            "\nPropertyManager\n".format(self.zopeconf, script, bar))
+            rf"['run', {self.zopeconf!r}, {script!r}, {bar!r}, 'baz']"
```
sallner, line 95: Hm, I wonder with the `r"..."` is needed here.

sallner (approved):
Looks good to me. It think, there should be another PR for the `4.x` branch, after this is merged.

sallner:
Depending on what we want to achieve with the appveyor builds, we could also consider to let the windows test run on github actions. Either in the same tests.yml or in a special win(dows).yml, if the configuration is too different.

src/Zope2/utilities/tests/test_zconsole.py
icemac: https://github.com/zopefoundation/Zope/blob/058107d60fcd7b0bb461b8dfb4ead7257236a35f/src/Zope2/utilities/tests/test_zconsole.py#L91-L92 has the explanation

icemac:
It should be possible to run tests on windows on GHA (as GH now belongs to MS). I first goal was to get the CI tests running again as before but scripted in the generation of the config files. Switching away from AppVeyor comes a bit down the list.

icemac:
Finally I found a way that the tests run on GHA an AppVeyor. 😺

dataflake (approved):
👍👍👍  Thanks for doing all that work!

icemac:

PR #944, opened 2021-02-05
Fix Unauthorized views

### LINKED ISSUE #943: Exception view for Unauthorized exceptions not working

It looks like registering a view for `Unauthorized` doesn't work correctly due to the way `Unauthorized` is handled in the WSGI publisher. The view will get called correctly, but in  https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/WSGIPublisher.py#L212:L216 any response status the view may have set, such as a 302 redirect, is overwritten by the status from the exception, which is 401 Unauthorized. This will cause the browser to pop up the basic HTTP authentication box.

### What I did:

Registered a view for Unauthorized, the view code attempts to redirect the browser to a login page (example taken from Products.CMFDefault):

```
  <browser:page
      for="zExceptions.Unauthorized"
      layer="Products.CMFDefault.interfaces.ICMFDefaultSkin"
      name="index.html"
      class=".authentication.UnauthorizedView"
      permission="zope.Public"
      />
```

### What I expect to happen:

The login page should show when I browse to a page I don't have enough privileges for

### What actually happened:

The browser pops up a basic HTTP authentication prompt

Zope 4.x branch on Python 2

dataflake:
Fixes #943

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.4 (unreleased)
 ------------------
 
+- Fixed Unauthorized exception handling when a view has been registered for it
+  (`#943 <https://github.com/zopefoundation/Zope/issues/943>`_)
+
 - Revised debug info GUI
   (`#937 <https://github.com/zopefoundation/Zope/pull/937>`_)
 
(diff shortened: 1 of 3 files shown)


icemac:
Do you see a chance to write a test for this change?
What does `Products.PluggableAuthService` differently as it does not seem to be affected by this problem?

dataflake:
I have looked at it again and noticed I can work around it. The registered view must set the response status and declare it as locked with `RESPONSE.setStatus(302, locked=True)`, that way the publisher cannot overwrite it with a 401 afterwards. I'll close this PR.

> What does `Products.PluggableAuthService` differently as it does not seem to be affected by this problem?

I just looked after coming up with the workaround and saw that plugins that show a login page, like the CookieAuthHelper, set the redirect status to "locked" as well. This works for both ZServer and WSGI instances. ZServer doesn't fiddle with the response status afterwards so it worked without that locked status before.

PR #945, opened 2021-02-09
Fix AppVeyor

icemac:
We still need easy_install (removed in setuptools 52) to install packages which have no wheels.


icemac:
This way at least the AppVeyor builds run successfully. Next step would be to find the difference to #942 where they fail.

PR #949, opened 2021-02-11
The ``compute_size`` method properly returns None if the content does…

### LINKED ISSUE #948: The compute_size method checks is not acquisition safe
I have a container that implements `get_size` and I call:
```python
container.compute_size(child)
```
If child has no `get_size` method the returned value is the one of the container.

ale-rt:
… not  have a ``get_size`` method but the parent has.

Fixes #948


ale-rt:
There has been some changes in Plone and now the Plone Site object has a `get_size` method.
With the patch:
image
without the patch:
image

Anyway I think the patch is good in general. AFAIK `compute_size` is only used to display the size of an object in the ZMI.

dataflake (approved):
LGTM - should go into 4.x as well.

ale-rt:
I squashed the changes and fixed a trailing space in the CHANGES file that was hurting my feelings.

4.x -> https://github.com/zopefoundation/Zope/pull/950

I also started a 5.0.x (I do not know if you are interested in having it patched as well):

5.0.x -> https://github.com/zopefoundation/Zope/pull/951

dataflake:
Thanks for the effort, but there's no need for a 5.0.x maintenance branch. I closed the PR and deleted the branch. We should only do maintenance branches if there's a good reason and we all agree to it before creating that branch. It all adds to the maintenance and release burden.

PR #951, opened 2021-02-11
[5.0.x] Backport fixes for the ``compute_size`` method

ale-rt:
The `compute_size` method properly returns `None` if the content does not have a `get_size` method but the parent has.

Refs #948


dataflake:
Please do not open any maintenance branches when there's no pressing need to have one. The only valid maintenance branches at this point are `4.x` for Zope 4 and `master` for Zope 5. I'll close this PR and remove the branch.

ale-rt:
Ok, it is just that Plone is still on 5.0 for the moment, but I think soon we will move to 5.latest

PR #953, opened 2021-02-26
Package reconfiguration based on zopefoundation/meta/config

dataflake:
I started out using zopefoundation/meta/config, but there were too many small issues. The main goal was to come close, and above all do CI using GitHub.


icemac (approved):
Thank you for bringing CI back to live for 4.x!
I think I understand why you did the changes manually and I am fine with it.

PR #955, opened 2021-03-01
Enforce Zope permissions for recursive XML-RPC data dumps

### LINKED ISSUE #954: XML-RPC object representations don't respect permissions
XML-RPC requests that don't correspond to specific methods will fall back to generating a simple XML representation of the object being accessed. If the object has attributes that are themselves instances, the code will continue recursing and will generate representations for them as well.

The security checks performed in this recursive function are too simplistic. It will only exclude any attributes starting with the underscore `_`. It will not attempt to look up and check Zope permission assertions on the instances it finds during recursion. So if you have a folder that is accessible to a user and that folder contains items that are not, those items will still be included in the XML output.

dataflake:
Fixes #954 

This PR attempts to detect attributes that are Zope objects during creation of an XML-RPC object dump and enforce at least the `View` permission on them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.5 (unreleased)
 ------------------
 
-- Enforce Zope permissions for recursive XML-RPC data dumps
+- Enforce Zope permissions during recursive XML-RPC data dumps
   (`#954 <https://github.com/zopefoundation/Zope/issues/954>`_)
 
 - The ``compute_size`` method properly returns None if the content does not
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.5.5 (unreleased)
 ------------------
 
+- Enforce Zope permissions for recursive XML-RPC data dumps
+  (`#954 <https://github.com/zopefoundation/Zope/issues/954>`_)
+
 - The ``compute_size`` method properly returns None if the content does not
   have a ``get_size`` method but the parent has.
   (`#948 <https://github.com/zopefoundation/Zope/issues/948>`_)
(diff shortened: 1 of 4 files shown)


mauritsvanrees (approved):
LGTM. Works okay in my first testing.

icemac (approved):
LGTM.

One open question remains: Is the actual feature meaningful/needed?
Could it be an idea to just return the `repr` of the object in case of such unspecific calls? Or am I completely mistaken here and this feature has a broader meaning for the discoverability of XML-RPC methods? – Maybe listening the instance attributes could be tight to DEBUG mode?

icemac:
Hm, it GHA did not run this PR, the status is „Queued“, see https://github.com/zopefoundation/Zope/actions/runs/610050475 – strange

dataflake:
I am against completely changing something that has existed in this form since Day 1. Let's keep it as-is and just add the permissions checking.

dataflake:
P.S.: We could consider removing all XML-RPC functionality for Zope 6? I don't know who is using it at all. XML-RPC has gone by the wayside, people use restful APIs and microservices now.

jugmac00:
Please do not remove XML-RPC. My app heavily relies on it in order to communicate with 3rd party applications.

I know it is old, but afaik it is the only, and especially a super easy interface to get data out of Zope via other web services.

We have no out-of-the-box support for REST, as Plone has.

dataflake:
OK, fair enough

d-maurer:
Jürgen Gmach wrote at 2021-3-2 00:14 -0800:
>Please do not remove XML-RPC. My app heavily relies on it in order to communicate with 3rd party applications.

The same is true for one of my clients.

d-maurer:
Michael Howitz wrote at 2021-3-1 23:24 -0800:

>One open question remains: Is the actual feature meaningful/needed?

Almost surely, this is an artefact:
it is nice to be able to return an instance from an XML-RPC method
and get this properly serialized -- like a nested dict.
Likely, it was not anticipated that the return value could have been
acquired.

PR #963, opened 2021-04-24
field2* converters should not handle files

### LINKED ISSUE #558: Several possible problems within the ZPublisher Converters module

jugmac00:
Until now field converters could be applied on files with unclear outcome.

The problems with trying to convert files to text (detect MIME type, guess charset...)
were discussed in length in https://github.com/zopefoundation/Zope/issues/558#issuecomment-485064104.

Handling files only makes sense for the `field2bytes` converter.

This fixes #558

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
   ``rfc850_date``, and ``rfc1123_date`` which used to be in ``App.Common``
   keeping backwards-compatibility imports in place.
 
+- With the exception of ``field2bytes``, field converters should not handle
+  files (`#558 <https://github.com/zopefoundation/Zope/issues/558>`_)
+
 
 5.1.2 (2021-03-02)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,8 @@ The change log for the previous version, Zope 4, is at
 https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 
-5.1.3 (unreleased)
-------------------
+5.2 (unreleased)
+----------------
 
 - Updated/fixed the poll application tutorial in the Zope Developers Guide
   (`#958 <https://github.com/zopefoundation/Zope/issues/958>`_)
@@ -20,8 +20,12 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
   ``rfc850_date``, and ``rfc1123_date`` which used to be in ``App.Common``
   keeping backwards-compatibility imports in place.
 
-- With the exception of ``field2bytes``, field converters should not handle
-  files (`#558 <https://github.com/zopefoundation/Zope/issues/558>`_)
+Backwards incompatible changes
+++++++++++++++++++++++++++++++
+
+- With the exception of ``field2bytes``, field converters do no longer try to
+  read file like objects
+  (`#558 <https://github.com/zopefoundation/Zope/issues/558>`_)
 
 
 5.1.2 (2021-03-02)
--- a/src/ZPublisher/Converters.py
+++ b/src/ZPublisher/Converters.py
@@ -23,13 +23,8 @@ default_encoding = 'utf-8'
 
 
 def field2string(v):
-    """Converts value to native strings.
-
-    So always to `str` no matter which Python version you are on.
-    """
-    if hasattr(v, 'read'):
-        return v.read()
-    elif isinstance(v, bytes):
+    """Converts value to string."""
+    if isinstance(v, bytes):
         return v.decode(default_encoding)
     else:
         return str(v)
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -20,6 +20,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
   ``rfc850_date``, and ``rfc1123_date`` which used to be in ``App.Common``
   keeping backwards-compatibility imports in place.
 
+- With the exception of ``field2bytes``, field converters should not handle
+  files (`#558 <https://github.com/zopefoundation/Zope/issues/558>`_)
```
icemac, line 24: Without reading the diff it was not clear to me what has changed. `should not` is not really expressive. I'd expect something like "no longer try to handle file like objects special".

I think this is a even a breaking change and should be marked as such here. (Usually we use a headline for breaking changes.) It should be enough to increase the version number to 5.2.

src/ZPublisher/Converters.py
```diff
@@ -23,20 +23,15 @@
 
 
 def field2string(v):
-    """Converts value to native strings.
-
-    So always to `str` no matter which Python version you are on.
-    """
-    if hasattr(v, 'read'):
-        return v.read()
-    elif isinstance(v, bytes):
```
icemac, line 32: I am not sure if there should be a deprecation (maybe we could add it to Zope 4) or at least a warning (here in Zope 5).

icemac (changes requested):
I general I like the idea, i just have some questions resp. suggestions.

CHANGES.rst
jugmac00: Thank you for the suggestions which I just applied.

src/ZPublisher/Converters.py
jugmac00: Well, that is up to you do decide.

I certainly agree it is a good idea to add a deprecation warning for Zope 4.

At the current state of this PR, a warning in Zope 5 is no longer reasonable, as the code path is gone.

If you insist on a deprecation warning in Zope 5, I'd need to create a new PR, and stall this one until Zope 6? I am not sure if this is necessary.

d-maurer (approved):
I do not think we need deprecation warnings in Zope 5: Zope 5 is Python 3 only and the file case is wrongly handled by the converters; thus, any user should already have been warned.

From my point of view, deprecation warnings are also not necessary in Zope 4: I assume that the "convert file" case is very rarely used.

icemac (approved):

src/ZPublisher/Converters.py
icemac: I am fine with the current solution.

jugmac00:
@icemac and @d-maurer, thank you for taking the time for the review!

Issue #962, opened 2021-04-24
deprecate "unicode" converters (ulines, utext, utoken, ustring)


jugmac00:
The above mentioned converters make no more sense since Zope is Python 3 only.

This was discussed in https://github.com/zopefoundation/Zope/issues/558#issuecomment-491155828

I suggest a two-step approach:
- rewrite the current converters so that e.g. `field2ustring` just calls `field2string` + a deprecation warning
- in a later version, e.g. Zope 5.2 or 6, let's remove all `u<xxx>` converters and their functions (`field2u<xxx>`)

That said, e.g. `field2utext` is currently untested and broken for byte input.

jugmac00:
I am about creating a PR to deprecate the `unicode` converters by updating the code like this..

```
+def field2ustring(v):
+    warnings.warn(
+        "The converter `field2ustring` is deprecated "
+        "and will be removed in Zope 6. "
+        "Please use field2string instead.",
+        DeprecationWarning)
+    return field2string(v)
```

This works really nicely... except for `field2ulines` and `field2lines` - while the former returns a list of strings, the latter returns a list of bytes - but only since a "recent" change, see https://github.com/zopefoundation/Zope/pull/206

Whether `field2lines` should return strings or bytes was heavily discussed, and it was decided to return bytes! :-(

I have both grepped the Zope repositories, all GitHub repositories, and searched on Google why one would use the `field2lines` converter and for which fields. I did not find much. Possibly this is used on textareas?

Our current documentation for Zope 5 (= Python 3) says:

> lines – Converts a variable to a Python list of native strings by splitting the string on line breaks. Also converts list/tuple of variables to list/tuple of native strings.

> ulines, utokens, utext – like lines, tokens, text, but always converts into unicode strings.

Both native string and unicode strings corresponds to text in Python 3.

I still suggest to deprecate all `u`-converters and let `field2lines` return text - and no longer bytes (after a deprecation period).

If anybody is against this change, please explain me what use cases there would be to convert incoming form submissions to a list of bytes (remember, we already have `field2bytes`), and why this use case is so important. Also, if there is really anybody out there in need of a `field2byteslist`, one could register it oneself, or we could add one.

d-maurer:
Jürgen Gmach wrote at 2021-5-8 06:10 -0700:
> ...
>Whether `field2lines` should return strings or bytes was heavily discussed, and it was decided to return bytes! :-(

We should rethink this decision: it makes `field2lines` really
[...]

PR #965, opened 2021-05-08
Deprecate usage of `unicode` converters

jugmac00:
After dropping Python 2, there is no more use of `unicode` converters.

This means, after a deprecation period, that e.g. field2utext will be
removed and only field2text should be used any more.

P.S.: This PR currently fails - for reasons see https://github.com/zopefoundation/Zope/issues/962#issuecomment-835357220

Also, please keep the discussion in https://github.com/zopefoundation/Zope/issues/962

### DIFF
--- a/src/App/tests/testManagement.py
+++ b/src/App/tests/testManagement.py
@@ -1,3 +1,5 @@
+import warnings
+
 import Testing.ZopeTestCase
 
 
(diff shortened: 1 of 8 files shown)


src/App/tests/testManagement.py
```diff
@@ -18,21 +20,39 @@ def test_Management__Navigation__manage_page_header__1(self):
 
     def test_Management__Navigation__manage_page_header__2(self):
         """It respects `zmi_additional_css_paths` ustring property."""
-        self.folder.manage_addProperty(
-            'zmi_additional_css_paths', '/foo/bar.css', 'ustring')
+        # unicode converters will go away with Zope 6
```
dataflake, line 23: I don't think there's a good reason why this test code uses `ustring`/`ulines` properties at all. I suggest you convert it to `string`/`lines` instead.

src/ZPublisher/Converters.py
```diff
@@ -22,14 +23,19 @@
 default_encoding = 'utf-8'
 
 
-def field2string(v):
-    """Converts value to string."""
+def convert_single_value_to_string(v):
```
dataflake, line 26: I'm confused, why did you create this additional function? Its body can go into `field2string` and then you can `field2string` where needed instead of this extra method.

dataflake (changes requested):

src/App/tests/testManagement.py
jugmac00: Good catch!

src/ZPublisher/Converters.py
jugmac00: Meanwhile the code for the module is way easier to read - this was not the case a couple of months ago.

The `field2xxx` converters called each other, partly multiple times, and the called `field2xxx` converter called another `field2xxx` converter. It was very hard to reason about the code.

I extracted the `value_to_string` into a separate function to clearly separate between what is a field converter and what is a common base function.

It is up to you - I could prefix `convert_single_value_to_string` with an underscore to show my intention, or I revert the change and let other converters call `field2string` again.

dataflake: My vote is for reverting that part of the change and let other converters just call `field2string`. I always found it logical and easy to follow that converters for compound values such as lines or tokens would utilize the converters for simple values internally.

jugmac00: Very well!

dataflake (approved):

jugmac00:
Thanks for the review!

PR #970, opened 2021-05-11
Fix generating documentation.

icemac:
Use `Sphinx < 4` to prevent problems repoze.sphinx.autointerface.


icemac:

PR #971, opened 2021-05-21
fix path expressions for the `chameleon.tales` expression engine

d-maurer:
Fixes the attribute name bug (`traverseMethod` instead of `traverse_method`) reported by @mauritsvanrees 

Also adds a test to verify path expressions when implemented via `[un]restrictedTraverse`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.2 (unreleased)
 ----------------
 
+- Fix path expressions for the ``chameleon.tales`` expression engine.
+
 - Provide friendlier ZMI error message for the Transaction Undo form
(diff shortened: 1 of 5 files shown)


dataflake:
This change may cause conflicts with my pending security fix. Let's please hold this until the security fix is in.

dataflake:
... or I can merge this into the security fix? What do you think?

d-maurer:
> ... or I can merge this into the security fix? What do you think?

As you like: this PR essentially only corrects the attribute name and adds a traversal test. There should not be a "real" conflict potential with the security fix.

dataflake:
The attribute name change is already part of the security fix, that's why I am concerned about conflicts. I don't know how this private temporary fork setup in GitHub reacts if there's a conflict so I'm trying to be safe here. I'll take your changes and move them there, thanks for the added tests!

dataflake:
Changes are now moved, so I'm closing this PR.

PR #974, opened 2021-06-03
Switch PageTemplates to use `zope.untrustedpython`.

icemac:
I see no change in the tests and in behavior when testing in a TTW PageTemplate document with:

`<span tal:content="python:f'{template.__class__}" />`

Currently I am not sure that we need this change or could benefit from it.


dataflake:
What exactly is the goal of this change?

icemac:
> What exactly is the goal of this change?

That's not yet clear. Possibly we do not need it because we have hooked up the page template engine in another way.

loechel:
I do not see any dependency on that for Plone. As much as I love the idea of having TTW Page Templates secured, I am not sure how many people still use this and if that also works with chameleon.

Drop me a not if I should check some things after the merge of other packages.

icemac:
Closing as not relevant for Zope 5.

PR #984, opened 2021-07-02
Make sure "Manager" users can always modify proxy roles

dataflake:
See https://github.com/zopefoundation/Products.PythonScripts/issues/50

It makes no sense that "Manager" users can effectively be prevented from modifying a DTML Method or DTML Document with a proxy role the Manager-level user doesn't have. This PR circumvents the proxy role check for Manager users.

This also utilizes the cleaner implementation of `_validateProxyRoles` from `Products.PythonScripts`.


icemac (approved):
LGTM.

PR #990, opened 2021-10-13
Fix lint.

icemac:
This seems to be the only way to make the current check-python-versions version happy?!


dataflake:
IMHO it makes sense. `setup.py` should not implicitly claim support for more than all other indicators claim.

icemac:
This PR is no longer needed, see zopefoundation/meta#123.

PR #991, opened 2021-10-29
Update zope.i18n to 4.8.0.

mauritsvanrees:
Plone uses this already.


icemac:
@mauritsvanrees Is there a reason to keep this branch?

dataflake:
How about turning on branch auto-deletion upon PR merge? It's active already in many of our repositories. I have never encountered a case where I want to keep a branch after merging.

icemac:
@dataflake It is already active but it seems @mauritsvanrees manually restored the branch after it was automatically deleted.

dataflake:
You're right, I see that now. My guess would be an unintended button press by Maurits...

mauritsvanrees:
Looks like it. If I had a reason at that time, I have forgotten it, so it is fine that the branch is gone now. :-)

Issue #987, opened 2021-08-06
lines property in Zope 5.3: new one has text, old one has bytes


mauritsvanrees:
Zope 5.3 changed the lines property from storing bytes to storing text. See issue https://github.com/zopefoundation/Zope/issues/962 and PR https://github.com/zopefoundation/Zope/pull/965.

What now happens:

- Create a Zope 5.2.1 site, add a Folder A, and in the ZMI add a lines property on it. It contains a tuple of bytes.
- Update to Zope 5.3. (I simply switch between git tags.)
- Create folder B and add a lines property. It contains a tuple of texts (strings).
- The lines property in folder A still contains a tuple of bytes.
- Save the properties form in folder A, and it is converted to a tuple of strings.

So depending on when you created or edited the lines property, `getProperty` in Zope 5.3 may return bytes or string.
I don't really know if this is a problem or not.

But there are already some [test failures in GenericSetup](https://github.com/zopefoundation/Products.GenericSetup/issues/109), which is why I started looking. Apart from some expectations in tests which need to be updated, GenericSetup needs a few fixes in the part of its code where it is adding or updating properties, so the affected code is limited. See [PR](https://github.com/zopefoundation/Products.GenericSetup/pull/110). The most difficult about that PR, is that there is code in GenericSetup which supports non-utf8 encodings in profile files, so the changes are more convoluted than they should have been. Other packages which do similar things, would need to update their code as well.

If we think this should be fixed in Zope, I can see two solutions:

1. In `getProperty` call the converter of the field and return the result. But this means extra code gets run whenever you call `getProperty`. Not nice for performance.
2. Add a function that goes over all properties (or all lines properties) of an object, calls the converters, and saves changes. Integrators could apply this with `zopeFindAndApply`. Plone could choose to do this in a GenericSetup upgrade step, although this may take long when your Data.fs is 100 GB.

Additionally, a function could be useful to go over all deprecated properties, for example turning `ulines` into `lines`. The two functions could be combined.

Does someone already have code like this? Even when Zope itself does not call the function anywhere, having it available in Zope would be useful.
cc @jugmac00 

--------------

O, let me share some output from debugging. I started with Zope 5.2.1 and several properties containing non-ascii: `äh`. In the ZMI this looks fine in all fields, except for the `lines` field, where it looks wrong: `Ã¤h`. With `bin/instance debug`:

```
>>> app.testfolder.getProperty("lines")
(b'foo', b'\xc3\xa4h')
[...]

PR #993, opened 2021-12-02
Add function ZPublisher.utils.fix_properties.

mauritsvanrees:
You can call this to fix lines properties to only contain strings, not bytes.
It also replaces the deprecated property types ulines, utext, utoken, and ustring with their non-unicode variants.
See https://github.com/zopefoundation/Zope/issues/987 for why this is needed.

Note: the code is not called by Zope itself. The idea is that integrators who think they need this in their database, can call it.
Sample use:

    app.ZopeFindAndApply(app, apply_func=fix_properties)

I intend to use this (or a temporary copy) in the Plone 6 upgrade code.
See https://github.com/plone/Products.CMFPlone/issues/3305

I tried it on a few databases, and it works there, although only very little is changed.

### DIFF
--- a/src/ZPublisher/utils.py
+++ b/src/ZPublisher/utils.py
@@ -185,8 +185,7 @@ def fix_properties(obj, path=None):
             for prop in obj._properties:
                 if prop.get("id") == prop_id:
                     prop["type"] = new_type
-                    # This is a tuple, so force an object update to be sure
-                    obj._properties = obj._properties
+                    obj._p_changed = True
                     break
             else:
                 # This probably cannot happen.
--- a/src/ZPublisher/utils.py
+++ b/src/ZPublisher/utils.py
@@ -19,6 +19,7 @@ from Acquisition import aq_inner
 from Acquisition import aq_parent
 
 
+logger = logging.getLogger('ZPublisher')
 AC_LOGGER = logging.getLogger('event.AccessControl')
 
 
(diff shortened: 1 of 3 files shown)


src/ZPublisher/utils.py
```diff
@@ -100,3 +101,119 @@ def basic_auth_decode(token):
...
+            # This could easily lead to:
+            # Exceptions.BadRequest: Invalid or duplicate property id.
+            #   obj._delProperty(prop_id)
+            #   obj._setProperty(prop_id, new, new_type)
+            # So fix it by using internal details.
+            for prop in obj._properties:
+                if prop.get("id") == prop_id:
+                    prop["type"] = new_type
+                    # This is a tuple, so force a safe, just to be sure.
+                    obj._properties = obj._properties
```
jugmac00, line 189: I neither understand the comment (is there something missing after the `safe`?, nor do I understand the assignment expression.

mauritsvanrees: We change a dict in a tuple. Without this assignment, which marks the object as changed, the change may not be persisted in the zodb.

d-maurer:
Maurits van Rees wrote at 2021-12-1 23:59 -0800:
***@***.*** commented on this pull request.
>
>
>
>> +                    # This is a tuple, so force a safe, just to be sure.
>+                    obj._properties = obj._properties
>

Still the comment could be improved.
Suggestion: We changed a mutable subcomponent; ensure the change is made
persistent

src/ZPublisher/utils.py
```diff
@@ -100,3 +101,119 @@ def basic_auth_decode(token):
...
+            # Replace with non-unicode variant.
+            # This could easily lead to:
+            # Exceptions.BadRequest: Invalid or duplicate property id.
+            #   obj._delProperty(prop_id)
+            #   obj._setProperty(prop_id, new, new_type)
+            # So fix it by using internal details.
+            for prop in obj._properties:
+                if prop.get("id") == prop_id:
+                    prop["type"] = new_type
+                    # This is a tuple, so force a safe, just to be sure.
```
dataflake, line 188:
```suggestion
                    # This is a tuple, so force an object update, just to be sure.
```

dataflake (approved):

src/ZPublisher/utils.py
```diff
@@ -100,3 +101,119 @@ def basic_auth_decode(token):
...
+            # Replace with non-unicode variant.
+            # This could easily lead to:
+            # Exceptions.BadRequest: Invalid or duplicate property id.
+            #   obj._delProperty(prop_id)
+            #   obj._setProperty(prop_id, new, new_type)
+            # So fix it by using internal details.
+            for prop in obj._properties:
+                if prop.get("id") == prop_id:
+                    prop["type"] = new_type
+                    # This is a tuple, so force an object update, just to be sure.
```
dataflake, line 188: This line is now too long and lint is failing

```suggestion
                    # This is a tuple, so force an object update to be sure
```

src/ZPublisher/utils.py
jamadden: > Without this assignment, which marks the object as changed, the change may not be persisted in the zodb.

If `obj` is a persistent object, isn't the correct thing to do to set `obj._p_changed = True`? If it may or may not be persistent, then `if hasattr(obj, '_p_changed'): obj._p_changed = True` works if you don't want to do `if IPersistent.providedBy(obj):`.

mauritsvanrees: I know `obj._p_changed = True` would work, and am pretty sure we only get persistent objects in here, but to me it looks more magical than the current code. It's debatable though.
Do you think the current code is okay with the updated comment?

jamadden: To me, setting `obj._p_changed = True` is much more explicit and self explanatory than `obj.thing = obj.thing`. Setting `_p_changed` would probably eliminate the need for a comment at all.

The former (`_p_changed`) is a documented part of [the `IPersistent` protocol](https://persistent.readthedocs.io/en/latest/api/interfaces.html#persistent.interfaces.IPersistent); the latter relies on a side effect (and the side effect may not work depending on the implementation of the object, for example if the class has made `thing` a special type of descriptor).

mauritsvanrees: Okay, you convinced me. Done.

PR #995, opened 2021-12-20
Remove deprecated ulines, utext, utokens, ustring from more code.

mauritsvanrees:
- Zope PageTemplate still created a title property with type ustring.
  Strangely, after this commit, for existing Page Templates it is automatically a string, without needing to change anything.
- OFS tests had an export.xml with a ulines.
- ZTUtils/Zope.py had unreachable code that used ustring:
  the condition `if isinstance(v, str)` was already handled a few lines up

Also, in the properties form, show deprecation for ulines, utext, utokens, ustring.
Add a link to https://github.com/zopefoundation/Zope/issues/962 where they got deprecated.
Note: you can still add these properties in this form.
I would almost be inclined to remove them, but we should probably keep them in Zope 5.

### DIFF
--- a/src/ZTUtils/Zope.py
+++ b/src/ZTUtils/Zope.py
@@ -298,10 +298,6 @@ def simple_marshal(v):
     if isinstance(v, bytes):
         # Py 3 only
         return ':bytes'
-    if isinstance(v, str):
-        # Py 2 only
-        encoding = _default_encoding()
-        return f':{encoding}:ustring'
     if isinstance(v, bool):
         return ':boolean'
     if isinstance(v, int):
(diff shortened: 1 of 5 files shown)


src/ZTUtils/Zope.py
```diff
@@ -298,10 +298,6 @@ def simple_marshal(v):
     if isinstance(v, bytes):
         # Py 3 only
```
icemac, line 299:
```suggestion
```

This line is also no longer needed.

icemac (approved):
LGTM.

dataflake (approved):

src/ZTUtils/Zope.py
mauritsvanrees: Well spotted. I have removed the comment.

PR #1004, opened 2022-01-06
Improve type guessing for the default WebDAV PUT factory

### LINKED ISSUE #997: Webdav _default_PUT_factory not working as expected
### What I did:

After a Zope update (possibly to 5.3 but it could be an earlier one) certain files do not get converted to Zope objects as expected when uploaded via webdav.

### What I expect to happen:
- Files with the extensions 'html' or 'pt' should result in a Page Template object
- Image files (jpeg or png) should result in an Image object.

### What actually happened:
- files with the extension 'dtml' result in DTML Document objects
- all other files result in File objects.

### Where's the problem
When I upload a file, the type is always 'application/octet-stream'. So that is problably what's changed, hence the new, different behaviour.

As a DTML Document is recognized from its file extension, this one still works.

### Workaround
As a workaround other files could also be recognized from their file extension.

- Operating Systems:
    - MacOS Catalina
    - FreeBSD 12.2
- Python 3.8.6
- Zope 5.3

dataflake:
Fixes #997 content guessing

The PR adds content type guessing purely by file name extension if the generic `application/octet-stream` content type is passed in. It also fixes up Page Template issues with content type guessing.

### DIFF
--- a/src/webdav/NullResource.py
+++ b/src/webdav/NullResource.py
@@ -108,7 +108,7 @@ class NullResource(Persistent, Implicit, Resource):
 
         # Guess the type of file if the passed content-type is
         # just the generic application/octet-stream
-        if typ == 'application/octet-stream':
+        if not typ or typ == 'application/octet-stream':
             typ, encoding = guess_content_type(name, body)
 
         if ext == '.dtml':
--- a/src/webdav/NullResource.py
+++ b/src/webdav/NullResource.py
@@ -106,21 +106,17 @@ class NullResource(Persistent, Implicit, Resource):
         if not isinstance(body, bytes):
             body = body.encode('UTF-8')
 
+        # Guess the type of file if the passed content-type is
+        # just the generic application/octet-stream
+        if typ == 'application/octet-stream':
+            typ, encoding = guess_content_type(name, body)
+
         if ext == '.dtml':
             ob = DTMLDocument('', __name__=name)
         elif typ in ('text/html', 'text/xml'):
             ob = ZopePageTemplate(name, body, content_type=typ)
-        elif typ[:6] == 'image/':
+        elif typ.startswith('image/'):
             ob = Image(name, '', body, content_type=typ)
-        elif typ == 'application/octet-stream':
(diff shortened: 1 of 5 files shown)


src/webdav/NullResource.py
```diff
@@ -111,6 +112,15 @@ def _default_PUT_factory(self, name, typ, body):
...
             ob = Image(name, '', body, content_type=typ)
+        elif typ == 'application/octet-stream':
+            if ext in ('.jpg', '.jpeg', '.gif', '.png', '.tiff'):
+                ob = Image(name, '', body)
+            elif ext in ('.xml',):
+                ob = ZopePageTemplate(name, body, content_type='text/xml')
+            elif ext in ('.pt', '.zpt', '.html', '.htm'):
+                ob = ZopePageTemplate(name, body, content_type='text/html')
+            else:
+                ob = File(name, '', body, content_type=typ)
```
d-maurer, line 123: I do not like this part:
* if the client does not have content type related evidence it should not send content type information; in this case we would guess the content type (and, of course, may be wrong); if the client sends content type information, we should use it and do not guess
* I would not special case some image suffixes but instead delegate to `guess_content_type`; if this guesses some `image` type, I would create an `Image`; special casing some pure Zope extensions is okay

d-maurer (changes requested):
I do not like that information explicitly provided by the client is overridden based on heuristics.

If we really think that our guessing is better than an `application/octed-stream` provided by the client, we could have: `if tyo == 'application/octed-stream': typ = guess_content_type(....)`.

Note that XML-files may be associated with either `text/xml` or `application/xml` (or even `application/xml+<something>). We definitely should not recognize them only if type is either `text/xml` or `application/octed-stream'.

src/webdav/NullResource.py
dataflake: I have changed the code to just call `guess_content_type` if no content type information exists or if it is the generic `application/octet-stream` to cover Georg's broken client

d-maurer (approved):

Issue #1006, opened 2022-01-13
setBody handles lists differently than documented


viktordick:
### What I did:

We are in the process of finally moving a system from Zope 2 to Zope 5. There are Python scripts that are executed both externally by calling the URL and internally from other scripts, and they return a list of integers (database IDs). The script breaks in Zope 5 with

```
Traceback (innermost last):

    Module ZPublisher.WSGIPublisher, line 167, in transaction_pubevents
    Module ZPublisher.WSGIPublisher, line 376, in publish_module
    Module ZPublisher.WSGIPublisher, line 281, in publish
    Module ZPublisher.HTTPResponse, line 1093, in setBody
    Module ZPublisher.HTTPResponse, line 533, in setBody

ValueError: bytes must be in range(0, 256)

```
To reproduce, install `Products.PythonScripts` and add a script called `test` that says `return [1, 5, 500]`.

### What I expect to happen:
According to the docstring of `setBody`:
```
        If body is not a string or unicode, but has an 'asHTML' method, use
        the result of that method as the body;  otherwise, use the 'str'
        of body.
```
If I use `str([1,5, 500])`, I get the string `"[1, 5, 500]"`, which would be a valid output and would give the same result as in Zope 2.
### What actually happened:
The code actually says
```
                body = bytes(body)
```
which tries to convert the numbers into bytes and fails because they are larger than 255.

I am not sure if reverting this part to use `str` again will always give the right result.

### What version of Python and Zope/Addons I am using:

I am using Zope 5.3 with `Products.PythonScripts`, I think the rest is not important.

d-maurer:
Viktor Dick wrote at 2022-1-13 02:59 -0800:
> ...
>Traceback (innermost last):
> ...
>    Module ZPublisher.HTTPResponse, line 533, in setBody
>
>ValueError: bytes must be in range(0, 256)
>
>```
> ### What I expect to happen:
> According to the docstring of `setBody`:
> ```
>         If body is not a string or unicode, but has an 'asHTML' method, use
>         the result of that method as the body;  otherwise, use the 'str'
>         of body.
> ```
> > ...
>The code actually says
>```
>                body = bytes(body)
>```
>which tries to convert the numbers into bytes and fails because they are larger than 255.

Logically, the response body contains a sequence of bytes.
Converting to bytes (at some place) is therefore correct.

For Python 2 (and therefore Zope 2), `bytes` and `str` are identical.
For Python 3, this is no longer the case.

[...]

PR #1007, opened 2022-01-13
Use string representation of response data

viktordick:
Response data that has no `__bytes__` method but a `__str__` method,
like lists, dicts and similar, can now be handled by the publisher as it
was handled in Zope 2, by returning the string representation.

Fixes https://github.com/zopefoundation/Zope/issues/1006

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
-- Use string representation of response data that has no ``__bytes__``
+- Use intermediate ``str`` representation for complex response data (i.e. not
+  ``str`` or ``bytes`` like)
   (`#1006 <https://github.com/zopefoundation/Zope/issues/1006>`_)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Use string representation of response data that has no ``__bytes__``
+  (`#1006 <https://github.com/zopefoundation/Zope/issues/1006>`_)
+
 
 5.4 (2022-01-09)
 ----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,8 +13,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
-- Use intermediate ``str`` representation for complex response data (i.e. not
-  ``str`` or ``bytes`` like)
+- Use intermediate ``str`` representation for non-bytelike response data unless
+  indicated differently by the content type.
   (`#1006 <https://github.com/zopefoundation/Zope/issues/1006>`_)
 
 
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -528,6 +528,8 @@ class HTTPBaseResponse(BaseResponse):
             body = self._encode_unicode(body)
         elif isinstance(body, bytes):
             pass
+        elif not hasattr(body, '__bytes__') and hasattr(body, '__str__'):
+            body = self._encode_unicode(str(body))
         else:
             try:
                 body = bytes(body)
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -501,10 +501,10 @@ class HTTPBaseResponse(BaseResponse):
         If body has an 'asHTML' method, replace it by the result of that
         method.
 
-        If body is unicode, encode it. If it is more complex (no str or bytes
-        and no '__bytes__' method), use its 'str' representation and encode the
-        result. Else, convert it to bytes, falling back to the encoded 'str'
-        representation if this yields an error.
+        If body is now bytes, bytearray or memoryview, convert it to bytes.
+        Else, either try to convert it to bytes or use an intermediate string
+        representation which is then converted to bytes, depending on the
+        content type.
 
         If is_error is true, format the HTML as a Zope error message instead
         of a generic HTML page.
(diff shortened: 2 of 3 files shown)


viktordick:
Tested this on the system where I encountered the problem described in #1006 and it seems to work. I am in the process of executing the tests.

viktordick:
One test fails on my system, but it also fails on `master`.

@d-maurer If you think this might be a solution, I could add a test for the situation I described in #1006 and mark this ready for review.

CHANGES.rst
```diff
@@ -13,6 +13,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Use string representation of response data that has no ``__bytes__``
```
d-maurer, line 16: Suggestion: "Use intermediate ``str`` representation for complex response data (i.e. not `str` or bytes like)

d-maurer (commented):
Could you also improve the docstring? It does not reflect the implementation (`asHTML` is used without preceding type check) and still uses Python 2 terms.

icemac:
I think this is a good approach. After the comments of @d-maurer are fixed, I'd like to approve this PR.

viktordick:
@icemac thanks. I adjusted the docstring and added a test. While writing the docstring, I was wondering if there are cases where `body` has no `__bytes__` method, but we *still* want to use `bytes(body)` instead of `self._encode_unicode(str(body))`. If there is no such situation, I guess a more transparent order of operations would be something like
```
if not isinstance(body, bytes) and hasattr(body, '__bytes__'):
    try:
        body = bytes(body)
    except (TypeError, UnicodeError):
        pass
if not isinstance(body, bytes):
    body = self._encode_unicode(str(body))
```
to replace the complete if-chain.
I am aware that the current state of the PR is less invasive that this, I just wanted to mention my thoughts after reading the code.

d-maurer:
Viktor Dick wrote at 2022-1-16 23:41 -0800:
***@***.*** thanks. I adjusted the docstring and added a test. While writing the docstring, I was wondering if there are cases where `body` has no `__bytes__` method, but we *still* want to use `bytes(body)` instead of `self._encode_unicode(str(body))`.

There are.
An example is an instance of `io.BytesIO`: it has no `__bytes__` method
but logically contains a sequence of bytes (returned by `bytes(bio)`).
It is a natural choice when the application wants to return
a bytes response constructed incrementally.
```pycon:
>>> from io import BytesIO
>>> s=BytesIO()
>>> s.__bytes__
AttributeError: '_io.BytesIO' object has no attribute '__bytes__'
>>> bytes(s)
b''
```
Please verify that your PR can handle such objects correctly.


I had a look (in Python 3.9) at `bytes_new`
(essentially the implementation of `bytes(obj)`). It (essentially)
checks: "has `__bytes__`", "unicode",
"bytes like", "buffer like", "list", "tuple".

viktordick:
Well, I just tried the following:
```
>>> from io import BytesIO
>>> d = BytesIO(b'test')
>>> bytes(d)
TypeError: 'bytes' object cannot be interpreted as an integer
>>> d.getvalue()
b'test'
```
so `bytes(d)` does not in fact return the content of the `BytesIO` object. Am I missing something?

viktordick:
That is, if I want to use `BytesIO` to incrementally construct the response data, I need to
```
return result.getbuffer()
```
both before and after this PR.

viktordick:
OK sorry for the many messages.
In fact, `BytesIO.getbuffer` returns a `memoryview`, which has no `__bytes__` but allows for `bytes(...)` - not sure how. So yes, this is probably a regression of this approach.
I will have to think about it.

d-maurer:
Viktor Dick wrote at 2022-1-17 00:37 -0800:
>Well, I just tried the following:
>```
>>>> from io import BytesIO
>>>> d = BytesIO(b'test')
>>>> bytes(d)
>Traceback (most recent call last):
>  File "<stdin>", line 1, in <module>
>TypeError: 'bytes' object cannot be interpreted as an integer
>>>> d.getvalue()
>b'test'
>```
>so `bytes(d)` does not in fact return the content of the `BytesIO` object. Am I missing something?

Apparently, I have been wrong: Python does not support `bytes(BytesIO)`,
at least not unless the `BytesIO` is empty.

I expected `BytesIO` to implement the "buffer protocol".
To my surprise, it does not do that.
Therefore, `bytes(BytesIO)` does not take the "buffer like" case
but the "iterator like" case. In this case, the iterator is expected
to generate a sequence of integers but `BytesIO` generates a sequence
of bytes which leads to the `TypeError` above.


Nevertheless, there are objects without `__bytes__` for which `bytes(obj)`
succeeds. New example `bytearray`:
```pycon:
>>> b=bytearray(b'test')
>>> b.__bytes__
AttributeError: 'bytearray' object has no attribute '__bytes__'
>>> bytes(b)
b'test'


Further candidates are all objects supporting the "buffer protocol".

viktordick:
> Nevertheless, there are objects without `__bytes__` for which `bytes(obj)` succeeds. New example `bytearray`:

Well this becomes difficult then. Because `bytes([1, 2, 3])` also succeeds, but `bytes([1, 2, 500])` does not. But I would not want a different handling of lists depending on if the values happen to be integers in the range `[0,255]`.

But I guess we will have to do one of the sacrifices.
Solution 1: We simply try `bytes(body)` and fall back to `self._encode_unicode(str(body))` in case of errors, which will give a significantly different handling of `[1, 2, 3]` than `[1, 2, 500]`.
Solution 2: We use a specific list of types that are converted to `bytes` directly and use the string representation otherwise. The list would contain `bytes`, `memoryview`, `bytearray` and possibly others (I tried to check for how to know if a type implements the buffer protocol, but could not find anything clear).

Side note: I would in fact also be able to live with solution 1. My concrete problem is that the transaction rolls back due the error. The actual result is not really used, only stored for debugging purposes, so I could live with such inconsistencies. If the result of a method that is called by URL is actually to be used and it is not a simple str or bytes, the method should make sure to pack the result accordingly and not depend on the specifics of the `ZPublisher` (i.e., it should probably pack the list into JSON).
But I think solution 2 is more elegant.

viktordick:
@d-maurer Thanks for the suggestion, I implemented it, extending the first point ("If it is a `bytes` instance") to also include `bytearray` and `memoryview`.

`setBody` will still raise a `ValueError` if for example a sequence is passed with non-integers or integers outside the range of [0, 255] after explicitly setting the content type to expect binary output, but I guess this is expected and actually a good thing.

d-maurer (approved):

PR #1009, opened 2022-02-02
Fix several exceptions when calling ZPublisher.utils.fix_properties.

mauritsvanrees:
I created this function in Zope 5.4 and it worked for the databases I tried it on.
But on another project I ran into several exceptions.
See tracebacks in Plone:
https://github.com/plone/plone.app.upgrade/issues/270

Note that `plone.app.upgrade` has a temporary copy of the same function.
I fixed the function there as well today:
https://github.com/plone/plone.app.upgrade/pull/271

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.4.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


icemac (approved):
LGTM from reading the code – although I never used this fixing utilities.

dataflake (approved):

Issue #1008, opened 2022-01-26
Improved cookie control


d-maurer:
Currently (Zope 5.4, Zope 4.7), Zope provides not much control over cookie handling: cookie parameter support is limited by `HTTPBaseResponse._cookie_list` and there is no way to establish cookie policy defaults.

Recent browser versions (see e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) have significantly changed  cookie handling. In some cases, control over the `SameSite` cookie parameter becomes necessary.

We should improve cookie control via
1. implement a cookie parameter registry to easily support new cookie parameters
2. introduce an `ICookiePolicy` utility (in `HTTPBaseResponse._cookie_list`) to allow applications to enforce a cookie policy (e.g. add `SameSite=Lax`).

icemac:
@d-maurer There is already #600 regarding cookies. Are they different issues?

d-maurer:
Michael Howitz wrote at 2022-1-26 23:08 -0800:
***@***.*** There is already #600 regarding cookies. Are they different issues?

There are overlaps, e.g. the cookie parameter registry, but
cookie policy support is new.

The motivation for the new issue comes from a concrete client
problem:
His application is sometimes embedded in an "iframe".
Recent changes in some browsers (particularly "chrome" and "firefox")
regarding cookies and the "SameSite" cookie attribute let those
cases fail (cookies got no longer delivered).
Those changes mean that the default for the "SameSite" cookie parameter
is no longer "None" (i.e. no restriction for cookie delivery) but
"Lax" (allow cookie delivery for top level navigation, but not otherwise).

The client solved the problem via cookie rewriting in an HTTP agent
in front of Zope; thus, there are solutions outside of Zope.
A Zope based solution might provide finer grained control.

icemac:
This sounds like an interesting idea. If you have the time please go forward resolving this issue.

icemac:
Fixed by #1010 resp. #1011.

PR #1010, opened 2022-02-04
enhanced cookie support

### LINKED ISSUE #600: Modernize Zope's cookie support?
There is a new (from 2011) [International cookie stanard -- RFC6265](https://tools.ietf.org/html/rfc6265) which obsoletes all former international cookie standards (i.e. RFC2109 and RFC2965). The most important change is that "..." no longer quotes its content; especially a `;` contained in a quoted string is still treated as parameter separator and will highly likely be wrongly interpreted.

The good thing: the current cookie handling allows an application to create RFC6265 compatible cookies if it either has only elementary cookie requirements or detailed knowledge about the respective standards (e.g. it knows that cookie attribute values must not contain plain `;`). Thus, there is no pressing need to change something.

On the other hand, it does not help an application with non elementary cookie functionality requirements and it lets errors easily slip in. For example, it invents its own names for cookie attributes (e.g. `http_only` for `HTTPOnly`, `max_age` for `Max-Age`) and silently ignores any cookie attribute it does not know. It delegates the standards' requirements e.g. related to `Expires` or `Comment` entirely to the application -- and it requires quite some knowledge to provide
a valid date for `Expires`. Thus, there is a large potential for improvement.

I propose to modernize the cookie support once Python 2 is no longer supported (i.e. for Zope 5) (to avoid the complexitiies introduced by an unknown encoding of `str` values) including:
 - support passing a `datetime` (maybe `DateTime`, too) as value for `Expires`; the proper representation is automatically computed
 - support passing an object as value for `Path`; the proper string value is automatically computed from the objects `getPhysicalPath` and the current request
 - support all standard attribute names - besides those invented by Zope
 - transparently quote `;` found in cookie attribute values (via RFC2047 quoting)
 - raise exceptions for unknown cookie attributes
 - maybe provide a registry to register new cookie attributes (not yet directly supported by Zope)

d-maurer:
fixes #600 and #1008.

This PR introduces two cookie related policies, `ICookieParamPolicy` and `ICookieValuePolicy`, and a cookie parameter registry.

The cookie policies allow customization of cookie handling. Details can be found in `ZPublisher.interfaces`.

Cookie parameter customization may be helpful  for some applications  wanting to support embedding in foreign sites because recent browser versions use `Lax` as default for the `SameSite` cookie parameter. As a consequence cookies are no longer delivered to applications embedded in a foreign site unless their definition carries the parameters `SameSite=None; Secure`.
The default cookie parameter policy automatically adds a missing `Expires` if the `Max-Age` parameter is present (to support HTTP/1.0).

The current cookie standard (i.e. RFC 6265) points out that cookies are quite easily tampered with even when transmitted over HTTPS. If an attacker gets access to the browser's host, they can typically also be analyzed. Therefore, RFC 6265 recommends to sign or even encrypt sensitive cookie values.  The cookie value policy allows applications to implement this recommendation. The default cookie value policy does not sign/encrypt because Zope lacks the respective infrastructure.

The cookie policies are looked up as utilities. However, cookie parsing happens very early during request processing, at a time when only the global component registry is available. Therefore, the `ICookieValuePolicy` utility is looked up only in this registry. To remove this restriction, cookie parsing would need to be delayed until the cookie is accessed.

The cookie parameter registry makes cookie parameter handling more flexible and extensible:
 * `ZPublisher.HTTPResponse.HTTPResponse.setCookie` can refer to a parameter either with its official name (defined by the standard) or its Zope name; case is not significant and `-` can be replaced by `_`. For example, the parameter `Max-Age` can  also be referred to by `max_age`; `HttpOnly` also by `http_only`. Internally, the official names are used.
 * parameter values used in `setCookie` need not always be ASCII strings. The value for `Expires`  can be int/float (interpreted as timestamp related to `time.time`), naive or time zone aware `datetime` (if naive, it is assumed to be an UTC `datetime`), a (Zope) `DateTime` or a string (representing an RFC 1123 date). The value for `Path` can be an object with `absolute_url_path` method or a string. The value for `Domain` is automatically IDNA2003 encoded, if necessary. Internally parameter values are represented as `None` (drop the parameter), `True` (drop the parameter value, i.e. render as boolean true parameter) or an ASCII string (as used "on the wire").
 * `registerCookieParameter` can be used to register new parameters and/or override existing ones.

### DIFF
--- a/src/ZPublisher/cookie.py
+++ b/src/ZPublisher/cookie.py
@@ -32,6 +32,7 @@ from time import time
 from six.moves.urllib.parse import quote
 from six.moves.urllib.parse import unquote
 
+from Acquisition import aq_base
 from DateTime import DateTime
 from zope.component import getGlobalSiteManager
 from zope.component import queryUtility
@@ -46,7 +47,7 @@ class DefaultCookieParamPolicy(object):
     """Default ``ICookieParamPolicy``.
 
     It adds ``Expires`` when it sees ``Max-Age`` (for compatibility
-    with HTTP/1.0) and urlquotes cookie values.
+    with HTTP/1.0).
     """
 
     @staticmethod
@@ -264,9 +265,9 @@ def path_converter(value):
     If the string contains ``%``, it is assumed that it is already
     quoted.
     """
-    ap = getattr(value, "absolute_url_path", None)
+    ap = getattr(aq_base(value), "absolute_url_path", None)
     if ap is not None:
-        return ap()
+        return value.absolute_url_path()
     if path_safe(value):
         return value
     if "%" in value:
--- a/src/ZPublisher/cookie.py
+++ b/src/ZPublisher/cookie.py
@@ -237,16 +237,13 @@ def domain_converter(value):
     the registration to get IDNA2008 if you observe domain
     related problems.
     """
-    if isinstance(value, bytes):
-        value = value.decode("utf-8")
-    if "xn--" in value:  # already encoded
-        # ATT: under Python 2, this returns unicode
-        # this may pose problems should other parameters
-        # provide encoded non-ASCII
+    # Python 2 requires unicode
+    u_value = value.decode("utf-8") if isinstance(value, bytes) else value
+    if "xn--" in u_value:  # already encoded
         return value
     from encodings.idna import ToASCII
     from encodings.idna import nameprep
-    return ".".join(to_str(ToASCII(nameprep(c))) for c in value.split("."))
+    return ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
 
 
 registerCookieParameter("Domain", domain_converter)
--- a/src/ZPublisher/cookie.py
+++ b/src/ZPublisher/cookie.py
@@ -1,0 +1,337 @@
+##############################################################################
+#
+# Copyright (c) 2022 Zope Foundation and Contributors.
+#
+# This software is subject to the provisions of the Zope Public License,
+# Version 2.1 (ZPL).  A copy of the ZPL should accompany this distribution.
+# THIS SOFTWARE IS PROVIDED "AS IS" AND ANY AND ALL EXPRESS OR IMPLIED
+# WARRANTIES ARE DISCLAIMED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF TITLE, MERCHANTABILITY, AGAINST INFRINGEMENT, AND FITNESS
+# FOR A PARTICULAR PURPOSE.
+#
+##############################################################################
+"""HTTP cookie support.
+
+The implementation is based on
+`RFC6265 <https://www.rfc-editor.org/rfc/rfc6265.html>`_.
+
+The module introduces 2 features:
+
+ 1. default ``ICookieParamPolicy`` and ``ICookieValuePolicy``
+ 2. a cookie parameter registry
+
+and exports the functions ``getCookieParamPolicy``, ``getCookieValuePolicy``
+``registerCookieParameter``,
+``normalizeCookieParameterName`` and ``convertCookieParameter``.
+"""
+import datetime
+from itertools import chain
+from re import compile
+from time import time
+
+from six.moves.urllib.parse import quote
+from six.moves.urllib.parse import unquote
+
+from DateTime import DateTime
+from zope.component import getGlobalSiteManager
+from zope.component import queryUtility
+from zope.interface import implementer
+
+from .interfaces import ICookieParamPolicy
+from .interfaces import ICookieValuePolicy
+
+
+@implementer(ICookieParamPolicy)
+class DefaultCookieParamPolicy(object):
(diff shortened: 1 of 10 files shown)


d-maurer:
Formerly, quoting the cookie value allowed to use characters forbidden in unquoted values (notable space and `;`); with RFC 6265, this is no longer the case: cookie values must not contain space, controls or delimiters, whether quoted or not; quoting no longer changes anything.
For this reason, the PR no longer quotes cookie values. The corresponding `setCookie` parameter is retained for compatibility reasons but ignored.

src/ZPublisher/cookie.py
```diff
@@ -0,0 +1,337 @@
...
+                yield (pn, pv)
+        if "Expires" not in attrs and attrs.get("Max-Age") is not None:
+            max_age = int(attrs["Max-Age"])
+            expires = "Wed, 31 Dec 1997 23:59:59 GMT" if max_age <= 0 \
+                      else time() + max_age
+            yield convertCookieParameter("Expires", expires)
+
+    @staticmethod
+    def check_consistency(name, attrs):
+        return  # does not enforce any consistency rules
```
dataflake, line 66: A little higher up a very old comment in the `_cookie_list` response method says "should really check size of cookies here". If the RFC mentions a maximum size wouldn't it make sense to check it here?

src/ZPublisher/cookie.py
```diff
@@ -0,0 +1,337 @@
...
+
+defaultCookieParamPolicy = DefaultCookieParamPolicy()
+
+
+@implementer(ICookieValuePolicy)
+class DefaultCookieValuePolicy(object):
+    """Default ``ICookieValuePolicy``.
+
+    It adds ``Expires`` when it sees ``Max-Age`` (for compatibility
+    with HTTP/1.0) and urlquotes cookie values.
```
dataflake, line 77:
```suggestion

```

The comment is wrong here

src/ZPublisher/cookie.py
```diff
@@ -0,0 +1,337 @@
...
+    def register(self, name, converter, aliases=()):
+        """register cookie parameter *name* with *converter* and *aliases*.
+
+        *converter* is a function which will be applied to
+        parameter values and which is expected to either
+        raise an exception or convert the value into
+        either ``None``, ``True``
+        or an ASCII string without control characters and ``;``.
+        ``None`` means "drop the parameter",
+        ``True`` menas "drop the parameter value",
```
dataflake, line 125:
```suggestion
        ``True`` means "drop the parameter value",
```

src/ZPublisher/cookie.py
```diff
@@ -0,0 +1,337 @@
...
+        otherwise, the return value is used as value representation.
+
+        Some aliases are automatically derived from *name*:
+        convertion to lower, ``-`` to ``_`` conversion.
+        Further aliases can be defined via *aliases*.
+
+        It is not an error to override existing registrations.
```
dataflake, line 132:
```suggestion
        Existing registrations can be overridden.
```

src/ZPublisher/cookie.py
```diff
@@ -0,0 +1,337 @@
...
+
+def path_converter(value):
+    """convert *value* to a path.
+
+    The convertion is based on ``absolute_url_path``.
+    If *value* lacks this method, it is assumed to be a string.
+    If the string contains ``%``, it is assumed that it is already
+    quoted.
+    """
+    ap = getattr(value, "absolute_url_path", None)
```
dataflake, line 267: If value is a Zope object it may acquire `absolute_url_path` if it doesn't have it, leading to a bad path. This is unlikely but possible.

dataflake (approved):
Looking really good, thanks.

src/ZPublisher/cookie.py
d-maurer: RFC 6265 does not specify a cookie size limit. In section 6.1, it recommends that a browser should support cookies of at least up to 4096 bytes size. Not sure whether this should cause us to reject larger cookies.

src/ZPublisher/cookie.py
d-maurer: It adds `Expires` but does nothing with the value. I keep the comment but fix it.

src/ZPublisher/cookie.py
d-maurer: I will check for this.

icemac (approved):
LGTM from reading through the code.

d-maurer:
Thank you for the reviews!

dataflake:
The documentation on RTD fails to build now. https://readthedocs.org/projects/zope/builds/16005243/

d-maurer:
> The documentation on RTD fails to build now. https://readthedocs.org/projects/zope/builds/16005243/

I did nothing with `ZConfig` (or other dependencies).

icemac:
I fixed the documentation build in https://github.com/zopefoundation/Zope/commit/f8d9c5af9ab97d9ba4cb59751fb369b585d2916b. RTD did not know where to find the requirements file.

PR #1013, opened 2022-02-24
Remove typing pin

mauritsvanrees:
[`typing`](https://pypi.org/project/typing/) is for Python 3.4 and lower.

Plus some pre-commit changes, and fixes for lint.


mauritsvanrees:
Windows Py 3.10 [fails to install AccessControl 5.2](https://github.com/zopefoundation/Zope/runs/5321734070?check_suite_focus=true#step:6:57). Any idea?

d-maurer:
Maurits van Rees wrote at 2022-2-24 08:45 -0800:
>Windows Py 3.10 [fails to install AccessControl 5.2](https://github.com/zopefoundation/Zope/runs/5321734070?check_suite_focus=true#step:6:57). Any idea?

Apparently, a newer C[++] compiler is required.

Due to C[++] compiler problems, C extensions apparently cannot be
build (at least not for `zope.interface` and `AccessControl`).

icemac:
There is https://github.com/zopefoundation/AccessControl/pull/121 which will create a Python 3.10 wheel for windows when released. A review is welcome.

icemac (approved):
LGTM.
We can merge it when we have an AccessControl wheel for Python 3.10.

src/ZPublisher/tests/test_cookie.py
```diff
@@ -143,12 +143,12 @@ def test_domain(self):
         self.assertEqual(v, "fussball.example")
         # check ``bytes`` handling
         _, v = convertCookieParameter("domain",
-                                      "Fußball.example".encode("utf-8"))
```
dataflake, line 146: I'd rather keep the explicit encoding for clarity instead of relying on defaults.

dataflake (changes requested):

src/ZPublisher/tests/test_cookie.py
mauritsvanrees: This must have been the choice of the pre commit hook. If I change it now, the next person will have the same problem.
Any way around that?

jugmac00: This rewrite is caused by [pyupgrade](https://github.com/asottile/pyupgrade#encodeutf-8) and it is - unfortunately - not configurable. I think the only way to prevent this - while keeping `pyugrade` and `pre-commit` - would be to exclude the complete file via configuring the pre-commit hook in `pre-commit-config.yaml`.

dataflake: 😒 Never mind then... (this is why I strongly dislike automated code changes using tools where someone else makes choices for you. Some of these choices are bad or not appropriate.)

dataflake (approved):

mauritsvanrees:
I pinned the new AccessControl 5.3, but it still fails on Windows as it tries to compile:
https://github.com/zopefoundation/Zope/runs/5373528042?check_suite_focus=true

icemac:
See #1014 where I am trying to figure out what's needed for successful tests. It seems there are still some steps to go.

icemac:
Now CI should become green again.

PR #1014, opened 2022-02-25
Update ExtensionClass to latest release.

icemac:
It is Python 3.11 compatible.

~~This also requires https://github.com/zopefoundation/AccessControl/pull/121 to be released.~~ done


dataflake (approved):

icemac:
Hmm, the Py310 wheel is still not used by the windows runner.
Have to investigate later.

icemac:
There is `AccessControl-5.3-cp310-cp310-win32.whl` on PyPI but the action tries to build it itself. My next idea is to update to `zc.buildout` v3 an a current `setuptools` version to see if this helps.

icemac:
I created a release for `Persistence` but AppVeyor got stuck while building the Python 3.10 wheel. Trying again using https://github.com/zopefoundation/Persistence/commit/e794644f4443c206ecccb5e006eb48ba64919452.

icemac:
Not it is missing a `BTrees` 3.10 windows wheel.

icemac:
It seems we have to keep the pip version in sync between tox.ini and `versions.cfg`.

icemac:
Now a Python 3.10 windows wheel for `zodbpickle` is missing.
Additionally `pip >= 22` does not longer support Python 3.6.

versions.cfg
```diff
@@ -46,6 +46,7 @@ mr.developer = 2.0.1
 nose = 1.3.7
 packaging = 21.3
 pathlib2 = 2.3.6
+# pip >= 22 no longer supports Python 3.6
 pip = 21.3.1
```
icemac, line 50: This version pin actually hurts:

* If set to a higher version Python 3.6 breaks.
* If kept at the currently used version `pip` has to be pinned in `tox.ini` because otherwise the buildout run breaks because it is not able to downgrade to this `pip` version.

Could we try to remove this version pin?
(I tried it locally and it did not hurt.)

icemac: Okay, let's try to remove it.

icemac: Previously it did not hurt because `zc.buildout` did not use `pip`, bot version 3 does.

PR #1016, opened 2022-03-02
fix `OFS.Image.File.PUT` -- Issue 1015

### LINKED ISSUE #1015: Zope 4.x: OFS.Image PUT() method does not work on Python 2
On 4.x branch, `OFS.Image` `PUT()` method only supports Python 3 and not Python 2 because it tries to encode a `str` directly and do not use `six.text_type`.

d-maurer:
Fixes #1015.

This PR essentially documents that `OFS.Image.File._read_data` returns a tuple *data*, *size* with *data* either a `bytes` or `Pdata` instance. Based on this assertion, it can simply drop the buggy code in the `PUT` method referred to in #1015.

@arnaud-fontaine could you please verify if this PR resolves your issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,11 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.8 (unreleased)
 ----------------
 
+- Improve source documentation for methods ``_read_data`` and
+  ``get_content_type`` of ``OFS.Image.File`` and
+  fix its ``PUT``
+  (`#1015 <https://github.com/zopefoundation/Zope/issues/1015>`_).
+
 - Enhance cookie support. For details, see
   `#1010 <https://github.com/zopefoundation/Zope/issues/1010>`_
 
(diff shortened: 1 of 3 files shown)


d-maurer:
The PR now checks if `cgi` has converted to a text file and converts back to a binary file -- if necessary via copying.

dataflake:
The failure on Python 3.5/Windows is due to `cffi` version 1.15.0 no longer providing a precompiled Python 3.5 wheel. My suggestion would be pinning `cffi` to 1.14.6 on the `4.x` branch.

arnaud-fontaine:
@d-maurer Thank you so much for your prompt handling of this issue. I have just got the results from our Unit Tests and everything pass with your fix.

dataflake (approved):
I added the cffi pin.

d-maurer:
Potentially, a similar fix is necessary for `master` as well (even though `master` is Python 3 only). This is the case when we must support "PUT" for large files (larger than 64 kB) without a "Content-Disposition" header with "filename" parameter. "PUT" is very rare apart from its use by WebDAV. Likely, this means that the fix is necessary should WebDAV use "PUT" without "Content-Disposition" header.

The "PUT" for other types might be affected in similar ways. Should such a problem occur, we know how to fix it.

PR #1018, opened 2022-03-10
Add pin for last version of importlib-resources that works on Python 2.

mauritsvanrees:
The previous commit updated `tqdm` (some command line progress bar thingie) to 4.63.0, which has `importlib-resources` as new requirement. This was not pinned yet. So Plone coredev 5.2 locally failed on Python 2.7 with the latest Zope 4.x dev versions.

Also added pin for `singledispatch`, which is required by `importlib-resources`. Latest version seems fine.

Note: I am also fixing this temporarily in Plone coredev. But since these packages are pulled in by `tqdm` which is already pinned in Zope, it makes sense to add the pins here.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -52,6 +52,8 @@ idna = 2.10
 imagesize = 1.3.0
 # tox and pluggy require importlib-metadata <1
 importlib-metadata = 0.23
+# tqdm 4.63+ requires importlib-resources, and 3.3.1 is the last that works on Python 2.
+importlib-resources = 3.3.1
 lxml = 4.8.0
 manuel = 1.10.1
 # Version 6+ needs Python 3.x
@@ -79,6 +81,7 @@ readme-renderer = 28.0
 requests = 2.27.1
 requests-toolbelt = 0.9.1
 scandir = 1.10.0
+singledispatch = 3.7.0
 snowballstemmer = 2.2.0
 # soupsieve 2 needs Python 3
 soupsieve = 1.9.6


dataflake:
I'm surprised, there were no issues when I re-ran the buildout after updating the `tqdm` pin.

I also wonder if there's some automated resource that can identify version pins for stuff that isn't needed anymore. The 4.x branch has become a nightmare with those version pins and is falling behind more and more as packages drop Python 2.7 and/or 3.5 support.

IMHO we have been too "good" in backporting every little thing from `master` even though 4.x is officially receiving security fixes only. We need to be more strict in what we port. 4.x is on minimal life support and that should be made more obvious. I'd be in favor of creating a 5.x branch from current master in the very near future and then doing a major Python version support cleanup on master/6.x. Matter of fact, the current work to support Python 3.11 should be 6.x only.

dataflake (approved):

mauritsvanrees:
> I'm surprised, there were no issues when I re-ran the buildout after updating the `tqdm` pin.

`tqdm` get pulled in by `twine`, which is a dependency of `zest.releaser`. You want to pin that, but it is likely not used in GitHub Actions.
With Python2.7, after I `pip install twine`, a growing list of packages is installed, including the above:

```
$ bin/pip list
Package             Version
------------------- -----------
bleach              3.3.1
certifi             2021.10.8
chardet             4.0.0
contextlib2         0.6.0.post1
docutils            0.18.1
idna                2.10
importlib-resources 3.3.1
packaging           20.9
pathlib2            2.3.7.post1
pip                 20.3.4
pkginfo             1.8.2
Pygments            2.5.2
pyparsing           2.4.7
readme-renderer     30.0
requests            2.27.1
requests-toolbelt   0.9.1
scandir             1.10.0
setuptools          44.1.1
singledispatch      3.7.0
six                 1.16.0
tqdm                4.63.0
twine               1.15.0
typing              3.10.0.0
urllib3             1.26.8
webencodings        0.5.1
wheel               0.37.1
zipp                1.2.0
```

> I also wonder if there's some automated resource that can identify version pins for stuff that isn't needed anymore. The 4.x branch has become a nightmare with those version pins and is falling behind more and more as packages drop Python 2.7 and/or 3.5 support.

That would be useful yes, also for Plone.
In Plone 6 (alpha) I might just delete all versions and start with the very latest, and see if any removed versions make sense to keep.

In Plone 5.2, if Package 3 is for Python 3 only, and Package 2 is for Python 2 and 3, we have usually opted to have different version pins: `Package = 2` in `[versions:python27]` and `Package = 3` in the default `[versions]` or in a few cases in a specific `[versions:python3]`. That is a way to not fall behind in package versions on Python 3.

> IMHO we have been too "good" in backporting every little thing from `master` even though 4.x is officially receiving security fixes only. We need to be more strict in what we port. 4.x is on minimal life support and that should be made more obvious. I'd be in favor of creating a 5.x branch from current master in the very near future and then doing a major Python version support cleanup on master/6.x. Matter of fact, the current work to support Python 3.11 should be 6.x only.

Development on Plone 5.2, which needs Zope 4, is slowing down, so there is less need from that site for newer versions, or at least less for newer features. Bug-fix-only may be fine around now.

icemac:
> I'd be in favor of creating a 5.x branch from current master in the very near future and then doing a major Python version support cleanup on master/6.x. Matter of fact, the current work to support Python 3.11 should be 6.x only.

Additionally this could allow us to drop support for Python 3.6 those sunset was in December 2021. I am not sure how we would handle back ports from `master` to 5.x than. Maybe we find a way which has less effort than the one current one used for supporting 4.x

icemac:
Additionally: Do we really need to pin `zest.releaser`? Yes, we install it in the `scripts` part of `buildout.cfg`, but is it actually used? (I personally use a globally installed `zest.releaser` as I use it for all my projects, so there is no need to have it in each project separately.)

dataflake:
I personally don't use it at all. It seems using zest.releaser is a matter of personal preference. It certainly is not in any way important for the Zope package itself. I would be all for a big spring cleaning where everything that has been added to the buildout configuration for convenience is removed.

PR #1026, opened 2022-04-01
backport waitress version pins to Zope 4

dataflake:
This PR backports new Python version-dependent waitress version pins to the 4.x branch.

I have not ported the requirements file generator script as that's completely incompatible with Python 2 now and I have removed its invocation in `buildout.cfg`. I am inclined to remove that script on this branch and manually maintain the requirements files instead, that seems less work than porting the convenience script.


icemac (approved):
LGTM.

PR #1028, opened 2022-04-13
Drop convenience scripts from buildout configuration

### LINKED ISSUE #1019: Remove zest.releaser and other scripts from buildout.cfg
It does not seem to be used and its dependencies bloat `versions.cfg`.

Tasks:

- [x] remove not really necessary dependencies from the `[scripts]` section in `buildout.cfg`
- [x] remove the version pins of the removed dependencies and their dependencies from `versions.cfg` (Hint: remove all pins in `versions.cfg`, rerun buildout with `allow-picked-versions=true` and `show-picked-versions=true` + add the listed pinnings back to `versions.cfg`.)
- [x] Look over the documentation under `docs` and add or change places that mention these convenience tools, users should be instructed to install them manually into their virtual environment.

dataflake:
Fixes #1019 

This cuts down on a lot of version pins that make our life hard.


d-maurer (approved):

mauritsvanrees (approved):
LGTM.

The only doubt I have is for `wheel`, which I always add to `requirements.txt` in my projects so I am sure that buildout can install from wheels. But I have never used the `bin/wheel` script itself. So should be fine.

dataflake:
> The only doubt I have is for `wheel`, which I always add to `requirements.txt` in my projects so I am sure that buildout can install from wheels. But I have never used the `bin/wheel` script itself. So should be fine.

I believe the only reason `wheel` was explicitly installed was the ability for maintainers to create the release wheel, it's not about having the script around. That's why I added the documentation update.

PR #1032, opened 2022-04-14
quote all components of redirect URL (not only path)

### LINKED ISSUE #1027: `ZPublisher.HTTPResponse.redirect` should handle `netloc` (likely even all URL components) similar to `path`
Under Python 3, Zope 4 and 5 automatically quote the path component of the `redirect` URL; this way special characters in paths are automatically treated correctly. However, special characters in the netloc URL component are not quoted; they are passed unchanged to the HTTP component which (likely) reports a "network protocol violation".

With wider use of international domain names, special characters in the `netloc` will become more frequent and likely should be supported similar to `path`.

Probably, not only `path` and `netloc` but all URL components should get automatically quoted.

d-maurer:
Fixes #1027.

All components of a redirect URL are properly quoted (not just the path component).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.8.2 (unreleased)
 ------------------
 
+- Quote all components of a redirect URL (not only the path component)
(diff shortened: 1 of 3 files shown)


dataflake (approved):
This is really nice, I like it.

PR #1039, opened 2022-05-03
Allow switching to full form in "manage_access"

jan-jockusch:
When the number of roles raises above 30, the form "manage_access" switches to a compact mode with no way of getting the full matrix view back.

While all functionality is still available, the compact form does not offer a full overview of permission settings.

Therefore, I believe an option to switch back to the full matrix view would be more user-friendly.

This PR provides this by adding a hyperlink with a request variable set to force the full permission matrix form even with more than 30 roles.

The behavior snaps back to normal on each click. The setting is not stored in the client or the session.


icemac:
@jan-jockusch Thank you for the PR. Could you please share a screenshot? I currently have no project at hand with so many roles.

jan-jockusch:
Here are two shots of a form using this.

The first highlights the new hyperlink, the second one is just the standard form, as rendered with the new request option.

2022-05-11_09-30-shortened-form

2022-05-11_09-31-full-matrix

icemac:
@jan-jockusch This looks nice. Could you please add an entry in the change log to make it perfect?

jan-jockusch:
Thanks for reviewing this! Changelog entry added.

jan-jockusch:
Thanks!

Issue #1033, opened 2022-04-15
Cookie parameter "domain" might be wrongly encoded (for non-ASCII-only domains)


d-maurer:
HTTP specifies "ISO-8859-1" as charset to be used in its headers; however, recently non-ASCII characters are deprecated altogether. This poses the question how to encode non-ASCII material in those headers.

There are many competing standards to encode non-ASCII as ASCII, even in the context of HTTP. Often, the relevant specifications do not answer the question which encoding should be used.

Originally, domain names have been ASCII-only; but then international domain names have been allowed together with the IDNA standard which specifies how international domain names can be ASCII encoded. Implementing the cookie enhancement, I chose this encoding for the `domain` parameter. This choice might have been wrong.
Working on #1027, I recognized that modern browsers might instead expect the usual UTF-8 than URL-quote  encoding (which is significantly simpler than IDNA) also for the `domain` cookie parameter.

I have no means to determine which is the "correct" (i.e. supported by modern browser) encoding.

PR #1041, opened 2022-05-17
Strip leading `.` from cookie domain names.

icemac:
If the value of the `domain` attribute in the cookie starts with `.` Zope currently breaks.
But according to https://www.rfc-editor.org/rfc/rfc6265#section-4.1.2.3 it should be allowed.

I added a test and a fix, without the fix the new test breaks with:

```
Error in test test_domain (ZPublisher.tests.test_cookie.CookieParameterRegistryTests)
  File "/.../lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/.../lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/.../Zope/src/ZPublisher/tests/test_cookie.py", line 150, in test_domain
    _, v = convertCookieParameter("domain", ".zope.dev")
  File "/.../Zope/src/ZPublisher/cookie.py", line 152, in convert
    return nn, (value if value is None else self[nn](value))
  File "/.../Zope/src/ZPublisher/cookie.py", line 249, in domain_converter
    value = ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
  File "/.../Zope/src/ZPublisher/cookie.py", line 249, in <genexpr>
    value = ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
  File "/.../lib/python3.7/encodings/idna.py", line 73, in ToASCII
    raise UnicodeError("label empty or too long")
UnicodeError: label empty or too long
```

### DIFF
--- a/src/ZPublisher/cookie.py
+++ b/src/ZPublisher/cookie.py
@@ -242,7 +242,14 @@ def domain_converter(value):
         return value
     from encodings.idna import ToASCII
     from encodings.idna import nameprep
-    return ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
+    add_leading_dot = False
+    if u_value.startswith("."):
+        u_value = u_value[1:]
+        add_leading_dot = True
+    value = ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
+    if add_leading_dot:
+        value = f".{value}"
+    return value
 
 
 registerCookieParameter("Domain", domain_converter)
--- a/src/ZPublisher/cookie.py
+++ b/src/ZPublisher/cookie.py
@@ -242,8 +242,9 @@ def domain_converter(value):
         return value
     from encodings.idna import ToASCII
     from encodings.idna import nameprep
+
     # According to https://www.rfc-editor.org/rfc/rfc6265#section-4.1.2.3 a
-    # leading dot is ignored. If it is there `ToASCII` breaks on the empty
+    # leading dot is ignored. If it is there `ToASCII`, breaks on the empty
     # string:
     u_value = u_value.lstrip('.')
     return ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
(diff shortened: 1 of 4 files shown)


dataflake (approved):

d-maurer (approved):

src/ZPublisher/cookie.py
```diff
@@ -242,6 +242,10 @@ def domain_converter(value):
         return value
     from encodings.idna import ToASCII
     from encodings.idna import nameprep
```
icemac, line 244: @d-maurer Is it necessary to do these imports here instead of at the top of the module?

dataflake: I don't see a good reason for that, just move them to the top.

src/ZPublisher/cookie.py
```diff
@@ -242,7 +242,14 @@ def domain_converter(value):
...
-    return ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
+    add_leading_dot = False
+    if u_value.startswith("."):
+        u_value = u_value[1:]
+        add_leading_dot = True
+    value = ".".join(to_str(ToASCII(nameprep(c))) for c in u_value.split("."))
+    if add_leading_dot:
+        value = f".{value}"
```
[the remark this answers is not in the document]
icemac: Fixed in https://github.com/zopefoundation/Zope/pull/1041/commits/39617784c47bf1695cca6d164a500ab789af0599.

d-maurer:
Michael Howitz wrote at 2022-5-17 23:25 -0700:
***@***.*** commented on this pull request.
>
>
>
>> @@ -242,6 +242,10 @@ def domain_converter(value):
>         return value
>     from encodings.idna import ToASCII
>     from encodings.idna import nameprep
>
***@***.*** Is it necessary to do these imports here instead of at the top of the module?

They are here for documentation/modularity purposes:
this documents that they are special to this converter and not used
elsewhere.


Maybe, we should consider
"https://github.com/zopefoundation/Zope/issues/1033":
meanwhile, I believe that using the IDNA encoding (in this place) is wrong
altogether and that it should get replaced by normal URL quoting
(with safe characters defined by
"https://datatracker.ietf.org/doc/html/rfc3986.html#section-3.2.2").
IDNA is likely necessary only in interactions with the DNS;
Browsers apparently do not expect IDNA encoding in the `host` part
of an URL but normal URL quoting.

dataflake:
I'm going to merge this, #1033 is a separate issue for a separate PR.

Issue #1042, opened 2022-05-19
Improvements on find_bad_templates()


wlang42:
`Products.PageTemplate.Extensions.find_bad_templates()` is a very handy tool!

However, it would be great to have two improvements:

1. It should be able to check Filesystem Page Templates too

   ie set obj_metatypes from ('Page Template',) to 
   ('Page Template', 'Filesystem Page Template')

2. It should display HTML Tags on "Page Template scan results" page

Currently the web report looks like this:

```
form.pt

Compilation failed

chameleon.exc.ParseError: Unexpected end tag.

 - String:     ""
 - Filename:   /bachtest/form.pt
 - Location:   (line 13: col 25)
 - Source:     WEITERE_KONTAKTPERS
                                  ^^^^^^
```
But it should look like this:

```
form.pt

Compilation failed

chameleon.exc.ParseError: Unexpected end tag.

 - String:     "</DIV>"
 - Filename:   /bachtest/form.pt
 - Location:   (line 13: col 25)
 - Source:     WEITERE_KONTAKTPERS</DIV>
                                  ^^^^^^
```
ie, just quote `<` to `&lt;` in the error string
```
errs = [err.replace('<', '&lt;') for err in pt._v_errors]
```

I will make a pull request...

PR #1043, opened 2022-05-19
improvements on find_bad_templates  (#1042)

wlang42:
add FS PageTemplates, show html tags

### DIFF
--- a/src/Products/PageTemplates/Extensions/find_bad_templates.py
+++ b/src/Products/PageTemplates/Extensions/find_bad_templates.py
@@ -46,8 +46,7 @@ def find_bad_templates(self):
             pt.pt_macros()
         except PTRuntimeError:
             # html quote "<" characters to be displayed as such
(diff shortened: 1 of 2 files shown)


src/Products/PageTemplates/Extensions/find_bad_templates.py
```diff
@@ -43,7 +45,9 @@ def find_bad_templates(self):
         try:
             pt.pt_macros()
         except PTRuntimeError:
-            pt_errors[pt_path] = pt._v_errors
+            # html quote "<" characters to be displayed as such
+            errs = [err.replace('<', '&lt;') for err in pt._v_errors]
+            pt_errors[pt_path] = errs
```
dataflake, line 50:
```suggestion
            pt_errors[pt_path] = [err.replace('<', '&lt;') for err in pt._v_errors]
```

dataflake (changes requested):
No need for the intermediary variable

dataflake (approved):

PR #1045, opened 2022-05-22
Scripts/Signature: keep setting func_code and func_defaults

perrinjerome:
`_setFuncSignature` was accidentally setting `__code__` and
`__defaults__` twice.

On Zope 4 it should also support the old python 2 names `func_code`
and `func_defaults`, because Zope 4 supports python 2

Fixes https://github.com/zopefoundation/Products.PythonScripts/issues/55


leorochael (approved):
Nice catch!

The conversion with 2to3 duplicated the `__code__` and `__defaults__` assignments on the same line.

That should've been noticed before.

dataflake (approved):

PR #1046, opened 2022-05-31
Update waitress to version 2.1.2

ale-rt:
See the changelog https://docs.pylonsproject.org/projects/waitress/en/latest/index.html#change-history


ale-rt:
Thanks for the reviews!

PR #1055, opened 2022-08-03
Remove requirements badge

### LINKED ISSUE #1051: Drop Python 3.6 support to get rid of `requirements insecure`?
Our README has a badge from https://requires.io.

This service thinks that our requirements are insecure:

Bildschirmfoto 2022-07-14 um 15 23 32

The reason seems to be that it sees we are depending on `waitress == 2.0.0` which has a security vulnerability. But this is only true for Python 3.6 which is not supported by newer versions of `waitress`.

Source: https://requires.io/github/zopefoundation/Zope/requirements/?branch=master

Would this be a reason to drop Python 3.6 support or should we drop that badge?

icemac:
It is not telling the truth for Python versions > 3.6.

Fixes #1051.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,10 +16,6 @@
         :target: https://pypi.org/project/Zope/
         :alt: Supported Python versions
 
-.. image:: https://requires.io/github/zopefoundation/Zope/requirements.svg?branch=master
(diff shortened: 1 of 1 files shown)


dataflake (approved):
👍

icemac:

PR #1056, opened 2022-08-20
Update to Products.ZCatalog 6.3

davisagli:


dataflake (approved):
This is OK, but we normally handle updating dependencies after a release so that there's some time to discover issues.

davisagli:
@dataflake Got it; that seems sensible.

PR #1059, opened 2022-09-29
Testing: make `publish` utility functions handle PATH_INFO

### LINKED ISSUE #1058: `Testing.ZopeTestCase.FunctionalTestCase.publish` can not be used with %-encoded path or unicode paths.

`Testing.ZopeTestCase.FunctionalTestCase.publish` can not be used with %-encoded or unicode paths.

### What I did:

I wrote a test for a Zope application, in this application some request paths are using non-ascii characters, I was not able to use `ZopeTestCase.FunctionalTestCase.publish` to test such requests.

A minimal reproduction would be something like this:

```py
# coding: utf-8
from Testing import ZopeTestCase


class MinimalReproductionTest(ZopeTestCase.FunctionalTestCase):
    def afterSetUp(self):
        self.folder_path = '/' + self.folder.absolute_url(1)
        self.folder.addDTMLMethod('täst', file=b'test')

    def testPublishDocumentNonAscii(self):
        response = self.publish(self.folder_path + '/täst')
        #   File "Zope/src/ZPublisher/WSGIPublisher.py", line 336, in publish_module
        #     path_info = path_info.decode('utf-8')
        # UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 17: invalid continuation byte
        self.assertEqual(response.getStatus(), 200)

    def testPublishDocumentNonAsciiUrlEncoded(self):
        response = self.publish(self.folder_path + '/t%C3%A4st')
        self.assertEqual(response.getStatus(), 200)
        # AssertionError: 404 != 200
```

### What I expect to happen:

Tests pass

### What actually happened:

Both tests failed

This is on Zope master on python 3.7.7

`testPublishDocumentNonAsciiUrlEncoded` never worked,
`testPublishDocumentNonAscii` was OK on Zope 4 and python 2

## Some more details from investigating

`publish(path, ...)` acts like a WSGI server, it prepares the environment and then calls `publish_module` (which is Zope's WSGI application), but because it does not do encode the `PATH_INFO` in environment like a WSGI server is supposed to do.

wsgiref does [this](https://github.com/python/cpython/blob/dd53b79de0ea98af6a11481217a961daef4e9774/Lib/wsgiref/simple_server.py#L80-L86):

```py
        if '?' in self.path:
            path,query = self.path.split('?',1)
        else:
            path,query = self.path,''

        env['PATH_INFO'] = urllib.parse.unquote(path, 'iso-8859-1')
        env['QUERY_STRING'] = query
```

waitress does [something similar](https://github.com/Pylons/waitress/blob/73fe701cba0b37dc6099858b064cb1e755e83e9e/src/waitress/parser.py#L376).

I believe we can change `ZopeTestCase.FunctionalTestCase.publish` to also change url-decode the path decode the bytes as iso-8859-1. I will send a pull request soon.

perrinjerome:
Fixes #1058

This applies to `Testing.ZopeTestCase.FunctionalTestCase.publish` and `html` from zopedoctest, both were using Zope's WSGI application, but were not preparing `PATH_INFO` in environ like a WSGI server is supposed to do.

As a result, it was not possible to use these utility with %-encoded paths or paths with non-ascii characters.

This fixes the issue by encoding the PATH_INFO and also changes the internals a bit not to pass a request, but using the request factory, so that the request environ is also set by WSGIPublisher, which does its own handling of encoded paths.

### DIFF
--- a/src/Testing/ZopeTestCase/functional.py
+++ b/src/Testing/ZopeTestCase/functional.py
@@ -17,6 +17,7 @@ After Marius Gedminas' functional.py module for Zope3.
 
 import sys
 from functools import partial
+from urllib.parse import unquote_to_bytes
 
 import transaction
 from Testing.ZopeTestCase import interfaces
@@ -80,13 +81,11 @@ class Functional(sandbox.Sandboxed):
         env['SERVER_PROTOCOL'] = 'HTTP/1.1'
         env['REQUEST_METHOD'] = request_method
 
-        p = path.split('?')
-        if len(p) == 1:
-            env['PATH_INFO'] = p[0]
-        elif len(p) == 2:
-            [env['PATH_INFO'], env['QUERY_STRING']] = p
-        else:
-            raise TypeError('')
+        query = ''
+        if '?' in path:
+            path, query = path.split("?", 1)
+        env['PATH_INFO'] = unquote_to_bytes(path).decode('latin-1')
+        env['QUERY_STRING'] = query
 
         if basic:
             env['HTTP_AUTHORIZATION'] = basic_auth_encode(basic)
@@ -97,12 +96,16 @@ class Functional(sandbox.Sandboxed):
 
         if stdin is None:
             stdin = BytesIO()
+        env['wsgi.input'] = stdin
 
         outstream = BytesIO()
         response = WSGIResponse(stdout=outstream, stderr=sys.stderr)
-        request = Request(stdin, env, response)
-        for k, v in extra.items():
-            request[k] = v
+
+        def request_factory(*args):
+            request = Request(*args)
+            for k, v in extra.items():
+                request[k] = v
+            return request
 
         wsgi_headers = BytesIO()
 
@@ -120,7 +123,10 @@ class Functional(sandbox.Sandboxed):
             wsgi_headers.write(headers)
             wsgi_headers.write(b'\r\n\r\n')
 
-        publish = partial(publish_module, _request=request, _response=response)
+        publish = partial(
+            publish_module,
+            _request_factory=request_factory,
+            _response=response)
         if handle_errors:
             publish = HTTPExceptionHandler(publish)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Change functional testing utilities to support percent encoded and unicode
+  paths (`#1058 <https://github.com/zopefoundation/Zope/issues/1058>`_).
+
 
 5.6 (2022-09-09)
 ----------------
--- a/src/Testing/ZopeTestCase/testFunctional.py
+++ b/src/Testing/ZopeTestCase/testFunctional.py
@@ -60,6 +60,9 @@ class TestFunctional(ZopeTestCase.FunctionalTestCase):
         # A method changing the title property of an object
         self.folder.addDTMLMethod('change_title', file=CHANGE_TITLE_DTML)
 
+        # A method with a non-ascii path
+        self.folder.addDTMLMethod('täst', file=b'test')
+
     def testPublishFolder(self):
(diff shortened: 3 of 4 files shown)


icemac:
@d-maurer You know about encodings of URLs, so could you please review this PR?

icemac:
@perrinjerome Thank you for this PR, could you please add a change log entry referring to the issue you wrote?

perrinjerome:
Thanks, before review please also make sure to read the [issue](https://github.com/zopefoundation/Zope/issues/1058) description, especially the "Some more details from investigating" section. Basically these changes consist in doing in `publish` what waitress also does.

src/Testing/ZopeTestCase/functional.py
```diff
@@ -80,13 +81,11 @@ def publish(self, path, basic=None, env=None, extra=None,
...
-        if len(p) == 1:
-            env['PATH_INFO'] = p[0]
-        elif len(p) == 2:
-            [env['PATH_INFO'], env['QUERY_STRING']] = p
-        else:
-            raise TypeError('')
+        query = ''
+        if '?' in path:
+            path, query = path.split("?", 1)
+        env['PATH_INFO'] = unquote_to_bytes(path).decode('latin-1')
```
d-maurer, line 87: In my opinion, it is conceptually wrong to "unquote" here: the path typically has a substructure; unquoting the complete path (rather than the individual components) can lose information necessary to correctly identify the components (e.g. in the case of quoted "/" or ";" to indicate their use as literal part of a component rather than as separator).
However, the handling is wrong in other `Zope` parts; thus, it is likely acceptable to be wrong here, too.

d-maurer (approved):

src/Testing/ZopeTestCase/functional.py
perrinjerome: Thanks for the feedback @d-maurer 

I agree with this point. During testing, I also noticed that it was not possible in Zope (not only in the testing utilities but also with real requests) to use a path such as `a/b/c%2Fd` and have it traversed as `['a', 'b', 'c/d'] - but this is not only Zope doing it wrong here, this is also in the WSGI spec.

Apparently, ASGI has [`raw_path`](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) for that use case. Some WSGI servers put the raw URI or path in environment ( for example https://github.com/Pylons/waitress/issues/346 ). To be correct, Zope could try to use these raw URI instead of the standard `PATH_INFO`, but in practice even if the current way is not correct, I guess it's not a problem affecting many people.

As you said, it's not only this part, with these changes we are making the testing utilities have a consistent behavior with the real requests, you have approved the changes, so I am going to merge. Thanks again !

I would also appreciate feedback on https://github.com/zopefoundation/Zope/pull/1060 . Does it make sense to backport these kinds of fixes on Zope4 ? if it's in Zope4 I would use it, otherwise I would use a monkey patch in my application.

d-maurer:
Jérome Perrin wrote at 2022-10-3 23:50 -0700:
> ...
>I would also appreciate feedback on https://github.com/zopefoundation/Zope/pull/1060 . Does it make sense to backport these kinds of fixes on Zope4 ? if it's in Zope4 I would use it, otherwise I would use a monkey patch in my application.

I think the standard procedure is to first get a PR for either
`master` or `4.x` and once this is finished/merged
backport/forward port to the other branch.

In my opinion, it is good to have consitent behavior between `4.x`
and `master`.

dataflake:
> I think the standard procedure is to first get a PR for either `master` or `4.x` and once this is finished/merged backport/forward port to the other branch. In my opinion, it is good to have consitent behavior between `4.x` and `master`.

This is not strictly true anymore. Zope 4 has really been in bugfix mode only. Features are not ported to Zope 4 at this point. We are trying to get people off Zope 4 and onto Zope 5 to reduce the maintenance burden, see https://www.zope.dev/developer/roadmap.html.

perrinjerome:
Thanks, my understanding is that this is more a bug fix than a new feature, so it's no problem to backport the changes at this point.

PR #1063, opened 2022-10-07
ZPublisher/utils: fix encoding in basic_auth_encode and basic_auth_decode

### LINKED ISSUE #1061: ZPublisher.utils.basic_auth_decode decodes with latin 1

`basic_auth_decode` is used in the tests and also used in [`Request._authUserPW`](
https://github.com/zopefoundation/Zope/blob/12c78aa402716611537c10e5f7d24b039705a01e/src/ZPublisher/HTTPRequest.py#L1559) that is used in [`PluggableAuthService`'s `HTTPBasicAuthHelper.extractCredentials`](https://github.com/zopefoundation/Products.PluggableAuthService/blob/c149e92d7a29f7c03fca1a7f3e3af0434e3a8c81/src/Products/PluggableAuthService/plugins/HTTPBasicAuthHelper.py#L65-L69). 

`basic_auth_decode` is decoded as `latin-1` here:
https://github.com/zopefoundation/Zope/blob/1a81aaa1578b674a2fb856e1090b82705b50f6ab/src/ZPublisher/utils.py#L101

On the other hand, the  "default" user folder from `AccessControl` has it's own `identify` method which uses a `decode` without argument (which uses `utf-8` by default)
https://github.com/zopefoundation/AccessControl/blob/master/src/AccessControl/userfolder.py#L123-L124
```py
                name, password = decodebytes(auth.split(b' ')[-1]) \
                    .decode().split(':', 1)
```

I researched the theory a bit, and I found that the charset of "Basic" authentication header is unspecified, https://www.rfc-editor.org/rfc/rfc7617#section-2 reads:

>   For backwards compatibility reasons, this
>   specification continues to leave the default encoding undefined, as
>   long as it is compatible with US-ASCII (mapping any US-ASCII
>   character to a single octet matching the US-ASCII character code)

According to https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_auth_handler_basic.cc;l=90-98;drc=8d399817282e3c12ed54eb23ec42a5e418298ec6 popular browsers use UTF-8.

requests use latin-1 but there's an issue https://github.com/psf/requests/issues/4564 

I don't think it makes sense to use `latin-1` here and I believe it would be better to use the default encoding, like it's the case in `AccessControl`. I'm planning to send a pull request, but probably not today ( I plan to change also `basic_auth_encode`, for the same reasons )

### What I did:

1. create a user with login `é` password `é` in the root `acl_users`
2. log in with user `é` in `/manage_main`
3. create a folder `/folder`, create a Products.PluggableAuthService in that folder, create a user with login `à` password `à` in  `/folder/acl_users`
4. login with user `à` in `/folder/manage_main`

### What I expect to happen:

Both log in should work

### What actually happened:

Only 2 works, 4 did not work.

I tried this with 
```
Zope = 4.5.1 # but it is same on master
Products.PluggableAuthService = 2.7.1
Products.StandardCacheManagers = 4.1.1
Products.Sessions = 4.14
Products.PluginRegistry = 1.10
Products.GenericSetup = 2.2.0
```

I used Chrome OS with Chrome 107 as browser

perrinjerome:
Fixes #1061

Basic authentication header charset is unspecified, but in practice these days user agents use UTF-8.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 - Change functional testing utilities to support percent encoded and unicode
   paths (`#1058 <https://github.com/zopefoundation/Zope/issues/1058>`_).
 
+- Decode basic authentication header as utf-8, not latin1 anymore
+  (`#1061 <https://github.com/zopefoundation/Zope/issues/1061>`_).
+
 
 5.6 (2022-09-09)
 ----------------
--- a/src/ZPublisher/tests/testHTTPRequest.py
+++ b/src/ZPublisher/tests/testHTTPRequest.py
@@ -878,6 +878,19 @@ class HTTPRequestTests(unittest.TestCase, HTTPRequestFactoryMixin):
         self.assertEqual(user_id_x, user_id)
         self.assertEqual(password_x, password)
 
+    def test__authUserPW_non_ascii(self):
+        user_id = 'usèr'
+        password = 'pàssword'
+        auth_header = basic_auth_encode(user_id, password)
+
+        environ = {'HTTP_AUTHORIZATION': auth_header}
+        request = self._makeOne(environ=environ)
+
+        user_id_x, password_x = request._authUserPW()
+
+        self.assertEqual(user_id_x, user_id)
+        self.assertEqual(password_x, password)
+
     def test_debug_not_in_qs_still_gets_attr(self):
         from zope.publisher.base import DebugFlags
 
--- a/src/ZPublisher/utils.py
+++ b/src/ZPublisher/utils.py
@@ -85,8 +85,8 @@ def basic_auth_encode(user, password=None):
     value = user
     if password is not None:
         value = value + ':' + password
-    header = b'Basic ' + base64.b64encode(value.encode('latin-1'))
-    header = header.decode('latin-1')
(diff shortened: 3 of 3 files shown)


icemac:
@d-maurer Could you please review this PR? – I think you have way more expertise here than me.

d-maurer (approved):
The modern specification of "Basic Authentication" is RFC7617 ("https://www.rfc-editor.org/rfc/rfc7617"). It notes that the charset used to encode the user-password pair into a sequence of octets is undefined, but that usually either  a locale specific encoding (e.g. `latin-1`) or `UTF-8` has been used. For backward compatibility, it leaves `charset` undefined.

The PR changes the encoding and may therefore introduce backward incompatibility.

RFC7617 allows a server to specify the expected encoding via the `charset` parameter of the basic auth related challenge headers. Its only allowed value is `UTF-8` which indicates that the RFC expects the general use of this encoding in the long run. Thus, the PR goes into the right direction. Maybe, it should get enhanced by the use of the `charset` parameter in challenge headers. This would document the use of this encoding and allow advanced (i.e. RFC7617 aware) clients to use `UTF-8` even if they otherwise would prefer a different encoding.

perrinjerome:
Thanks @d-maurer it was also my understanding that it could be good to set the `charset` header. I started making the changes but thought it would be better to suggest the change separately. I will submit another PR with these changes. Let's merge this already since it was approved.

PR #1064, opened 2022-10-07
Zope4: ZPublisher/utils: fix encoding in basic_auth_encode and basic_auth_decode

### LINKED ISSUE #1062: Zope4: ZPublisher.utils.basic_auth_encode only supports ascii on PY2
### What I did:

I am updating a python2 application from Zope2 to Zope4 and have failures with tests using `Testing.ZopeTestCase.functional.FunctionalTestCase.publish` with a non-ascii login / password. This uses `ZPublisher.utils.basic_auth_encode`, here: https://github.com/zopefoundation/Zope/blob/337bc76e0695f09fca97527514bcf29c104e33b8/src/Testing/ZopeTestCase/functional.py#L95
 and `basic_auth_encode` is defined as https://github.com/zopefoundation/Zope/blob/337bc76e0695f09fca97527514bcf29c104e33b8/src/ZPublisher/utils.py#L86-L94

The Zope 2 version was https://github.com/zopefoundation/Zope/blob/4dfe4360931d986427171db661285f74c4e010f7/src/Testing/ZopeTestCase/functional.py#L88

The problem I noticed with `basic_auth_encode` is that it encodes the value with `latin-1`. This BTW also looks wrong also for python3 (see also https://github.com/zopefoundation/Zope/issues/1061 )

with a python2 `str`, [this `.encode('latin1')`](https://github.com/zopefoundation/Zope/blob/337bc76e0695f09fca97527514bcf29c104e33b8/src/ZPublisher/utils.py#L91) causes an `UnicodeDecodeError`:
```
(Pdb) ZPublisher.utils.basic_auth_decode(ZPublisher.utils.basic_auth_encode('h\xc3\xa9:h\xc3\xa9'))
*** UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
```

with a python2 `str`, the utf-8 string is decoded as latin1 (and as a consequence authentication fails, because in my user database the login and password are `'h\xc3\xa9`)
```
(Pdb) ZPublisher.utils.basic_auth_decode(ZPublisher.utils.basic_auth_encode('h\xc3\xa9:h\xc3\xa9'.decode('utf-8')))
('h\xe9', 'h\xe9')
```

### What I expect to happen:

On py2, there's no need to re-encode already encoded str

### What actually happened:

the login/password were wrongly decoded as latin1   

This specific problem was on python 2 / Zope 4

perrinjerome:
Fixes #1062

and backports #1063 for 4.x

Basic authentication header charset is unspecified, but in practice these days user agents use UTF-8.

On python 2 we also don't want basic_auth_encode to encode the str because we don't expect an unicode but a native string.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,11 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 - Change functional testing utilities to support percent encoded and unicode
   paths (`#1058 <https://github.com/zopefoundation/Zope/issues/1058>`_).
 
+- Decode basic authentication header as utf-8, not latin1 anymore
+  (`#1061 <https://github.com/zopefoundation/Zope/issues/1061>`_).
+
+- Make ``ZPublisher.utils.basic_auth_encode`` support non-ascii strings on Python 2
+  (`#1062 <https://github.com/zopefoundation/Zope/issues/1062>`_).
 
 4.8.2 (2022-06-01)
 ------------------
--- a/src/ZPublisher/tests/testHTTPRequest.py
+++ b/src/ZPublisher/tests/testHTTPRequest.py
@@ -1,4 +1,5 @@
 ##############################################################################
+# coding: utf-8
 #
 # Copyright (c) 2002 Zope Foundation and Contributors.
 #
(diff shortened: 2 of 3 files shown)


icemac (approved):

d-maurer (approved):
Note: you can request reviews via the `reviewers` section in the left column of the `conversation` page.

perrinjerome:
Thank you ! I'll use the "request review" feature from now on

PR #1065, opened 2022-10-12
Use UTF-8 charset for WWW-Authenticate headers in challenge responses

perrinjerome:
This tells user agents that we expect the user-pass to be UTF-8 encoded, as per section "2.1.  The 'charset' auth-param" of RFC7617.

https://datatracker.ietf.org/doc/html/draft-ietf-httpauth-basicauth-update-07#section-2.1


perrinjerome:
Following up discussion on https://github.com/zopefoundation/Zope/pull/1063#pullrequestreview-1136928618

d-maurer (approved):
Please add an entry to `CHANGES.rst` describing the change.

perrinjerome:
Thanks I added changelog entry;

> Use UTF-8 charset for WWW-Authenticate headers in challenge responses, as described in [RFC7617](https://datatracker.ietf.org/doc/html/draft-ietf-httpauth-basicauth-update-07#section-2.1) ( https://github.com/zopefoundation/Zope/pull/1065).

( I still always forget about those, sorry )

Issue #957, opened 2021-03-31
Argument conversion, add a :json converter


gbastien:
In a HTML template, we use an attribute data-filters where value is json :

`<a href="#" data-filters:json="{'to_print': true}" >`

This is used by a JS code that pass it it a view thru an ajax call.

This result in a call like `http://.../@@my_view?filters:json={"to_print":true}`

### What I expect to happen:

In the `__call__` method of `@@my_view`, I expected the `filters` argument to have been auto-magically converted to python, like it is the case when argument are suffixed with `int` or `boolean`.

### What actually happened:

I get a string representation of the json value, I need to json.load the value.

### What version of Python and Zope/Addons I am using:

This happen with Zope2/Py2.7 (Plone 4.3.x) and Zope4/Py3.8 (Plone 5.2.x)

This was discussed here : https://community.plone.org/t/zope-publisher-converter-for-json-like-it-exists-for-int/13586

What I actually would like to know is if I propose a PR for this, that would be merged?  As json is not really a `type`...

We already added a converter here : https://github.com/IMIO/imio.helpers/blob/master/src/imio/helpers/converters.py and it works, just wondered if it could get into ZPublisher...

Thank you!

Gauthier

jugmac00:
+1 from me! Please also write a test for new code.

>  As json is not really a type..

When you have a look at the following list, the name `type` does not fit anyway... no worries about that.

https://github.com/zopefoundation/Zope/blob/25119a5ac7f783b7f59cd3dcaf2519676abeee65/src/ZPublisher/Converters.py#L226-L243

d-maurer:
Gauthier Bastien wrote at 2021-3-31 06:12 -0700:
>In a HTML template, we use an attribute data-filters where value is json :
>
>`<a href="#" data-filters:json="{'to_print': true}" >`
>
>This is used by a JS code that pass it it a view thru an ajax call.
>
>This result in a call like `http://.../@@my_view?filters:json={"to_print":true}`
>
>### What I expect to happen:
>
>In the `__call__` method of `@@my_view`, I expected the `filters` argument to have been auto-magically converted to python, like it is the case when argument are suffixed with `int` or `boolean`.

I implemented something like this in `dm.zopepatches.ztutils`
via a general extension mechanism for `ZPublisher`'s `converters`
and `ZTUtils` `complex_marshal` (which is used for its `make_query`
and `make_hidden_input`).



--
Dieter

dataflake:
[...]

PR #1066, opened 2022-10-15
Added `:json` converter in `ZPublisher.Converters`.

gbastien:
See #957


gbastien:
Hi @jugmac00 @dataflake 

I finally managed to make this PR...  During the PloneConf sprint!

The code is already working for us in a monkeypatch.

Thank you for review and patience!

Gauthier

jugmac00 (approved):
Thank you for working on this.

This simple solution looks perfectly fine to me.

A quick note - I disabled auto-merge, as - while I still use packages from the Zope ecosystem, I no longer use Zope2/4/5, and I want at least a second opinion from e.g. @icemac before this will be included in Zope core.

dataflake (approved):

PR #1067, opened 2022-10-19
addzopeuser accepts custom location of zope.conf

jensens:
Script `addzopeuser` accepts now parameter '-c' or '--configuration'. This allows passing in an custom location for the `zope.conf` file to use. If not specified behavior not altered.

On you request @pbauer 

I found it was possible already to specify the path setting an environment variable `ZOPE_CONF`. But I think it is more intuitive to pass it as a parameter.

While at it and because it simplifies the code and auto-documents the script I introduced usage of argparse here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.6.1 (unreleased)
 ------------------
 
+- Script `addzopeuser` accepts now parameter '-c' or '--configuration'.
+  This allows passing in a custom location for the `zope.conf` file to use.
+  If not specified, behavior is not not altered.
+
 - Update to newest compatible versions of dependencies.
 
 - Change functional testing utilities to support percent encoded and unicode
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Script `addzopeuser` accepts now parameter '-c' or '--configuration'.
   This allows passing in a custom location for the `zope.conf` file to use.
-  If not specified, behavior is not not altered.
+  If not specified, behavior is not altered.
 
 - Update to newest compatible versions of dependencies.
 
--- a/docs/operation.rst
+++ b/docs/operation.rst
@@ -254,7 +254,12 @@ this using ``addzopeuser`` as follows:
 
   $ bin/addzopeuser user password
 
-The script expects to find the configuration file at ``etc/zope.conf``.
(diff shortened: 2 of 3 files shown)


davisagli:
I don't like the inconsistency between using a named option here while existing scripts like zconsole take it as a positional arg.

On the other hand, I like it better as a named option.

Hmm. Not sure what to do.

jensens:
I first thought about making it positional, but its so counterintuitive and ugly... 

Really like `addzopeuser joe 123t4est instance/etc/zope.conf` ?

Would need to be the last arg in order to make it optional. 
More opinions on this? I can change it...

davisagli:
@jensens No, I think it's better the way you have it.

jensens:
It looks like the windows tests are a bit flaky...

jensens:
all green, who wants to review?

CHANGES.rst
```diff
@@ -10,6 +10,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.6.1 (unreleased)
 ------------------
 
+- Script `addzopeuser` accepts now parameter '-c' or '--configuration'.
+  This allows passing in a custom location for the `zope.conf` file to use.
+  If not specified, behavior is not not altered.
```
ale-rt, line 15:
```suggestion
  If not specified, behavior is not altered.
```

docs/operation.rst
```diff
@@ -254,7 +254,12 @@ this using ``addzopeuser`` as follows:
 
   $ bin/addzopeuser user password
 
-The script expects to find the configuration file at ``etc/zope.conf``.
+The script expects to find the configuration file at ``etc/zope.conf`` by default.
+If it is located on a different location this can be specified by the `--configuration` option:
```
dataflake, line 258:
```suggestion
If it is located in a different location you can specify it with the `--configuration` option:
```

ale-rt (commented):
LGTM to me but there is a typo

src/Zope2/utilities/adduser.py
```diff
@@ -25,18 +26,26 @@ def adduser(app, user, pwd):
...
-    except IndexError:
-        print("%s <username> <password>" % argv[0])
-        sys.exit(255)
+    parser = argparse.ArgumentParser(
+        description='Add a Zope management user to the root Zope user folder.'
+    )
+    parser.add_argument(
+        "-c",
+        "--configuration",
+        help="Zope configuration file",
```
dataflake, line 35:
```suggestion
        help="Path to Zope configuration file",
```

dataflake (changes requested):

dataflake (approved):

PR #1075, opened 2022-12-14
HTTPResponse: default content type to text/plain

perrinjerome:
* drop implicit detection of content type
 * drop support for returning HTML with ``response.setBody((title, body))``

### DIFF
--- a/docs/zdgbook/ObjectPublishing.rst
+++ b/docs/zdgbook/ObjectPublishing.rst
@@ -261,25 +261,6 @@ the chapter.
   Depending on the used client, it looks like nothing happens.
 
 
-Optionally, the published method can return a tuple with
-the title and the body of the response. In this case, the publisher
-returns a generated HTML page, with the first item of the tuple used
-for the value of the HTML ``title`` tag of the page, and the second
-item as the content of the HTML ``body`` tag.
-
-
-For example a response of::
-
-  ("my_title", "my_text")
-
-
-is turned into this HTML page::
-
-  <html>
-  <head><title>my_title</title></head>
-  <body>my_text</body>
-  </html>
-
 
 Controlling Base HREF
 ---------------------
@@ -348,10 +329,10 @@ base with a *base* tag in your ``index_html`` method output.
 Response Headers
 ----------------
 
-The publisher and the web server take care of setting response headers
-such as *Content-Length* and *Content-Type*. Later in the chapter
-you'll find out how to control these headers and also how exceptions
-are used to set the HTTP response code.
+The publisher and the web server take care of setting *Content-Length*
+response header. Later in the chapter you'll find out how to control
+response headers and also how exceptions are used to set the HTTP
+response code.
 
 
 Pre-Traversal Hook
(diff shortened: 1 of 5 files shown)


perrinjerome:
references:  https://github.com/perrinjerome/buildout.coredev/commit/f288a5cefb7c2b5e8e326ca734476219f3cc1828 
 https://github.com/plone/plone.protect/commit/c03cfb4cfa9f8ab117c07bd77cd22a1afcd55acf

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -348,10 +329,10 @@ base with a *base* tag in your ``index_html`` method output.
 Response Headers
 ----------------
 
-The publisher and the web server take care of setting response headers
-such as *Content-Length* and *Content-Type*. Later in the chapter
-you'll find out how to control these headers and also how exceptions
-are used to set the HTTP response code.
+The publisher and the web server take care of setting *Content-Length*
```
dataflake, line 332:
```suggestion
The publisher and the web server take care of setting the *Content-Length*
```

dataflake (commented):
Where I don't follow yet is the removal of the old tuple behavior. Do you see a way of making such a call with a tuple that can be hijacked?

perrinjerome:
> Where I don't follow yet is the removal of the old tuple behavior. Do you see a way of making such a call with a tuple that can be hijacked?

yes, a method that returns a tuple from user data, when the tuple has length 2.

perrinjerome:
More about the `setBody((title, body))`:  The patch drops this special case of tuple with length 2, publishing methods returning tuples always use the str() representation of the tuple, served as text/plain.

I tried to change all the usages inside Zope (it was used for some error pages). I assume that there is not many places where people rely on this, building an html page with only title and body arguments does not seem common.

dataflake:
Thanks for the explanation, so it looks there are a number of methods out there returning such tuples, some of which may be callable anonymously as well.

dataflake (approved):

mauritsvanrees:
I have started a pull request on the Plone core development buildout: https://github.com/plone/buildout.coredev/pull/834
Locally I quickly saw several more errors when running only unit tests without layers, with `bin/test -u`. The gh-action test on Python 3.8 that does the same, [passes though](https://github.com/plone/buildout.coredev/actions/runs/3708156763/jobs/6285405418).

Jenkins is running now on four Python versions. I will report back. I started the 3.11 one earlier, so that should be first to finish:
https://jenkins.plone.org/job/pull-request-6.0-3.11/114

mauritsvanrees (approved):
LGTM.

All Plone 6.0 tests pass. I have merged your `plone.protect` test fix.

docs/zdgbook/ObjectPublishing.rst
```diff
@@ -348,10 +329,10 @@ base with a *base* tag in your ``index_html`` method output.
 Response Headers
 ----------------
 
-The publisher and the web server take care of setting response headers
-such as *Content-Length* and *Content-Type*. Later in the chapter
-you'll find out how to control these headers and also how exceptions
-are used to set the HTTP response code.
+The publisher and the web server take care of setting the *Content-Length*
```
d-maurer, line 332: You have removed `Content-Type` from the description; but `Content-Type` is still set just in a different way. I suggest to readd `Content-Type` to the description.

d-maurer (approved):

dataflake:
I have re-added that documentation on both `master` and `4.x`

perrinjerome:
Thank you everybody for the help here.

PR #1079, opened 2022-12-16
Fix broken ZMI pages due to the changed default content type

dataflake:
Fixes #1078 

There may be more affected pages

PR #1081, opened 2022-12-17
Serve MessageDialogs and exception views as text/html

perrinjerome:
Now that Zope uses text/plain by default, MessageDialogs were served as text/plain. Keep compatibility by returning a special string with `asHTML` method, that ZPublisher.HTTPResponse.HTTPResponse.setBody understands.

MessageDialogs are deprecated and do not integrate well in Zope >= 4 ZMI, but they are used in some old products.

Fixes issues discussed in https://github.com/zopefoundation/Zope/pull/1079#issuecomment-1356246105

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,8 +11,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.7.3 (unreleased)
 ------------------
 
-- Adjust the old ``App.Dialogs.MessageDialog`` to be served as HTML after
-  from PR https://github.com/zopefoundation/Zope/pull/1075 .
+- Explicitly serve ``App.Dialogs.MessageDialog`` and exception views as HTML
+  due to the changed default content type from `#1075
+  <https://github.com/zopefoundation/Zope/pull/1075>`_.
+
 
 5.7.2 (2022-12-17)
 ------------------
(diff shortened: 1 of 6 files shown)


perrinjerome:
image

looks odd in the new ZMI, but it's better than

image

---

This implementation is a bit tricky, it uses https://github.com/zopefoundation/Zope/blob/6aa5e4d570bc819643be5659ae962122c066ec8e/src/ZPublisher/HTTPResponse.py#L541-L544 the MessageDialogs are supposed to be returned directly ( see https://github.com/nakagami/Products.ZSQLiteDA/blob/c89c190f9572070607893ee09f627c3862370905/Products/ZSQLiteDA/DA.py#L128-L133 ), so at this level we don't have access to requests and response

/cc @dataflake @GiaBen

perrinjerome:
the test on zope 4 seems to show a regression with this approach

perrinjerome:
> the test on zope 4 seems to show a regression with this approach

It's OK now.

For the details: in the previous version I was changing `App.special_dtml.HTML` to return html and this made this DTML Document with content `index` served as text/html instead of the expected text/plain https://github.com/zopefoundation/Zope/blob/6c199cdacec0ea3b72d3510bad7071e28557c4fe/src/Testing/ZopeTestCase/zopedoctest/testFunctionalDocTest.py#L118

perrinjerome:
Thanks @dataflake for the amendments 👍

dataflake (approved):

PR #1082, opened 2022-12-17
Serve MessageDialogs and exception views as text/html (4.x)

perrinjerome:
Now that Zope uses text/plain by default, MessageDialogs were served as text/plain. Keep compatibility by returning a special string with `asHTML` method, that ZPublisher.HTTPResponse.HTTPResponse.setBody understands.

MessageDialogs are deprecated and do not integrate well in Zope >= 4 ZMI, but they are used in some old products.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ https://zope.readthedocs.io/en/2.13/CHANGES.html
 4.8.6 (unreleased)
 ------------------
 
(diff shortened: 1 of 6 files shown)


perrinjerome:
https://github.com/zopefoundation/Zope/pull/1081 backported for 4.x

dataflake (approved):

PR #1083, opened 2022-12-19
Drop Python 3.6 support

dataflake:
The changes in configuration files were done manually for right now, not with meta/config, because that will want to put Python 3.6 back in.


mauritsvanrees (commented):
Technically fine, but see my [comment](https://github.com/zopefoundation/meta/issues/177#issuecomment-1357585757) on meta about including this now.

requirements-full.txt
```diff
@@ -31,8 +30,7 @@ five.localsitemanager==3.4
 funcsigs==1.0.2
 future==0.18.2
 ipaddress==1.0.23
-jeepney==0.7.1; python_version == '3.6'
-jeepney==0.8.0; python_version > '3.6'
+jeepney==0.8.0
 mock==4.0.3
```
d-maurer, line 34: Did we not want to replace `mock` by `unittest.mock`?

d-maurer (approved):

requirements-full.txt
icemac: Zope itself does not use `mock` at all (according to GitHub search), but our dependencies might still use it.

dataflake:
> LGTM. Maybe the current meta/config version could be run on this branch, to see if there any additional differences.

I just did, the only change was removal of the `with-legacy-python` directive from `.meta.toml`. The failing tests are still macOS and python 3.10/3.9.

perrinjerome:
> The failing tests are still macOS and python 3.10/3.9.

Is this problem understood  ? 

I tried to understand by using https://mxschmitt.github.io/action-tmate/ and so far I understood that the github action uses "macos-latest" which is sometimes macos-11 and sometimes macos-12 and that the zope-interface problem happens with macos-12, but not when runner is macos-11 .

dataflake:
> > The failing tests are still macOS and python 3.10/3.9.
> 
> Is this problem understood ?
> 
> I tried to understand by using https://mxschmitt.github.io/action-tmate/ and so far I understood that the github action uses "macos-latest" which is sometimes macos-11 and sometimes macos-12 and that the zope-interface problem happens with macos-12, but not when runner is macos-11 .

Your suspicion seems correct, when I force the runners to use `macos-11` instead of `macos-latest` the tests succeed.

The problem is only partially understood. It's a combination of misbehaviors and bugs in `zc.buildout`, `setuptools` and how the GHA macos runner's Python is set up. None of the issues show in "normal" setups where Python is compiled locally to support the specific local architecture, arm64 or x86_64. I can reproduce it when I force Python to compile in `universal2` mode.

I had spent a while debugging it when we first saw it in the combination of `macos-latest` and Python 3.11. I even filed comments and issues in both the `zc.buildout` and `setuptools` projects, but honestly, I don't expect any help there. See https://github.com/buildout/buildout/issues/609#issuecomment-1317190892 and https://github.com/pypa/setuptools/issues/3687.

In the end I decided I spent too much time on supporting a weird Python setup (`universal2`) and that it's not worth it at the moment. I also did not have an Intel-based Mac available to test on. I may take another look, I am supposed to get an Intel Mac for development from a client soon.

PR #1086, opened 2022-12-23
History: support classes modifying instances in __setstate__ 

perrinjerome:
Fixes the problem described on https://bugs.launchpad.net/zope2/+bug/735999 that still happens when comparing history of `Products.PythonScripts.PythonScript` created with an old script magic or python magic.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Fix history page for classes modifying instances in ``__setstate__``,
+  such as ``Products.PythonScripts.PythonScript`` instances.
+  See `launchpad issue 735999
+  <https://bugs.launchpad.net/zope2/+bug/735999>`_.
+
 
 5.7.3 (2022-12-19)
 ------------------
--- a/src/OFS/History.py
+++ b/src/OFS/History.py
@@ -45,18 +45,25 @@ class HystoryJar:
 
     def __init__(self, base):
         self.__base__ = base
+        self._needs_to_join = True
+        self._registered_objects = []
 
     def __getattr__(self, name):
         return getattr(self.__base__, name)
 
-    def commit(self, object, transaction):
-        if object._p_changed:
-            raise TemporalParadox("You can't change history!")
+    def commit(self, transaction):
+        raise TemporalParadox("You can't change history!")
 
     def abort(*args, **kw):
         pass
 
-    register = tpc_begin = tpc_finish = abort
+    tpc_begin = tpc_finish = abort
+
+    def register(self, obj):
+        if self._needs_to_join:
+            self.transaction_manager.get().join(self)
+        if obj is not None:
+            self._registered_objects.append(obj)
 
 
 def historicalRevision(self, serial):
--- a/src/OFS/History.py
+++ b/src/OFS/History.py
@@ -62,11 +62,11 @@ class HystoryJar:
 def historicalRevision(self, serial):
     state = self._p_jar.oldstate(self, serial)
     rev = self.__class__.__basicnew__()
-    rev._p_jar = HystoryJar(self._p_jar)
     rev._p_oid = self._p_oid
     rev._p_serial = serial
     rev.__setstate__(state)
     rev._p_changed = 0
+    rev._p_jar = HystoryJar(self._p_jar)
     return rev
 
 
--- a/src/OFS/History.py
+++ b/src/OFS/History.py
@@ -56,7 +56,7 @@ class HystoryJar:
     def abort(*args, **kw):
         pass
 
-    tpc_begin = tpc_finish = abort
+    register = tpc_begin = tpc_finish = abort
 
 
 def historicalRevision(self, serial):
--- a/src/OFS/tests/testHistory.py
+++ b/src/OFS/tests/testHistory.py
@@ -10,6 +10,7 @@ import Zope2
 from OFS.Application import Application
 from OFS.History import Historical
 from OFS.SimpleItem import SimpleItem
+from Testing.makerequest import makerequest
 from ZODB.FileStorage import FileStorage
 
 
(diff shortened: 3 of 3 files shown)


CHANGES.rst
```diff
@@ -11,6 +11,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 ------------------
 
 - Update to newest compatible versions of dependencies.
+- Fix history page for classes modifying instances in ``__setstate__``,
```
dataflake, line 14:
```suggestion

- Fix history page for classes modifying instances in ``__setstate__``,
```

dataflake (approved):
If you leave out the change in `History.py` and the tests fail with the same AssertionError explained in the Launchpad issue then that seems fine. I don't have much knowledge about the History stuff, though.

perrinjerome:
> If you leave out the change in History.py and the tests fail with the same AssertionError explained in the Launchpad issue then that seems fine

I think so, on the launchpad issue, the description of the error was:

> ... it fails with an AssertionError in ZOPE/Connection.py:register on "assert obj._p_jar is self".

and without the change in `History.py` test errors like this:

```console
$ ./bin/test --test OFS.tests.testHistory
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.


Failure in test test_manage_historicalComparison (OFS.tests.testHistory.HistoryWithSetStateTest)
  File "/opt/slapos-shared/python3/6f204393301d3c4787363493d65ac803/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/slapos-shared/python3/6f204393301d3c4787363493d65ac803/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/srv/slapgrid/slappart3/srv/project/Zope/src/OFS/tests/testHistory.py", line 109, in test_manage_historicalComparison
    keys=[r[1]['key'], r[2]['key']]))
  File "/srv/slapgrid/slappart3/srv/project/Zope/src/OFS/History.py", line 192, in manage_historicalComparison
    rev1 = historicalRevision(self, serial)
  File "/srv/slapgrid/slappart3/srv/project/Zope/src/OFS/History.py", line 68, in historicalRevision
    rev.__setstate__(state)
  File "/srv/slapgrid/slappart3/srv/project/Zope/src/OFS/tests/testHistory.py", line 153, in __setstate__
    self.changed_something = True
  File "/srv/slapgrid/slappart3/srv/project/Zope/eggs/ZODB-5.8.0-py3.7.egg/ZODB/Connection.py", line 817, in register
    assert obj._p_jar is self
AssertionError

  Ran 9 tests with 1 failures, 0 errors and 0 skipped in 0.452 seconds.
Tearing down left over layers:
  Tear down zope.testrunner.layer.UnitTests in 0.000 seconds.
```

CHANGES.rst
perrinjerome: Thanks, I applied this locally and push-forced

src/OFS/History.py
```diff
@@ -56,7 +56,7 @@ def commit(self, object, transaction):
     def abort(*args, **kw):
         pass
 
-    tpc_begin = tpc_finish = abort
+    register = tpc_begin = tpc_finish = abort
```
icemac, line 59: I think this can be left out: `register` was part of an old and no longer supported API, if I remember correctly.

icemac (approved):
I do not know much about the topic either but I have a little suggestion to try.

icemac:
Additionally We need a rebase onto master when #1083 is merged to get MacOS tests fixed.

src/OFS/History.py
icemac: After reading the whole PR thread, I retract my comment. My memories seem to betray me. Time to get some fresh air.

dataflake:
I just merged master with the button here. All green, this is good to go.

perrinjerome:
> I just merged master with the button here. All green, this is good to go.

Thanks, so this means I should click merge. I will do in a few minutes /hours if no objectsion

dataflake:
When tests are green and there are no unresolved conversations or change requests and you have approvals then you can merge at any time.

d-maurer:
Michael Howitz wrote at 2022-12-23 00:52 -0800:
>
>I do not know much about the topic either but I have a little suggestion to try.
>
>> @@ -56,7 +56,7 @@ def commit(self, object, transaction):
>     def abort(*args, **kw):
>         pass
>
>-    tpc_begin = tpc_finish = abort
>+    register = tpc_begin = tpc_finish = abort

`register` is still a `ZODB.Connection.Connection` method
(like the other methods).

perrinjerome:
Thanks, I have been checking a bit more and I was not comfortable that this `raise` was not covered by the test:

https://github.com/zopefoundation/Zope/blob/a58333605a88ad9a24a7f07a820f4d837e2f7cd0/src/OFS/History.py#L52-L54

This code path has been unreachable since 2004 ( https://github.com/zopefoundation/ZODB/commit/61db515972d7a809ba2bb3588329a9b489890d0e ), I made more changes to cover this and keep the original intention of this `raise`: that attempting to commit a transaction changing old revisions of objects is an error.

To achieve this, I had to change the logic a bit. In that first version of the patch, modifying an old revision was just ignored because the object was not registered to the transaction, so this was a bit against the original idea of `TemporalParadox`.

In that new version, it is still possible to modify, but committing the transaction afterwards raises a `TemporalParadox` error.

To achieve this, I had to change `historicalRevision` to assign the `_p_jar` later, now for classes modifying instances in `__setstate__`, `historicalRevision` first create an object not associated with a database, call `__setstate__` and only then sets the `_p_jar`, this way, modification in `__setstate__` are ignored. This is actually similar to the patch that was attached to the launchpad issue.

dataflake:
I can only repeat myself: My knowledge of the history mechanism and the intricate details of the ZODB is not enough to really review this. If the tests fail without the code changes and succeed with them in place I am guessing it's OK.

dataflake (approved):

perrinjerome:
I'm merging this now, I wanted to be back from vacation before merging. Thanks again for the review and happy new year :)

Issue #1031, opened 2022-04-14
Try out pre-commit.ci


icemac:
### What I did:

I hear some complaints about the necessity of commits to make flake8 or the other code checkers happy.

### What I expect to happen:

Most of the issues could be fixed by a machine, so it should do it.

### What actually happened:

I have to run `tox` before the commit to see the problems which will arise in CI.

### Suggestions

* Configure `pre-commit` to run the code checkers.
    * This allows developers with installed and activated `pre-commit` to automatically see the problems when committing and get them fixed if they are automatically fixable.
* Configure repository to use https://pre-commit.ci to automatically run the pre-commit tests as CI system.
    * It updates the PR if it finds fixable issues raised by the code checkers.
* Use https://github.com/isidentical/teyit to convert deprecated `unittest` assert calls.

### Notes

- [ ] Configure `tox -elint` to run `pre-commit` on all files to use the same configuration and to allow to keep existing habits.
- [ ] Document how to install and activate `pre-commit`.
- [ ] GHA no longer needs to run the `lint` environment.
- [ ] This is an experiment, currently just the `Zope` repository should be part of it.

jugmac00:
Thanks, happy to see that `pre-commit` will be considered! :+1:  #891

gforcada:
For what is worth: `plone/meta` uses and benefits from it 😄

See https://github.com/plone/meta/blob/main/config/default/pre-commit-config.yaml.j2

dataflake:
This issue is obsolete, we are using pre-commit now.

PR #1093, opened 2023-01-11
Get rid of DeprecationWarnings during unit tests

dataflake:
See #1091 - Fixes at least those parts that are triggered by Zope itself.

### DIFF
--- a/src/Testing/ZopeTestCase/testZopeTestCase.py
+++ b/src/Testing/ZopeTestCase/testZopeTestCase.py
@@ -398,4 +398,4 @@ def test_suite():
         unittest.defaultTestLoader.loadTestsFromTestCase(TestPlainUserFolder),
         unittest.defaultTestLoader.loadTestsFromTestCase(
             TestWrappingUserFolder),
-        ))
+    ))
--- a/src/Testing/ZopeTestCase/testZopeTestCase.py
+++ b/src/Testing/ZopeTestCase/testZopeTestCase.py
@@ -20,6 +20,8 @@ example test cases. See testSkeleton.py for a quick
 way of getting started.
 """
 
+import unittest
+
 import transaction
 from AccessControl import getSecurityManager
 from Acquisition import aq_base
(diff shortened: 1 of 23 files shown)


d-maurer (approved):
I am fine with your changes.

However, @icemac reported also a `cgi.FieldStorage` related deprecation warning. For it, https://peps.python.org/pep-0594/#cgi says
```
FieldStorage/MiniFieldStorage has no direct replacement, but can typically be replaced by
using [multipart](https://pypi.org/project/multipart/) (for POST and PUT requests)
or urllib.parse.parse_qsl (for GET and HEAD requests)
```

dataflake:
Yes, I know about the other warnings. My goal was not to fix everything, but all those that can be fixed in the time that I have available for these tasks.

src/Zope2/App/tests/test_startup.py
```diff
@@ -74,6 +74,4 @@ def test_interface(self):
 
 
 def test_suite():
-    suite = unittest.TestSuite()
-    suite.addTest(unittest.makeSuite(StartupTests))
-    return suite
+    return unittest.defaultTestLoader.loadTestsFromTestCase(StartupTests)
```
icemac, line 77: If there are only unittests collected in the `test_suite` method, it can be completely removed as `zope.testrunner` has a mechanism to detect unittest tests automatically.

src/Testing/ZopeTestCase/testZopeTestCase.py
```diff
@@ -391,10 +393,9 @@ def testLoggedInUserIsWrapped(self):
...
-    suite.addTest(makeSuite(TestZopeTestCase))
-    suite.addTest(makeSuite(TestPlainUserFolder))
-    suite.addTest(makeSuite(TestWrappingUserFolder))
-    return suite
+    return unittest.TestSuite((
+        unittest.defaultTestLoader.loadTestsFromTestCase(TestZopeTestCase),
+        unittest.defaultTestLoader.loadTestsFromTestCase(TestPlainUserFolder),
+        unittest.defaultTestLoader.loadTestsFromTestCase(
+            TestWrappingUserFolder),
+    ))
```
icemac, line 401: This change looks like you did it manually. Do you know about the feature in `pyupgrade` which automatically does this conversion?
What a pity I did not have the time yet to work on #1031 which could have done this automatically.

icemac (approved):
Thank you for your time investment. I like the changes but tried to ensure that the issue is not closed because this does not fix all deprecation warnings.

src/Testing/ZopeTestCase/testZopeTestCase.py
dataflake: I did not use `pyupgrade`. I do it when the number of changes is very high or for simple stuff, but in this case I preferred a manual process because I also wanted to clean up how modules are imported etc.

dataflake: P.S.: I personally DO NOT like it when tools like pre-commit force changes. I want the ability to review and change if needed before anything gets committed.

src/Zope2/App/tests/test_startup.py
dataflake: It is not that simple. Removing actually ends up with **more** tests run because all of a sudden e.g. base classes get run as well. I saw that with tests that use `Testing.ZopeTestCase`. Figuring that out is a job for another day.

PR #1102, opened 2023-03-09
Accept charset in Content-Type header application/x-www-form-urlencoded.

mauritsvanrees:
Warn when this does not fit the default encoding.
This is my proposal from https://github.com/plone/buildout.coredev/pull/844#issuecomment-1462845225
It fixes several problems in Plone.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
   Fix encoding handling and ``:bytes`` converter.
 
+  Accept but ignore illegal charset in Content-Type header ``application/x-www-form-urlencoded``.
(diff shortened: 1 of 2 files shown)


d-maurer (commented):
I will create an alternative PR which fully supports the `charset` parameter of content type `application/x-www-form-urlencoded` (and maybe even a `charset` for individual form fields).
Apparently, `jquery` does not honor the specification of the above content type and creates values with `charset=utf-8`. While `Plone` might require `utf-8` as default encoding (and therefore this PR might be sufficient for `Plone`), `Zope` still supports a different default encoding. For the latter case, the current PR would be insufficient.

mauritsvanrees:
Thanks! Sounds good.

mauritsvanrees:
Closed in favour of PR #1104.

Issue #1085, opened 2022-12-20
Preparation for Python 3.12


icemac:
This issue links to the issues inside the dependencies which need to be solved in order to test Zope on Python 3.12.

- [x] https://github.com/zopefoundation/AccessControl/issues/136
- [x] https://github.com/zopefoundation/RestrictedPython/issues/246
- [x] #1103

icemac:
Tests run on Python 3.12a5, so this issue is resolved.

PR #1103, opened 2023-03-10
Test against Python 3.12a5.

icemac:
See #1085.
Using https://github.com/zopefoundation/meta/pull/195


gforcada:
Note that since 2 days there's [a new alpha out](https://www.python.org/downloads/release/python-3120a6/) 😅 

According to the [release schedule](https://peps.python.org/pep-0693/) there will be a last alpha version in a month, and then we will start the beta process.

dataflake (approved):
The change is fine, but you'll need to add an exclusion for Python 3.12/macOS, similar to the existing exclusions for newer Pythons on macOS. See https://github.com/zopefoundation/meta/issues/181

icemac:
> Note that since 2 days there's [a new alpha out](https://www.python.org/downloads/release/python-3120a6/)

I know but it is not available on GHA.

icemac:
> The change is fine, but you'll need to add an exclusion for Python 3.12/macOS, similar to the existing exclusions for newer Pythons on macOS. See [zopefoundation/meta#181](https://github.com/zopefoundation/meta/issues/181)

See https://github.com/zopefoundation/meta/pull/195 where I implemented this "fix".

icemac:

PR #1104, opened 2023-03-10
Support `charset` parameter for content type `application/x-www-form-encoded`

d-maurer:
This PR fixes a problem reported in
https://github.com/plone/buildout.coredev/pull/844#issuecomment-1461044889

Apparently, `jquery` uses not standard conform content types of the form
`application/x-www-form-urlencoded; charset=utf-8`. With the switch from `cgi.FieldStorage` to `multipart`, `Zope` no longer recognized not standard `application/x-www-form-urlencoded` content type headers. As a result, some Plone tests broke.

This PR supports the `charset` parameter for `application/x-www-form-urlencoded`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,15 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Support the (non standard) ``charset`` parameter for
+  content type ``application/x-www-form-urlencoded``.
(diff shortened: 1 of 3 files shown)


mauritsvanrees (approved):
The code seems good.  Manual testing in Plone shows the known failures are fixed.  I am running the tests on Jenkins now in this PR:
https://github.com/plone/buildout.coredev/pull/844
If that passes, it should be good to go.

Thanks a lot!

PR #1105, opened 2023-03-15
Update package versions.

mauritsvanrees:
Includes mostly major releases from Zope Foundation packages, dropping support for older Python versions.

One thing to note is that this includes `z3c.checkversions` 2.0 which says in its changelog: "Pin `setuptools < 66` as newer versions are no longer compatible."
Perhaps I should revert this version, until https://github.com/zopefoundation/z3c.checkversions/issues/24 is fixed.


icemac:
@mauritsvanrees Older versions of `z3c.checkversions` also do not support current versions of `setuptools`, so keeping an older version does not make it better.

icemac (changes requested):
Please also run buildout on your branch so `requirements.txt` and `constraints.txt` get updated.

dataflake:
In the past we have made the package updates *after* a release so that there's more time to discover incompatibilities. I thought that was a good policy.

mauritsvanrees:
@icemac wrote:
> Older versions of `z3c.checkversions` also do not support current versions of `setuptools`, so keeping an older version does not make it better.

At least older versions do not prevent the rest of the system from using a newer setuptools.  I might want to do that in Plone.  I will revert this part of the PR for now.

Oops, and I will run buildout.

@dataflake wrote:
> In the past we have made the package updates _after_ a release so that there's more time to discover incompatibilities. I thought that was a good policy.

The updates in this PR focus on using the releases from the past two months that drop Python 2 support.  Should be safe, and would be nice to have, but it is a judgement call.
I will revert a few version upgrades to make it clearer.

Regardless of merging this PR or not: I would like a Zope release.  I plan on creating a Plone 6 bugfix release at the end of next week.  All Plone tests pass with the current Zope master checkout and master versions.  Would a Zope release be possible at the latest next Wednesday?

constraints.txt
```diff
@@ -25,21 +25,21 @@ persistent==5.0
...
-roman==3.3
+roman==4.0
 six==1.16.0
 soupsieve==2.3.2.post1
 transaction==3.0.1
 waitress==2.1.2
 z3c.pt==3.3.1
 zExceptions==4.3
-zc.lockfile==2.0
+zc.lockfile==3.0.post1
```
dataflake, line 35: @icemac I notice you created that `post1` release. Please let's avoid silly version numbers like that.

dataflake (approved):

dataflake:
I can make a release before Wednesday, yes. In the future let's stick with our current policy and set new version pins after a release.

dataflake:
`z3c.checkversions` actually has more issues. Version 1.2 as currently pinned blows up on the empty Zope pin on `master` because version "" is invalid. It also blows up on the pin `zc.buildout >= 3.0.1` in `tox.ini`.

icemac:
@dataflake Could you please create an issue in `z3c.checkversions` for the problems you found?

constraints.txt
icemac: @dataflake I do not think this is a silly version number. According to https://peps.python.org/pep-0440/#post-releases these kind of version numbers can be used if the actual code did not change but there was a packaging problem. The change log of this package shows that `python_requires` was missing, which does not affect the package itself but allowed to install it on too old versions. (See https://pypi.org/project/zc.lockfile/#post1-2023-02-28.)

PR #1112, opened 2023-03-24
Implement code change suggestions from CodeQL scanning

dataflake:
First cut with just Python suggestions

### DIFF
--- a/src/Products/PageTemplates/tests/test_ptfile.py
+++ b/src/Products/PageTemplates/tests/test_ptfile.py
@@ -22,9 +22,8 @@ class TypeSniffingTestCase(unittest.TestCase):
             DefaultUnicodeEncodingConflictResolver
         from zope.component import provideUtility
 
-        self.tmpfile = tempfile.NamedTemporaryFile(suffix=".zpt", delete=False)
-        self.TEMPFILENAME = self.tmpfile.name
-        self.tmpfile.close()
+        with tempfile.NamedTemporaryFile(suffix=".zpt", delete=False) as fp:
+            self.TEMPFILENAME = fp.name
 
         # Make sure we use the new default chameleon engine
         useChameleonEngine()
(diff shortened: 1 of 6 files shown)


src/Zope2/Startup/tests/test_schema.py
```diff
@@ -43,26 +42,23 @@ class WSGIStartupTestCase(unittest.TestCase):
 
     def setUp(self):
         self.default_encoding = ZPublisher.HTTPRequest.default_encoding
+        self.TEMPNAME = tempfile.mkdtemp()
+        self.TEMPVAR = os.path.join(self.TEMPNAME, "var")
+        os.mkdir(self.TEMPVAR)
```
ale-rt, line 47: What about using a context manager when need instead of storing the created directory in an instance variable?

src/Zope2/Startup/tests/test_schema.py
```diff
@@ -43,26 +42,23 @@ class WSGIStartupTestCase(unittest.TestCase):
...
 
     def tearDown(self):
+        shutil.rmtree(self.TEMPNAME)
```
ale-rt, line 50: Using a context manager will save you from taking care of removing the created temp folder

dataflake: Like I said in my other comment, I am not storing the folder itself. The documentation states that the user is responsible for deletion, anyway.

src/Products/PageTemplates/tests/test_ptfile.py
```diff
@@ -15,15 +15,17 @@
...
     def setUp(self):
         from Products.PageTemplates.interfaces import \
             IUnicodeEncodingConflictResolver
         from Products.PageTemplates.unicodeconflictresolver import \
             DefaultUnicodeEncodingConflictResolver
         from zope.component import provideUtility
 
+        self.tmpfile = tempfile.NamedTemporaryFile(suffix=".zpt", delete=False)
+        self.TEMPFILENAME = self.tmpfile.name
+        self.tmpfile.close()
```
gforcada, line 27: We can use [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) as a context manager, so the last line is done automatically

dataflake: I changed those places to use a context manager

icemac:
@dataflake Where did you find these suggestions? I've tried several times to get something out of the CodeQL scans but did not find anything useful.

dataflake:
Click on the "Security" tab at the top of the screen and then you'll see "Code scanning" as one of the options on the left hand side.

icemac:
@dataflake Thank you for the hint.

PR #1115, opened 2023-03-30
Make zope.ini.in safe for operating systems that use backslashes

### LINKED ISSUE #1106: mkwsgiinstance utility invalid log paths in Windows
### What I did:

Used the mkwsgiinstance command line script to make a wsgi instance.

### What I expect to happen:

Create a zope.ini etc

### What actually happened:

zope.ini contained invalid paths for [handler_accesslog] args and [handler_eventlog] args. Example:
`args = ('C:\Users\wohnlice\PycharmProjects\Plone6/var/log/event.log', 'a')` The backslashes need to be escaped on Windows otherwise it reads \U as a unicode value and throws an error.

Windows 10, python 3.10, Zope2=4.0

ewohnlich:
Fixes #1106


dataflake (changes requested):
Please add a link to the issue that is fixed.

dataflake (approved):

PR #1116, opened 2023-03-31
Add specific gitignore rule for generated .mo files in the docs folder

dataflake:
This change removes the blanket ignoring of `.mo` files because some test artifacts exist under `Products.Five`. Instead, only those in the locale tree underneath the docs folder are ignored.


icemac (approved):
LGTM.

PR #1117, opened 2023-03-31
Do not break with GET requests that have a query string and a Content-Type header 

ale-rt:
We cannot make a GET request that contains both a Content-Type header and a query string.

Examples that work fine:
```shell
$ curl https://classic.demo.plone.org/@@ok?bar=1
OK
$ curl -H 'Content-Type: text/plain' https://classic.demo.plone.org/@@ok
OK
```

Example that should work fine, but it is not working fine:
```shell
$ curl -H 'Content-Type: text/plain' https://classic.demo.plone.org/@@ok?a=1
{"error_type": "NotImplementedError"}
```

Example that should actually trigger the error:
```shell
$ curl -H 'Content-Type: text/plain' -d @<(echo b=1) https://classic.demo.plone.org/@@ok?a=1
{"error_type": "NotImplementedError"}
```

Fixes plone/plone.restapi#1611

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1441,7 +1441,7 @@ class ZopeFieldStorage(ValueAccessor):
                 raise NotImplementedError("request parameters and body")
             if fpos is not None:
                 fp.seek(fpos)
-        elif fp and url_qs:
+        elif url_qs and fp is not None:
             raise NotImplementedError("request parameters and body")
         fl = []
         add_field = fl.append
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1441,7 +1441,7 @@ class ZopeFieldStorage(ValueAccessor):
                 raise NotImplementedError("request parameters and body")
             if fpos is not None:
                 fp.seek(fpos)
-        elif url_qs and content_type != "application/x-www-form-urlencoded":
+        elif fp and url_qs:
             raise NotImplementedError("request parameters and body")
         fl = []
         add_field = fl.append
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,9 +10,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.2 (unreleased)
 ------------------
 
-- Support Plone REST API calls with a query string.
-  For details see `plone/plone.restapi#1611
-  <https://github.com/plone/plone.restapi/issues/1611>`_.
+- Do not break on GET requests that pass a query string
+  and a `Content-Type` header.
+  For details see `#1117 <https://github.com/zopefoundation/Zope/pull/1117>`_.
 
 - Implement code change suggestions from CodeQL scanning.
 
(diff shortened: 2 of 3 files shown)


icemac:
@ale-rt Thank you for your PR. Do you think you can come up with a test as the issues seems to be an untested edge case.

src/ZPublisher/HTTPRequest.py
```diff
@@ -1441,7 +1441,7 @@ def __init__(self, fp, environ):
                 raise NotImplementedError("request parameters and body")
             if fpos is not None:
                 fp.seek(fpos)
-        elif url_qs and content_type != "application/x-www-form-urlencoded":
+        elif fp and url_qs:
```
d-maurer, line 1444: I suggest to replace this by `elif url_qs and fp is not None`.
`fp` is supposed to be either `None` (no body)  or a file like object (containing the body). While `bool(fp)` and `fp is not None` is equivalent for this case, the latter is more explicit.

d-maurer (approved):
The PR fixes a bug (the existence of a request body should have been checked explicitly rather than via a check against the default content type).

On the other hand, it is unclear why `plone.restapi` sends a `Content-Type` header when it does not send a request body. https://datatracker.ietf.org/doc/html/rfc2616#section-7.2.1 defines `Content-Type` as the body type **when** there is a body.

The PR contains a quick fix for a doubtful use of the `Content-Type` request header. A better fix would be to remove `ZPublisher`'s limitation to be unable to handle both request parameters and a request body.

ale-rt:
> The PR fixes a bug (the existence of a request body should have been checked explicitly rather than via a check against the default content type).
> 
> On the other hand, it is unclear why `plone.restapi` sends a `Content-Type` header when it does not send a request body. https://datatracker.ietf.org/doc/html/rfc2616#section-7.2.1 defines `Content-Type` as the body type **when** there is a body.
> 
> The PR contains a quick fix for a doubtful use of the `Content-Type` request header. A better fix would be to remove `ZPublisher`'s limitation to be unable to handle both request parameters and a request body.

Thanks for your insights!
I think that the JS library we are using incorrectly adds that `Content-Type` header instead of the `Accept` one. So it is not really an issue with `plone.restapi` itself.
I will report when I have more infos and update the PR accordingly.

ale-rt:
Writing the tests I find out that actually, even if the body is submitted, for `GET` requests it is discarded.
Maybe it should be like that also for other requests method.
Anyway, I think I was able to write a significant test for a `PUT` request.

Would you mind @d-maurer to have another look if I grokked that right?

Thanks a lot!

d-maurer:
Alessandro Pisa wrote at 2023-5-10 03:58 -0700:
>Writing the tests I find out that actually, even if the body is submitted, for `GET` requests it is discarded.

`GET` (and `HEAD`) requests do not have a body -- by the HTTP specification.
Therefore, ignoring a body in that case is justifyable.

>Maybe it should be like that also for other requests method.

This would be against the HTTP specification.

>Anyway, I think I was able to write a significant test for a `PUT` request.
>
>Would you mind @d-maurer to have another look if I grokked that right?

I will look at it tomorrow.

d-maurer (approved):

PR #1124, opened 2023-05-12
Allow ZPublisher to handle a query string together with a request body

### LINKED ISSUE #1122: query-parameters on post-request no longer supported with new ZopeFieldStorage
## BUG/PROBLEM REPORT / FEATURE REQUEST

First of all: Thanks for your hard work you all do on this project. Zope feels really, really professional maintained at the moment! Every thing is stable and works like a charm. I only got one small issue caused by refactoring the ZopeFieldStorage part.

### What I did:

Our zope-instance is behind a Apache-Webserver which does a sub-request to zope over mod_proxy.

We use RewriteRules like this:
`Rewrite (de|en)/view.json http://localhost:8080/some/dir/view_json?project_id=123&language_key=$1 [L,P]`

If I do a `POST`-request to the `/de/view.json` with `Content-type: application/json`, query-params are added by Apache, Zope raises a `NotImplementedError("request parameters and body")` in the latest version.

This worked for many years. Two days ago I updated our staging-system to Zope-5.8.1 and it broke.

I think it`s not an error to have both, query-params and a request-body, in one http-request according to http-standard, is it?

### What I expect to happen:

Add the query-parameters and the request-body to the HTTPRequest-instance like Zope did before.
So no Exception.

### What actually happened:

Zope raises a `NotImplementedError("request parameters and body")`

Zope-5.8.1
Python 3.11

### What helped

As always I was a little bit stressed, so I needed a quick fix. So I patched two lines:

removed:
https://github.com/zopefoundation/Zope/blob/6977b8880d0c18d9befb1699c437bc1346bafc1c/src/ZPublisher/HTTPRequest.py#L1441

added `self._file = fs.file` before:
https://github.com/zopefoundation/Zope/blob/6977b8880d0c18d9befb1699c437bc1346bafc1c/src/ZPublisher/HTTPRequest.py#L523

I also tried to add a test-case in the style of `test_processInputs_xxx` but this seems to be a little bit complicated for this use-case (`_makePostEnviron`). So I decided to spend my time in writing a good issue, sorry.

Whould be gread, if you find a better solution than me and bring it upstream! 

Thanks again

Michael

d-maurer:
Formerly, ZPublisher had the restriction to handle either request parameters (from a query string or a body with form data (i.e. a body with content type `application/x-www-form-urlencoded` or `multipart/form-data) or a request body not containing form data. A query string together with a body not containing form data raised a `NotImplementedError`.

This PR removes the restriction: request parameters from a query string can be combined with a request body not containing form data. Request parameters are made available in the request's `form` attribute (a `dict`), the request body can be accessed via the request key `BODY` (a `bytes` object) or `BODYFILE` (a file like object).

This PR fixes #1122.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,13 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.2 (unreleased)
 ------------------
 
+- Allow ``ZPublisher`` to handle both a query string and a request body;
+  the request parameters from the query string are made available
+  in the request attribute ``form`` (a ``dict``),
+  the request body can be accessed via the request keys ``BODY``
+  (a ``bytes`` object) or ``BODYFILE`` (a file like object).
+  Fixes `#1122 <https://github.com/zopefoundation/Zope/issues/1122>`_.
+
 - Do not break on GET requests that pass a query string
   and a `Content-Type` header.
   For details see `#1117 <https://github.com/zopefoundation/Zope/pull/1117>`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
   (a ``bytes`` object) or ``BODYFILE`` (a file like object).
   Fixes `#1122 <https://github.com/zopefoundation/Zope/issues/1122>`_.
 
+- Support access to the request's ``BODY`` key for WSGI servers
+  which hand over an unseekable request body (such as e.g.
+  ``Gunicorn``).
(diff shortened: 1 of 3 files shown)


ale-rt (approved):
LGTM! Thanks :)

dataflake (approved):

mtoepfl:
Works for my use case in #1122. Thank you @d-maurer for fixing!

d-maurer:
The PR now also supports access to the request's `BODY` key for WSGI servers which hand over an unseekable request body (such as e.g. `Gunicorn`). Formerly, an `AttributeError: tell` has been raised in this case.
This fixes #1125.

laulaz (approved):

PR #1133, opened 2023-06-15
fix handling of a request parameter of type `file` if no file has been selected

### LINKED ISSUE #1132: Changed Request handling of file uploads
The behavior of file uploads has changed, most likely with the changes flowing into Zope 5.8.1, such as #1098 and maybe #1094. The issue is illustrated in #1131.

In the past, the variable name for a file upload widget would only appear on the Request if someone had selected a file and the widget contained a path. If not, the widget name didn't show up at all. So it was safe to assume "if the name shows up on the Request, I can be sure it points to a file instance". Now all of a sudden the name always shows up, and if the field was left empty it contains an empty string.

#1131 is only one place that required code changes. I am seeing this issue pop up in customer code as well. I consider this a regression and a bug.

@d-maurer Can we get the old behavior back?

d-maurer:
Fixes #1132.

`ZPublisher` used to map request parameters of type `file` to `FileUpload` objects; the application could check for a non-empty `filename` to determine whether a file has been selected for this parameter.
The switch from `cgi.fieldstorage` to `multipart` changed this behavior: a `file` request parameter was no longer mapped to a `FileUpload` object when no file has been selected.
This PR restores the former behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.3 (unreleased)
 ------------------
 
+- Fix handling of a request parameter of type ``file`` if no value
+  has been specified;
+  fixes `#1130 <https://github.com/zopefoundation/Zope/issues/1130>`_.
+
 - Fix adding Page Templates without valid file input from the ZMI
   (`#1130 <https://github.com/zopefoundation/Zope/issues/1130>`_)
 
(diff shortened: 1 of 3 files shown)


d-maurer:
The Python 3.12 check failure can be ignored: a single test fails: `testPerformance` from the `ZCatalog` test suite - this test is known to occasionally fail (due to a not very robust design).

dataflake (approved):
Thank you for handling this so quickly. I added one more test simulating adding a page template in the ZMI with no input. If you merge I'll create a new release.

PR #1135, opened 2023-06-17
Make `mapply`  `__signature__` aware

### LINKED ISSUE #1134: Feature request: support Python 3 style signature declarations in `ZPublisher.mapply.mapply`
Python 3 has introduced the `__signature__` attribute to inform a caller about the signature of a callable. This significantly simplifies the implementation of signature preserving decorators.

Unfortunately, `ZPublisher.mapply.mapply` does not yet support `__signature__`.
As a consequence, methods decorated via `decorator 5+` decorators cannot be published when they have parameters
(earlier `decorator` versions implement decorators in a much more complicated Python 2 and `mapply` compatible way).

Extend `Zpublisher.mapply.mapply` to support `__signature__`.

d-maurer:
Fixes #1134.

This PR makes `mapply` `__signature__` aware.

`__signature__` has been introduced by Python 3 to allow (usually decorators) to specify the signature of a function. It makes it easy to implement signature preserving decorators.

With this PR `mapply` looks for `__signature__`  and (if present) uses it to determine the parameter names and their default. Otherwise, `mapply` determines this information as before.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Make ``mapply`` ``__signature__`` aware.
   This allows to publish methods decorated via a decorator
-  which sets ``__signature__` on the wrapper to specify
+  which sets ``__signature__`` on the wrapper to specify
   the signature to use.
   For details, see
   `#1134 <https://github.com/zopefoundation/Zope/issues/1134>`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,15 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Make ``mapply`` ``__signature__`` aware.
+  This allows to publish methods decorated via a decorator
+  which sets ``__signature__` on the wrapper to specify
+  the signature to use.
+  For details, see
+  `#1134 <https://github.com/zopefoundation/Zope/issues/1134>`_.
+  Note: ``mapply`` still does not support keyword only, var positional
+  and var keyword parameters.
+
 
 5.8.3 (2023-06-15)
 ------------------
--- a/src/ZPublisher/mapply.py
+++ b/src/ZPublisher/mapply.py
@@ -58,7 +58,7 @@ def mapply(object, positional=(), keyword={},
             # We use ``getfullargspec`` because it packages
             # the signature information in the way we need it here.
             # Should the function get deprecated, we could do the
-            # packaging outselves
+            # packaging ourselves
             argspec = getfullargspec(f)
             defaults = argspec.defaults
             names = argspec.args[count:]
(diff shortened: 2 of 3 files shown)


src/ZPublisher/mapply.py
```diff
@@ -50,9 +52,20 @@ def mapply(object, positional=(), keyword={},
...
-        code = f.__code__
-        defaults = f.__defaults__
-        names = code.co_varnames[count:code.co_argcount]
+        if hasattr(f, "__signature__"):
+            # The function has specified the signature to use
+            # (likely via a decorator)
+            # We use ``getfullargspec`` because it packages
+            # the signature information in the way we need it here.
+            # Should the function get deprecated, we could do the
+            # packaging outselves
```
gforcada, line 61:
```suggestion
            # packaging ourselves
```

src/ZPublisher/tests/test_mapply.py
```diff
@@ -90,3 +90,17 @@ class NoCallButAcquisition(Acquisition.Implicit):
...
         self.assertRaises(TypeError, mapply, ob, (), {})
+
+    def testFunctionWithSignature(self):
+        from inspect import Parameter
+        from inspect import Signature
+
+        def f(*args, **kw):
+            return args, kw
+
+        f.__signature__ = Signature(
```
icemac, line 101: In the changelog you write about decorators setting `__signature__`. Would it be possible to use such a decorator here to make the test consistent with the change log?

d-maurer: Do you think this is necessary?

The change log entry addresses users. Like one of my clients, they likely observe that some methods decorated with help of the `decorator` package can no longer be published correctly. Cause: up to `decorator 4.x`, the package used a complex, Python 2 and `mapply` compatible way to implement signature preserving decorators; `decorator 5.x` dropped Python 2 support and now simply informs about the signature of the decoration wrapper by setting its `__signature__`.

The test checks that `__sigature__` is honored. It is not important how or for what reason it has been set.

If you think, the change log entry and the test must be consistent, I will remove the decorator reference in the change log entry.

icemac: I do not think adapting the test is absolutely necessary. My idea was to make the test more end-to-end by letting the decorator generate the `__signature__` instead of doing it manually. It seems that to achieve this goal a dependency on the `decorator` package is needed.

So let's keep it like it is, as a new dependency just for a single test case seems too heavy.

CHANGES.rst
```diff
@@ -12,6 +12,15 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Make ``mapply`` ``__signature__`` aware.
+  This allows to publish methods decorated via a decorator
+  which sets ``__signature__` on the wrapper to specify
```
icemac, line 17:
```suggestion
  which sets ``__signature__`` on the wrapper to specify
```

icemac (approved):
LGTM.

PR #1137, opened 2023-06-19
Disable `ZCatalog`  performance test

### LINKED ISSUE #1136: Question: disable `ZCatalog` performance test? How?
The ZCatalog performance test (`Products.ZCatalog.PluginIndexes.CompositeIndex.tests.testCompositeIndex.CompositeIndexPerformanceTest.testPerformance`) occasionally fails for `Zope` test runs on Github (as it highly dependents on load conditions). Such failures are inconvenient as they require an analysis what went wrong and whether the problem is serious (it is not for this failure). It would be preferable not to run this test as part of the `Zope` tests.

My analysis in "https://github.com/zopefoundation/Products.ZCatalog/issues/147" revealed that the test already provides a way to get disabled: run unit tests only. Unfortunately, this cannot be used (directly) as part of the `Zope` tests because they contain important non-unit tests.

Questions:
 * should we disable the test above?
 * what would be a good way to do this?

d-maurer:
Fixes #1136.

This PR disables a `ZCatalog` (more precisely `Products.PluginIndexes`) performance test (`Products.ZCatalog.PluginIndexes.CompositeIndex.tests.testCompositeIndex.CompositeIndexPerformanceTest.testPerformance`) which depends highly on load conditions and therefore occasionally fails on GITHUB.

The test has been designed to get disabled when running unit tests only. It achieves this using an associated layer. The `ZCatalog` (and in particular the `Products.PluginIndexes`) tests do not use layers apart from this test. Therefore, this PR excludes the test by excluding tests with layers from `Products.PluginIndexes`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.4 (unreleased)
 ------------------
 
+- Disable a ``ZCatalog`` (more precisly: ``Products.PluginIndexes``)
+  performance test which occasionally fails on GITHUB.
+  For details, see
+  `#1136 <https://github.com/zopefoundation/Zope/issues/1136>`_.
+
 - Update to newest compatible versions of dependencies.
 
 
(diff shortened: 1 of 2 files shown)


icemac (approved):
The ZCatalog performance tests still run on `coverage` but the failure should happen less often with the changes in this PR.

PR #1138, opened 2023-06-19
Restore filename on code objects: `App.Extensions.getObject()`.

icemac:
Since 4.0a6 the filename is missing there, resulting in problems while debugging (debugger is unable to show the code line) and coverage of Python modules loaded via Products.ExternalMethods.

The commit introducing this change was https://github.com/zopefoundation/Zope/commit/8d30ff802e60eae55711f200c3197916fa5f90e1

The previous code cannot be restored because in Python 3 `exec` no longer takes an open file as argument but just `str`, `bytes`, or `Code` objects.

### DIFF
--- a/src/App/tests/test_Extensions.py
+++ b/src/App/tests/test_Extensions.py
@@ -1,0 +1,37 @@
+import types
+import unittest
+from pathlib import Path
+
+import App.config
+from App.Extensions import getObject
+
+
+class GetObjectTests(unittest.TestCase):
+    """Testing ..Extensions.getObject()."""
+
+    def setUp(self):
+        cfg = App.config.getConfiguration()
+        assert not hasattr(cfg, 'extensions')
+        cfg.extensions = Path(__file__).parent / 'fixtures'
+
+    def tearDown(self):
+        cfg = App.config.getConfiguration()
+        del cfg.extensions
+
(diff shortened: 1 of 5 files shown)


src/App/tests/test_Extensions.py
```diff
@@ -0,0 +1,38 @@
...
+        cfg = App.config.getConfiguration()
+        assert not hasattr(cfg, 'extensions')
+        cfg.extensions = Path(__file__).parent / 'fixtures'
+
+    def tearDown(self):
+        cfg = App.config.getConfiguration()
+        del cfg.extensions
+
+    def test_Extensions__getObject__1(self):
+        """It returns an the requested function.
```
d-maurer, line 22: Something is wrong here.
I assume, the documentation should be something like:
`Check that "getObject" returns the requested function and that its code object has the path set.`

d-maurer (approved):
Okay -- apart from the text change.

src/App/tests/test_Extensions.py
icemac: Fixed in https://github.com/zopefoundation/Zope/pull/1138/commits/e0aff9562db2bfb623c81740ebbb538dadab0439.

icemac:

PR #1142, opened 2023-07-13
make Zope's parameters for denial of service protection configurable

### LINKED ISSUE #1141: ZPublisher.HTTPRequest.FORM_MEMORY_LIMIT too low
Hello @d-maurer ,
the choosen data limit for FORM_MEMORY_LIMIT now is limited to only 1 Mb;
https://github.com/zopefoundation/Zope/blob/345d656557bfc414f70b92637fef33dde747a97d/src/ZPublisher/HTTPRequest.py#L54C9-L58

This FORM_MEMORY_LIMIT value was introduced here:
https://github.com/zopefoundation/Zope/pull/1094
https://github.com/zopefoundation/Zope/pull/1094/commits/25990c13f15727d14e118f4170c8e266ef24bb1a

Actually 1Mb is quite low if you work e,g, with [pdfmake,js](https://github.com/bpampuch/pdfmake) to generate PDF in the browser:

_FORM_MEMORY_LIMIT =  1Mb blocks PDF generation in the frontend_
DESY_V22_PDF_Formlimit0

_FORM_MEMORY_LIMIT  "unlimited"_
DESY_V22_PDF_Formlimit1

If the value FORM_MEMORY_LIMIT  was chosen arbitrarily, I 'd like to suggest setting it a little bit higher, e.g. to 2\*\*22 (=4Mb).

Best regards
f

d-maurer:
Fixes #1141.

### DIFF
--- a/src/Zope2/Startup/tests/test_schema.py
+++ b/src/Zope2/Startup/tests/test_schema.py
@@ -190,3 +190,46 @@ class WSGIStartupTestCase(unittest.TestCase):
             self.assertFalse(webdav.enable_ms_public_header)
         finally:
             webdav.enable_ms_public_header = default_setting
+
+    def test_dos_protection(self):
+        from ZPublisher import HTTPRequest
+
+        params = ["FORM_%s_LIMIT" % name
+                  for name in ("MEMORY", "DISK", "MEMFILE")]
+        defaults = dict((name, getattr(HTTPRequest, name)) for name in params)
+
+        try:
+            # missing section
+            conf, handler = self.load_config_text("""\
+                instancehome <<INSTANCE_HOME>>
+                """)
+            handleWSGIConfig(None, handler)
+            for name in params:
+                self.assertEqual(getattr(HTTPRequest, name), defaults[name])
+
+            # empty section
+            conf, handler = self.load_config_text("""\
+                instancehome <<INSTANCE_HOME>>
+                <dos_protection />
+                """)
+            handleWSGIConfig(None, handler)
+            for name in params:
+                self.assertEqual(getattr(HTTPRequest, name), defaults[name])
+
+            # configured values
+
+            # empty section
+            conf, handler = self.load_config_text("""\
+                instancehome <<INSTANCE_HOME>>
+                <dos_protection>
+                  form-memory-limit 1KB
+                  form-disk-limit 1KB
+                  form-memfile-limit 1KB
+                </dos_protection>
+                """)
+            handleWSGIConfig(None, handler)
+            for name in params:
+                self.assertEqual(getattr(HTTPRequest, name), 1024)
+        finally:
+            for name in params:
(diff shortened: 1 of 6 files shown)


dataflake (approved):

src/Zope2/Startup/tests/test_schema.py
```diff
@@ -190,3 +190,46 @@ def test_ms_public_header(self):
             self.assertFalse(webdav.enable_ms_public_header)
         finally:
             webdav.enable_ms_public_header = default_setting
+
+    def test_dos_protection(self):
+        from ZPublisher import HTTPRequest
+
+        params = ["FORM_%s_LIMIT" % name
+                  for name in ("MEMORY", "DISK", "MEMFILE")]
+        defaults = dict((name, getattr(HTTPRequest, name)) for name in params)
```
ale-rt, line 199: Why not using the more modern approach?
```suggestion
        defaults = {name: getattr(HTTPRequest, name) for name in params}
```
also it might be interesting to provide a default to `getattr`.

ale-rt: This is a test, nevermind about the default thing

d-maurer: I was not sure whether dict comprehension was already available in 3.7 (and too lazy to check it).

drfho: Thank you, I sucessfully completed some user tests on Py3.8 playing with some threshold values. So I can control whether my use-case (PDF-generation with JS) works or fails ;-)

In general I have some doubts whether this config is a good idea because even the default may generate troubles. And these errors are difficult to trace. Moreover (D)DOS protection can be considered as a webserver/firewall task and not the task of the app server. Maybe someone has a hint why this complication is needed at all?

drfho (approved):

d-maurer:
drfho wrote at 2023-7-13 07:15 -0700:
> ...
>In general I have some doubts whether this config is a good idea because even the default may generate troubles. And these errors are difficult to trace. Moreover (D)DOS protection can be considered as a webserver/firewall task and not the task of the app server. Maybe someone has a hint why this complication is needed at all?

The parameters (including their default values) come from
`multipart`, the third party product used for form processing
as replacement for Python's `cgi.FieldStorage`.
As `multipart` already contains the logic, there is not much
complexity added to `ZPublisher`.

Apparently, WSGI does not contain logic to parse/preprocess the
request for request variables but delegates this to the APP server
(or maybe WSGI middleware).

src/Zope2/Startup/tests/test_schema.py
icemac: > I was not sure whether dict comprehension was already available in 3.7 (and too lazy to check it).

According to https://peps.python.org/pep-0274/ it is already in 2.7 and 3.0, so feel free to use it everywhere.

PR #1143, opened 2023-07-14
fix (and test) limit for  `request.BODY`

d-maurer:
This is a fix for yesterdays #1142 (therefore no separate change log entry).

The documentation for #1142 promised that `FORM_MEMORY_LIMIT` would limit the size of an access to `request.BODY`. Unfortunately, I forgot to make a test for this promise -- and indeed, the promise has been broken.

This PR adds the missing test and fixes the implementation.


d-maurer:
The initial failure of `tests/maxos-py37` was caused by another flaky `ZCatalog` test ("...scoring"). Rerunning the tests has been successful.

dataflake (approved):

PR #1147, opened 2023-08-30
4.x: Update `RestrictedPython` to 5.4 to mitigate a security issue.

icemac:


d-maurer (approved):

dataflake (approved):

icemac:

icemac:
@dataflake @d-maurer Do you think we should cut a new 4.x release after the merge of this PR?

dataflake:
I have no idea what the 4.x branch is used for. Even Zope 4, which is as dead, uses the 5.x release. I don't see a reason for such a release.

d-maurer:
Michael Howitz wrote at 2023-8-30 23:31 -0700:
>@dataflake @d-maurer Do you think we should cut a new 4.x release after the merge of this PR?

`Zope` does not use the `safer_getattr` from `RestrictedPython`
but `guarded_getattr` from `AccessControl.ZopeGuards` which
enforces considerably more restrictions.
I do not know whether it prevents access to the `format` functionality
(but I seem to remember that we did something about it -- maybe not
for all attack vectors).

I will look into this in the next days.

d-maurer:
Jens Vagelpohl wrote at 2023-8-30 23:39 -0700:
>I have no idea what the 4.x branch is used for.

Plone 5.2 (the current stable Plone release) still uses Zope 4.x versions.

>Even Zope 4, which is as dead, uses the 5.x release. I don't see a reason for such a release.

While not impossible, it would be strange if Zope 4.x versions
would use 5.x (rather than 4.x) branches.

dataflake:
> While not impossible, it would be strange if Zope 4.x versions would use 5.x (rather than 4.x) branches.

They do, see https://github.com/zopefoundation/Zope/blob/4.x/versions-prod.cfg

d-maurer:
Jens Vagelpohl wrote at 2023-8-31 00:14 -0700:
>> While not impossible, it would be strange if Zope 4.x versions would use 5.x (rather than 4.x) branches.
>
>They do, see https://github.com/zopefoundation/Zope/blob/4.x/versions-prod.cfg

I think @icemac (and me) speaks about the Zope 4.x branch,
not the `RestrictedPython` 5.x versions.

@icemac has updated the Zope 4.x branch to use the newer
`RestrictedPython==5.4` version and asks whether this change
should be followed by a new Zope 4 release.

dataflake:
You're right, my bad, I thought this was about RestrictedPython releases. I didn't look closely enough.

About a Zope 4 release: I personally have no interest in Zope 4 anymore. Anyone feel free to either make a release or not.

PR #1151, opened 2023-09-09
Tighten down the ZMI frame source logic to only allow site-local sources

dataflake:
The ZMI frame logic uses a request variable `came_from` to determine the source location for the right-hand side frame when JavaScript is disabled. This can be passed in via query string, which means it's out of our control and must be sanity-checked. Only site-local pages should load into that frame.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,10 @@ The change log for the previous version, Zope 4, is at
 https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 
-5.9 (unreleased)
-----------------
+5.8.5 (unreleased)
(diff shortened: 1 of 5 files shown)


drfho (approved):
LGTM

icemac (approved):

d-maurer (approved):

mauritsvanrees (commented):
I am seeing some possible problems still, although they could be theoretical: I have a test case that accepts a came_from that should not be accepted, but in a browser nothing goes wrong yet.
I will try to come up with a PR today.

dataflake:
Please do. The way the code is written now it should be easy to add additional cases.

PR #1152, opened 2023-09-11
Harden getZMIMainFrameTarget

mauritsvanrees:
Do not accept a 'tainted' came_from parameter.
Do not accept a came_from path-only url starting with `//` either. Browsers can interpret the rest of the path as a domain.

A question here is: can we rely on possibly dangerous came_from parameters to always automatically be marked as tainted?
I tried adding this test:

```
request['came_from'] = '<script>alert("hi");</script>'
self.assertEqual(app.getZMIMainFrameTarget(request),
                    f'{URL1}/manage_workspace')
```

This fails, both on master and with the current PR:

```
Failure in test test_getZMIMainFrameTarget (OFS.tests.testApplication.ApplicationTests.test_getZMIMainFrameTarget)
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
  File "/Users/maurits/community/plone-coredev/6.0/src/Zope/src/OFS/tests/testApplication.py", line 173, in test_getZMIMainFrameTarget
    self.assertEqual(app.getZMIMainFrameTarget(request),
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 873, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 1253, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/local/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 703, in fail
    raise self.failureException(msg)
AssertionError: '<script>alert("hi");</script>' != 'http://nohost/manage_workspace'
- <script>alert("hi");</script>
+ http://nohost/manage_workspace
```

That sounds like we would still need to fix something.  But this is only theoretical: when translated to an actual url for a browser, you would get this:

    http://localhost:8080/manage?came_from=<script>alert("hi");</script>

Or what the browser actually sends:

    http://localhost:8080/manage?came_from=%3Cscript%3Ealert(%22hi%22);%3C/script%3E

On the master branch this gives a traceback:

```
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 390, in publish_module
  Module ZPublisher.WSGIPublisher, line 285, in publish
  Module ZPublisher.mapply, line 98, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module Shared.DC.Scripts.Bindings, line 333, in __call__
  Module Shared.DC.Scripts.Bindings, line 370, in _bindAndExec
  Module plone.app.theming.zmi, line 17, in _exec
  Module App.special_dtml, line 222, in _exec
  Module DocumentTemplate._DocumentTemplate, line 144, in render_blocks
  Module DocumentTemplate._DocumentTemplate, line 168, in render_blocks_
  Module DocumentTemplate.DT_Util, line 215, in eval
   - __traceback_info__: REQUEST
  Module <string>, line 1, in <module>
  Module OFS.Application, line 126, in getZMIMainFrameTarget
  Module urllib.parse, line 394, in urlparse
  Module urllib.parse, line 133, in _coerce_args
  Module urllib.parse, line 117, in _decode_args
  Module urllib.parse, line 117, in <genexpr>
  Module AccessControl.tainted, line 111, in __getattr__
AttributeError: 'str' object has no attribute 'decode'
```

This PR fixes the traceback, and disallows such tainted parameters because we do not trust this.

So I repeat the question: can we rely on potentially dangerous parameter strings to always be marked as tainted?
If not, I think the safest option is to copy [`Products.isurlinportal`](https://github.com/plone/Products.isurlinportal/blob/master/Products/isurlinportal/__init__.py) in here and remove the Plone-specific parts.  Or maybe I could extract the non-Plone parts (which should be most of it) into a separate package that both Zope and Plone could use.
See its [tests, starting from here](https://github.com/plone/Products.isurlinportal/blob/master/Products/isurlinportal/tests.py#L109) for some creative urls that may give problems.  Some may not be easily tested in a browser, as I saw above, but would need either `curl` or maybe they can be sent via javascript.

But: I could not trigger a problem in Firefox, except with the double double slashes, which is something else I fix in this PR.

### DIFF
--- a/src/OFS/Application.py
+++ b/src/OFS/Application.py
@@ -24,6 +24,7 @@ from AccessControl import ClassSecurityInfo
 from AccessControl.class_init import InitializeClass
 from AccessControl.Permission import ApplicationDefaultPermissions
 from AccessControl.Permissions import view_management_screens
+from AccessControl.tainted import TaintedString
 from Acquisition import aq_base
 from App import FactoryDispatcher
 from App.ApplicationManager import ApplicationManager
@@ -122,14 +123,29 @@ class Application(ApplicationDefaultPermissions, Folder.Folder, FindSupport):
         if not came_from:
             return default
 
+        # When came_from contains suspicious code, it will not be a string,
+        # but an instance of AccessControl.tainted.TaintedString.
+        # Passing this to urlparse, gives:
+        # AttributeError: 'str' object has no attribute 'decode'
+        # This is good, but let's check explicitly.
+        if isinstance(came_from, TaintedString):
+            return default
+        try:
+            parsed_came_from = urlparse(came_from)
+        except AttributeError:
+            return default
         parsed_parent_url = urlparse(parent_url)
-        parsed_came_from = urlparse(came_from)
 
         # Only allow a passed-in ``came_from`` URL if it is local (just a path)
         # or if the URL scheme and hostname are the same as our own
-        if (not parsed_came_from.scheme and not parsed_came_from.netloc) or \
-           (parsed_parent_url.scheme == parsed_came_from.scheme
+        if (parsed_parent_url.scheme == parsed_came_from.scheme
                 and parsed_parent_url.netloc == parsed_came_from.netloc):
+            return came_from
+        if (not parsed_came_from.scheme and not parsed_came_from.netloc):
+            # This is only a path.  But some paths can be misinterpreted
+            # by browsers.
+            if parsed_came_from.path.startswith("//"):
+                return default
             return came_from
 
         return default
--- a/src/OFS/tests/testApplication.py
+++ b/src/OFS/tests/testApplication.py
@@ -1,5 +1,6 @@
 import unittest
 
+from AccessControl.tainted import TaintedString
 from Testing.ZopeTestCase import FunctionalTestCase
 
 
@@ -145,6 +146,12 @@ class ApplicationTests(unittest.TestCase):
             self.assertEqual(app.getZMIMainFrameTarget(request),
                              '/new/path')
 
+            # Tainted local path.  came_from can be marked as 'tainted' if it
+            # suspicious contents.  It is not accepted then.
+            request['came_from'] = TaintedString('/new/path')
+            self.assertEqual(app.getZMIMainFrameTarget(request),
+                             f'{URL1}/manage_workspace')
+
             # came_from URL outside our own server
             request['came_from'] = 'https://www.zope.dev/index.html'
             self.assertEqual(app.getZMIMainFrameTarget(request),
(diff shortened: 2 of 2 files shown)


dataflake:
> A question here is: can we rely on possibly dangerous came_from parameters to always automatically be marked as tainted?

I do not know. Dieter might know.

I don't know about copying a whole bunch of code here. The idea of sharing a simple package that does not have any Plone dependencies sounds like a better idea.

dataflake (approved):

d-maurer:
Maurits van Rees wrote at 2023-9-11 06:11 -0700:
> ...
>A question here is: can we rely on possibly dangerous came_from parameters to always automatically be marked as tainted?

An `str` (or `bytes`) is considered unsafe, if it contains `<` and
then will be marked as "tainted".
**BUT** `HTTPRequest.__getitem__` does **NOT** automatically
return a value tainted in this way. It must be called with `returnTaints=True`!

Therefore, you should access `came_from` via
`request.get("came_from", returnTaints=True)`
(not `request["came_from"]`).

d-maurer (changes requested):
`urllib.urlparse.urlparse` will parse '//...' not as a path only URL. Thus, this case was already handled.

To get a tainted value from the request object, the 'get' (or `getitem`) must be called with `returnTaints=True` (otherwise, you will always get the untainted value).

d-maurer:
Dieter Maurer wrote at 2023-9-11 15:51 +0200:
>Maurits van Rees wrote at 2023-9-11 06:11 -0700:
>> ...
>>A question here is: can we rely on possibly dangerous came_from parameters to always automatically be marked as tainted?
>
>An `str` (or `bytes`) is considered unsafe, if it contains `<` and
>then will be marked as "tainted".
>**BUT** `HTTPRequest.__getitem__` does **NOT** automatically
>return a value tainted in this way. It must be called with `returnTaints=True`!
>
>Therefore, you should access `came_from` via
>`request.get("came_from", returnTaints=True)`
>(not `request["came_from"]`).

The test should be adjusted accordingly:
`HTTPRequest.processInputs` puts the untained form values into the
`form` dict and the tainted form values into `taintedform`.
The test should therefore put an untainted `came_from` into `form`
and a tained one into `taintedform`.

mauritsvanrees:
> `urllib.urlparse.urlparse` would not parse `//...` as a path-only URL. Thus, this case is already taken care of. ```python
> > from urllib.parse import urlparse >> urlparse("//abc.com")
> ParseResult(scheme='', netloc='abc.com', path='', params='', query='', fragment='')
> >> urlparse("//ab")
> ParseResult(scheme='', netloc='ab', path='', params='', query='', fragment='') ```

True, but not this variant:

```
>>> from urllib.parse import urlparse
>>> urlparse("////abc.com")
ParseResult(scheme='', netloc='', path='//abc.com', params='', query='', fragment='')
```

With such a came_from, the abc.com website is actually loaded in the frame.
Yes, it is sneaky.

d-maurer (approved):
Convinced by your comments :-)

mauritsvanrees:
Thanks for the reviews!

PR #1156, opened 2023-09-15
WSGIPublisher: set REMOTE_USER even in case of error

### LINKED ISSUE #1155: Username not in access log for error requests
### What I did:

Create a zope instance:
```
mkwsgiinstance -d username_log_repro  -u  admin:admin
runwsgi  ./username_log_repro/etc/zope.ini
# add a `standard_error_message` at root ( this is necessary for error requests to be logged at all in access log - this looks like another independent issue )
curl -d id=standard_error_message 'http://admin:admin@127.0.0.1:8080/manage_addProduct/OFSP/addDTMLMethod'
```

Make a successful HTTP request and see in the log that username (`admin`) appears
```console
$ curl --silent -o /dev/null  http://admin:admin@127.0.0.1:8080 ; tail -n 1 username_log_repro/var/log/Z4.log
127.0.0.1 - admin [15/Sep/2023:16:13:03 +0200] "GET / HTTP/1.1" 200 2 "-" "curl/7.87.0"
```

Make an error HTTP request, the username field is empty:

```console
$ curl --silent -o /dev/null  http://admin:admin@127.0.0.1:8080/error ; tail -n 1 username_log_repro/var/log/Z4.log
127.0.0.1 - - [15/Sep/2023:16:14:30 +0200] "GET /error HTTP/1.1" 404 229 "-" "curl/7.87.0"
```

### What I expect to happen:

The username field should be present even for error requests.  

### What actually happened:

The username field is empty, for error requests. This is not only for "404 not found" requests, but also for server side errors.

This is on current master

---

I investigated and found that username is set in `environ` by https://github.com/zopefoundation/Zope/blob/59f68703c249a0a0f5b3b092395709820f221108/src/ZPublisher/WSGIPublisher.py#L388-L395 

but this code is not executed if `_publish` gets an exception. I'm making a pull request.

perrinjerome:
This fixes a problem that user name was empty in access log for error pages.

Fixes #1155

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
+- Added image dimensions to SVG file properties
+  `#1146 <https://github.com/zopefoundation/Zope/pull/1146>`_.
 
 5.8.4 (2023-09-06)
 ------------------
@@ -49,8 +51,6 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 - Update ``AccessControl`` to version 6.2 to mitigate a security problem.
   (CVE-2023-41050)
 
-- Added image dimensions to SVG file properties 
-  `#1146 <https://github.com/zopefoundation/Zope/pull/1146>`_.
 
 5.8.3 (2023-06-15)
 ------------------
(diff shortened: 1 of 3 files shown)


dataflake (dismissed):
Please add a change log entry

perrinjerome:
I'm sorry I forgot changelog again ! I just added one. These PR also has the changes from https://github.com/zopefoundation/Zope/pull/1158 to simplify git conflict

icemac (approved):

PR #1146, opened 2023-08-26
Added Image Info for SVG files

drfho:
SVG images are common today: now OFS.Image.getImageInfo can read the image dimensions from the xml file data in pixels (int-normalized).

svg_image_props

PR #1157, opened 2023-09-18
SVG-File: avoid attr error on missing dimensions

drfho:
In addition to https://github.com/zopefoundation/Zope/pull/1146:
SVG font files do not contain width/height dimension. This case of missing attrs is covered now and the preview show the browser rendered XML.

svg_fonts


icemac (approved):
LGTM.

Issue #1161, opened 2023-09-20
Update ACE editor


icemac:
The version we are using is more than four years old. I was even not able to find the actual version number we are using.

See https://github.com/zopefoundation/Zope/tree/master/src/zmi/styles/resources

drfho:
zmi version is 1.4.4, current built version is 1.5.0:
https://github.com/ajaxorg/ace-builds/releases/tag/v1.5.0
I will create a PR

icemac:
Fixed in #1166.

PR #1166, opened 2023-09-26
Update ACE editor 1.4.4 to 1.5.0

drfho:
Ref: https://github.com/zopefoundation/Zope/issues/1161

Updated with built version 1.5.0 
https://github.com/ajaxorg/ace-builds/releases/tag/v1.5.0

Screenshot: user tests
zmi_ace_usertest


drfho:
ad: _Code scanning / CodeQL / "Uncontrolled data used in path expression"_
I do not know whether this is just a theoretical problem, because `ace` is only used in the access-restricted GUI. Besides that, I am sorry to say, I have no idea how to prevent that check failure.

dataflake (commented):
Can't review this right now. On vacation till next week and the phone is all I have to look at the changes.

icemac:
@dataflake I think the review has time until you are back at your computer.

dataflake:
I have dismissed all those CodeQL alerts as not relevant. First of all, this is not our code and I doubt any of us as enough knowledge to judge the importance for Zope and make fixes. Secondly, the editor is only available for those having ZMI access, and not untrusted users.

There's no point in looking at individual JS files for me. I can't review those. If @drfho thinks this makes sense I trust him.

@drfho just to be on the safe side, can you verify that all those resources in the folder `src/zmi/styles/resources/ace.ajax.org` are truly only available to users with ZMI access? None of those URLs to these javascript files should be reachable by anonymous users or those who cannot view the ZMI.

dataflake:
I have added the following changes:

- guard access to `/++resource++zmi` with the "View management screens" permission, which is required for anyone using the ZMI
- create a new registration `/++resource++logo` that points to the icon and image files at `src/zmi/styles/resources/logo`, which are useful outside the ZMI and don't pose any harm
- use the icon and image files in new default content for the auto-generated root `index_html` so it's not completely naked like right now.

drfho (commented):
because there is not much structure/text,  I made it responsive to screen-size, reduced header lines, added a mimimal css for the resposive behaviour:

default_page

dataflake:
Super schick, Danke!

PR #1167, opened 2023-09-27
Base logic for CVE-2023-42458 on the media type proper

d-maurer:
ignore parameters and whitespace and normalize to lowercase.

Motivated by the report from Almir Zunic on "mailto:security@plone.org" " [p/z-sec] XSS - found in Management interface" I reexamined the logic developed for CVE-2023-42458 and found it incomplete. It compares the whole `content_type` against a list of types. But, at least if set manually, `content_type` can contain parameters and/or additional whitespace and may use uppercase letters which could not be directly compared with a types list.

When the logic is based on an allow list (the default), files with an unexpected `content_type` are delivered as attachment (which is safe). However, if the logic is based on a deny list, potentially dangerous files could be presented inline.

This PR bases the  CVE-2023-42458 logic on the media type proper: it ignores parameters and whitespace and normalizes to lowercase.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,9 +13,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
-- Base the inline/attachment logic developped for CVE-2023-42458
-  on the media type proper (ignoring parameters and leading/trailing
-  whitespace).
+- Base the inline/attachment logic developed for CVE-2023-42458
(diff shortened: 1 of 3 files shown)


icemac (approved):
LGTM

dataflake (approved):

mauritsvanrees (approved):
Yes, this works for me when I test it in Plone 6.0.  Thanks!

d-maurer:
Thank you for the reviews!

PR #1169, opened 2023-10-02
zmi: whitespace-cleanup, minor html fixes

drfho:
some dtml code cleaning of minor relevance


dataflake (approved):
I have decided to pull out the behavior change that sparked the issue and will handle that separately. Thank you for the cleanup work!

PR #1170, opened 2023-10-06
SVG-File: avoid attr error on missing dimensions (2)

drfho:
Determining SVG dimensions had still an logical problem on 'break'-exit: I indented the lines where resetting width/height happens too much:

https://github.com/zopefoundation/Zope/blob/4a221e929c7c44b2b133ce4726b6e3e22e1532bf/src/OFS/Image.py#L871-L872

Ref: https://github.com/zopefoundation/Zope/pull/1157

Now it's fixed:

[1] left: 'break'-exit resulted in "no dimension", right: fix
svg_size

[2] actually missing dimensions
svg_nosize

Sorry for that
f


dataflake (approved):

dataflake:
@drfho Is anything holding this up?

PR #1172, opened 2023-10-12
honor a request's `Content-Length`

### LINKED ISSUE #1171: Zope 5.8.1 breaks using `wsgiref`
### What I did:

I am using Python's stdlib `wsgiref` in selenium tests to start a simple WSGI server.

### What I expect to happen:

Using `wsgiref` should be possible.

### What actually happened:

`wsgiref` uses a `socket` as `stdin` for the WSGI environment.
And Zope 5.8.1+ uses `read()` to get the data from this socket, the execution is blocked and later on runs into a timeout.

Previous versions (via `cgi` module) called `readline()` on the socket which does not block.

(Sorry I cannot provide a minimal test example as it involves too much custom code. Additionally I did not find an easy way to start a Zope instance using `wsgiref` – the examples for using a different WSGI server in the docs seem to require some preparation.)

* Zope 5.8.6
* MacOS 13.6
* Python 3.11

d-maurer:
Fixes #1171.

This PR honors a request's `Content-Length` if the incoming input stream is not seekable (and therefore, the WSGI server might not have done this already).

It should ensure correct behavior even if the WSGI server does not honor the WSGI recommendation to emulate an EOF once `Content-Length` bytes have been read. It works by wrapping the input stream implementing the recommendation.

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1505,7 +1505,7 @@ class LimitedFileReader:
     """File wrapper emulating EOF."""
 
     # attributes to be delegated to the file
-    DELEGATE = set("close closed fileno mode name".split())
+    DELEGATE = set(["close", "closed", "fileno", "mode", "name"])
 
     def __init__(self, fp, limit):
         """emulate EOF after *limit* bytes have been read.
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1427,9 +1427,17 @@ class ZopeFieldStorage(ValueAccessor):
     VALUE_LIMIT = Global("FORM_MEMORY_LIMIT")
 
     def __init__(self, fp, environ):
-        self.file = fp
         method = environ.get("REQUEST_METHOD", "GET").upper()
         url_qs = environ.get("QUERY_STRING", "")
+        content_length = environ.get("CONTENT_LENGTH")
+        if content_length:
+            try:
+                fp.tell()
+            except Exception:
+                # potentially not preprocessed by the WSGI server
+                # enforce ``Content-Length`` specified body length limit
+                fp = LimitedFileReader(fp, int(content_length))
+        self.file = fp
         post_qs = ""
         hl = []
         content_type = environ.get("CONTENT_TYPE",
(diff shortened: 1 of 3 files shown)


d-maurer:
@icemac Would you check that this PR fixes #1171.

d-maurer:
The `lint` failure is in fact caused by a mismatch between `setup.py` and `python_requires` (not introduced by myself).
Apparently, `python_requires` says `3.12` is acceptable, but `setup.py` does not.
As far as I know, we cannot yet declare Python 3.12 supported because we do not yet know that `RestrictedPython` does.

src/ZPublisher/HTTPRequest.py
```diff
@@ -1493,6 +1501,53 @@ def __init__(self, fp, environ):
             add_field(field)
 
 
+class LimitedFileReader:
+    """File wrapper emulating EOF."""
+
+    # attributes to be delegated to the file
+    DELEGATE = set("close closed fileno mode name".split())
```
icemac, line 1508:
```suggestion
    DELEGATE = set(["close", "closed", "fileno", "mode", "name"])
```

Using `split` it is easier to write but harder to read, because it is necessary to read the whole line to understand that not a string is given to `set` but a list.

icemac (approved):
I can confirm that this change fixes #1171.

Thank you for tackling this issue.

dataflake (approved):

d-maurer:
Jens Vagelpohl wrote at 2023-10-12 03:38 -0700:
>The lint issue is known and will be solved once Zope declares Python 3.12 support in the trove classifiers. This is waiting for the review of RestrictedPython for Python 3.12.

When I understood this correctly, work has been done in the `meta` project
to separate those (mostly release related) version checks from the
lint check proper.

Could we get his work into `Zope`?
The current `lint` check failure will prevent merging this and other PRs
into `master`.
Reviewing `RestrictedPython` for Python 3.12 may take considerable time.
Until then, several PRs may have piled up.
Additional work (mostly related to merge conflict resolution for
`CHANGES.rst`) would result.

dataflake:
We are still waiting for reviews for several items in `meta`. Until then I have added you to the "Release Managers" team, which means you can circumvent the test requirements and merge.

PR #1173, opened 2023-10-13
Fix #1172

d-maurer:
#1172 has defined  `ZPublisher.HTTPRequest.LimtedFileReader.__del__` as
```python
    def __del__(self):
        return self.fp.__del__()
```
This is wrong: `fp` might be referenced from somewhere else; then its `__del__` should not yet be called.

In fact, we do not need a  `__del__`  definition.
Therefore, this PR removes it.

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -1539,9 +1539,6 @@ class LimitedFileReader:
             raise StopIteration()
         return data
 
-    def __del__(self):
-        return self.fp.__del__()
-
     def __getattr__(self, attr):
         if attr not in self.DELEGATE:
             raise AttributeError(attr)


dataflake (approved):

d-maurer:
@dataflake Thank you for bringing the linting/release check separation into `Zope`.

I wait with the merge to give @icemac the chance to check it: the wrong `__del__` definition might have had a beneficial side effect: closing the channel after the request has been processed even though it might still be in use; potentially, this has avoided the problem mentioned in the last comment of #1171.

icemac:
@d-maurer With this change `wsgiref` still works in my local test which broke before #1172.

icemac (approved):

d-maurer:
@dataflake @icemac Thank you for your reviews!

However, I meanwhile believe that the workaround in #1172 is not optimal. #1171 should be tackled by a WSGI middleware component between a not well behaving WSGI server and `Zope`, not within `Zope` itself.

Therefore, I propose that we revert #1172 and close this PR. Instead we look for an appropriate WSGI middleware component.

I have browsed WSGI related projects on PyPI. Unfortunately, there are thousands of them. Looking at a few hundreds, I found an appropriate base classs (`LimitedStream`) in `Werkzeug` (the WSGI library behind `Flask`), but not yet a ready to use WSGI middleware for our purposes. I will continue the search.

Should we find nothing, I propose the creation of a new `zopefoundation` project `zope.wsgi`, collecting WSGI middleware components which might be helpful together with `Zope` (and other WSGI applications). Its first member could be `LimitInput`, a middleware component enforcing the recommendation of note 1 of "https://peps.python.org/pep-3333/#input-and-error-streams". A further middleware candidate could provide transfer encoding support (as required by HTTP/1.1).

What do you think?

dataflake:
I don't care either way. Michael is the one with issues and he should decide.

d-maurer:
`Zope` uses `paste.deploy` to configure the WSGI environment.
`paste.deploy` supports WSGI middleware components via the  "filter" concept. A filter is a function taking a WSGI application and optionally global configuration options as parameters and returns a new WSGI application (a WSGI middleware component transforms a WSGI application into a new one).

In the default configuration, `Zope` uses 2 filters: `translogger` (defined by `Paste`) and `Zope#httpexcptions` (--> `ZPublisher.httpexceptions.HTTPExceptionHandler`). This means that `Zope` already defines a WSGI middleware component.
If we do not like the creation of a new project, the new WSGI middleware components could also be defined by `Zope` itself.

d-maurer:
This PR does not yet ensure HTTP/1.1 compatibility. Concretely, it violates "https://datatracker.ietf.org/doc/html/rfc7230#section-6.3":
```
...
A server MUST read
the entire request message body or close the connection after sending
its response, since otherwise the remaining data on a persistent
connection would be misinterpreted as the next request.
...
```
We do need a `__del__` and it must read (and discard) any remaining data.

icemac:
We could either go the WSGI middleware route or just add a section after https://zope.readthedocs.io/en/latest/operation.html#problematic-wsgi-servers declaring `wsgiref` as not supported because of violation of the WSGI standards.

In my tests I solved the urgent problem by switching from `wsgiref` to `waitress`, so it is not pressing me.
Probably no-one else is using `wsgiref` but I do not know if there are other WSGI servers which also misbehave.

d-maurer:
Michael Howitz wrote at 2023-10-16 23:00 -0700:
>We could either go the WSGI middleware route or just add a section after https://zope.readthedocs.io/en/latest/operation.html#problematic-wsgi-servers declaring `wsgiref` as not supported because of violation of the WSGI standards.

Thank you for your feedback.

I am working on the WSGI middleware route.

d-maurer:
Obsoleted by #1175.

PR #1175, opened 2023-10-17
Wsgi content length

### LINKED ISSUE #1171: Zope 5.8.1 breaks using `wsgiref`

d-maurer:
Allows administrators to fix #1171.

The PR defines the new `paste.filter_app_factory` entry point `content_length` corresponding to a WSGI middleware component honoring a `Content-Length` request header.

Administrators can integrate this middleware into the WSGI pipeline when the used WSGI server does not honor the `Content-Length` request header. The [WSGI spec](https://peps.python.org/pep-3333/) recommends in note 1 of section [Input and Error Streams](https://peps.python.org/pep-3333/#input-and-error-streams) that servers should honor the `Content-Length` request header. Therefore, most servers will do and the middleware should rarely be necessary; but there are exceptions, e.g. Python's `wsgiref.simple_server`.


src/ZPublisher/tests/test_paste.py
icemac, line 1: The module under test is called `pastefilter`, so I'd prefer the test module to be called `test_pastefilter`.

icemac (approved):
I like the approach especially the change in the zope.ini skel file documenting the usage.

src/ZPublisher/tests/test_paste.py
dataflake: I agree

dataflake (approved):
Just the one test module name change

d-maurer:
Michael Howitz wrote at 2023-10-17 23:07 -0700:
>
>I like the approach especially the change in the zope.ini skel file documenting the usage.
>
>On src/ZPublisher/tests/test_paste.py:
>
>The module under test is called `pastefilter`, so I'd prefer the test module to be called `test_pastefilter`.

I have used the more general `test_paste` because initially
I wanted to test more of the WSGI related configuration.
e.g. `paste.deploy.loadapp` (applied to `Zope2.utilities:skel/zope.ini.in`).
As far as I know, there are not yet tests for this.

After the recognition that the WSGI middleware already used
by `Zope` (--> `httpexceptions`) has no filter definition
in `zope.ini` but is included in the pipeline via a direct `egg:` reference,
I decided to do the same with the new middleware component.
Its test therefore did not require a `zope.ini` reference
and I postponed other WSGI configuration tests.

Do you think that such tests would be beneficial?
Would `ZPublisher` be an appropriate place for such tests?
If you answer twice with "yes", I think `test_paste` could be retained
as container for future paste/WSGI configuration tests.
If you would put such tests elsewhere (they depend on `setup.py`
for `egg:` references and potentially `Zope2.utilities:skel`
for server, application tests),
I agree that `test_paste` should get renamed.

src/ZPublisher/tests/test_paste.py
d-maurer: I have renamed `test_paste`.

Should we find in the future that more `paste` related tests are beneficial and `ZPublisher` is a good place, we can then potentially restore the old name.

PR #1178, opened 2023-10-26
Separate ZODB connection information into new ZODB Connections view

dataflake:
This PR creates a new Control Panel tab "ZODB Connections" that contains the ZODB connection information shown at the bottom of the former "Debug" page. Debug is renamed to "Reference Counts". The goal is to make this information easier to access and read.

When Zope is running and you visit the "Debug" page several times the page can get very long and getting to the ZODB connection information is inconvenient and cumbersome.

The new "ZODB Connections" page also aims to show more useful information at a glance. It hides the full request information at first and reveals it when you click on the request URI that is shown. The request data is also formatted and sorted now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.7 (unreleased)
 ------------------
 
+- Separate ZODB connection information into new ZODB Connections view
+
 - Update the Ace editor in the ZMI.
 
 - Restrict access to static ZMI resources.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.7 (unreleased)
 ------------------
 
-- Separate ZODB connection information into new ZODB Connections view
+- Separate ZODB connection information into new ZODB Connections view.
+
+- Move the cache detail links to the individual database pages.
+
+- Fix the auto refresh functionality on the Reference Count page
 
 - Update the Ace editor in the ZMI.
 
--- a/src/App/ApplicationManager.py
+++ b/src/App/ApplicationManager.py
@@ -238,8 +238,7 @@ class DebugManager(Tabs, Traversable, Implicit):
 InitializeClass(DebugManager)
 
 
-class ApplicationManager(CacheManager,
-                         Persistent,
+class ApplicationManager(Persistent,
                          Tabs,
                          Traversable,
                          Implicit):
@@ -316,7 +315,7 @@ class ApplicationManager(CacheManager,
         return getConfiguration().clienthome
 
 
-class AltDatabaseManager(Traversable, UndoSupport):
+class AltDatabaseManager(CacheManager, Traversable, UndoSupport):
     """ Database management DBTab-style
     """
     id = 'DatabaseManagement'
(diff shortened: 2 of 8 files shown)


dataflake:
@drfho Any help making it more pretty is appreciated ;-)

d-maurer (commented):
I like the separation.

But something is wrong with the buttons at the bottom of `DebugInfo/manage_main`: any of them activates automatic refresh while only one suggests this in its name (and the "auto-refresh" makes an analysis quite difficult). Likely, the bug was there before. Nevertheless, we might want to fix it.

Now that `DebugInfo` has a more narrow aim, we might want to rename it, e.g. to `DebugRefcounts`.

The buttons at the bottom of `DebugInfo/manage_main` affect the refcount display. They are independent from the "cache" related links. I suggest to put the "cache" related links after the buttons (or even move them to a new tab "CacheDebug").

dataflake:
Thanks for the feedback @d-maurer, I will look at that today or tomorrow. About renaming the `DebugInfo` ID, I tried to make sure no existing bookmarks etc break. I personally don't mind renaming it. I just don't know whose expectations or bookmarks that might break.

drfho (commented):
hi @dataflake , added some more spacing and typography styles.

ZODB_Connections

(maybe I should do that for tab "DebugInfo" as well...)

dataflake:
@drfho Thanks for the help! Question: I am wondering why it says "GET None" for the request URI. Is the request value for `REQUEST_URI` empty in your setup? Does the request value `PATH_INFO` contain the expected path `/Control_Panel/ZODBConnections/manage_main`?

dataflake:
@drfho Please update your sandbox, I have made a change that should make the correct request path show in your environment.

dataflake:
> But something is wrong with the buttons at the bottom of `DebugInfo/manage_main`: any of them activates automatic refresh while only one suggests this in its name (and the "auto-refresh" makes an analysis quite difficult). Likely, the bug was there before. Nevertheless, we might want to fix it.

Fixed.

> The buttons at the bottom of `DebugInfo/manage_main` affect the refcount display. They are independent from the "cache" related links. I suggest to put the "cache" related links after the buttons (or even move them to a new tab "DebugCache").

I moved those cache links to the place where the rest of the cache information is found, the Database page.

drfho:
> @drfho Please update your sandbox,

@dataflake, thanks, now the requesting path is visible:

image

Note: I do not how much connections are expected; if only 4 the table need not be forced into the narrow spacing (CSS-class `table.table-sm`); but if the number may be much higher, table-sm is perfect.

dataflake:
> Note: I do not how much connections are expected; if only 4 the table need not be forced into the narrow spacing (CSS-class `table.table-sm`); but if the number may be much higher, table-sm is perfect.

This is a matter of configuration. The default is 7 maximum connections and up to 4 are in active use so they would show on that page. The number can be increased, but I believe this is rarely done.

d-maurer:
drfho wrote at 2023-10-26 04:11 -0700:
> ...
>Note: I do not how much connections are expected; if only 4 the table need not be forced into the narrow spacing (CSS-class `table.table-sm`); but if the number may be much higher, table-sm is perfect.

The number depends on the `pool_size` (default 7) of `ZODB.DB`
and the number of worker threads (often 2 or 4).

`ZODB.DB` removes closed connections if the connection number
exceeds `pool_size`. However, temporarily, the connection number
can be larger.

Usually, a worker thread uses a single ZODB connection.
Therefore, the worker thread number is usually an upper limit
for the number of connections.
However, in special cases, ZODB connections can be opened
by other threads or a worker thread may open an additional ZODB
connection.

A large number (larger than 10) of connections should be rare.

drfho (approved):
LGTM

davisagli:
Nice improvement!

Issue #1177, opened 2023-10-25
D.O.S. Protection Limits


gogobd:
Hello, every 01!

I just came across 

```
FORM_MEMORY_LIMIT = 2 ** 20   # memory limit for forms
FORM_DISK_LIMIT = 2 ** 30     # disk limit for forms
FORM_MEMFILE_LIMIT = 2 ** 12  # limit for `BytesIO` -> temporary file switch
```

In `ZPublisher/HTTPRequest.py` and I have a question: how do these values relate, "memory limit" is responsible for limiting the request "stream", but I don't understand the documentation of "disk limit" and "memfile limit".

```
    <key name="form-memory-limit" datatype="byte-size" default="1MB">
      <description>
       The maximum size for each part in a multipart post request,
       for the complete body in an urlencoded post request
       and for the complete request body when accessed as bytes
       (rather than a file).
      </description>
    </key>

    <key name="form-disk-limit" datatype="byte-size" default="1GB">
      <description>
       The maximum size of a POST request body
      </description>
    </key>

    <key name="form-memfile-limit" datatype="byte-size" default="4KB">
      <description>
       The value of form variables of type file with larger size 
       are stored on disk rather than in memory.
      </description>
    </key>
```

If "disk limit" is the request body size, why is it called "disk" limit?

Do I understand it right that "memfile limit" is a "switch" so that smaller data are stored in memory whereas data larger than 4k are being written to disk?

Additionally I wanted to report that our employees were unable to upload image files with the very restrictive default values after an update; could we consider to raise the form memory limit to say 10MB or more? Will the "memfile" limit of 4k lead to problems with uploads?

Thank you for clarification!

d-maurer:
Georg Gogo. BERNHARD wrote at 2023-10-25 03:05 -0700:
> ...
>If "disk limit" is the request body size, why is it called "disk" limit?

The names come originally from `multipart`.

`multipart` maintains parts in memory when they are sufficiently small
(controlled by `form_memory_limit`) and stores them on disk otherwise.
The `form-disk-limit` is the maximal amount of disk space usable
for a form.


>Do I understand it right that "memfile limit" is a "switch" so that smaller data are stored in memory whereas data larger than 4k are being written to disk?

Right.

>
>Additionally I wanted to report that our employees were unable to upload image files with the very restrictive default values after an update; could we consider to raise the form memory limit to say 10MB or more? Will the "memfile" limit of 4k lead to problems with uploads?

[...]

PR #1180, opened 2023-10-31
Improved/corrected documentation for the DOS protection options

d-maurer:
#1177 has shown that Zope's DOS protection options documentation is partially wrong and difficult to understand.

This PR significantly extends the documentation by a description of the process influenced by the options and a precise specification of their imposed limits.

@mauritsvanrees #1177 might indicate that `Plone` needs a higher `form-memory-limit` default (currently 1MB). What is your experience?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,11 +10,6 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.7 (unreleased)
 ------------------
 
-- Increase default value for ``form-memory-limit`` from 1MB to 8MB
-  to avoid the need for ``Plone`` to configure this option.
-  See
-  `#1180 <https://github.com/zopefoundation/Zope/pull/1180>`_.
-
 - Separate ZODB connection information into new ZODB Connections view.
 
 - Move the cache detail links to the individual database pages.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.7 (unreleased)
 ------------------
 
+- Increase default value for ``form-memory-limit`` from 1MB to 8MB
+  to avoid the need for ``Plone`` to configure this option.
+  See
+  `#1180 <https://github.com/zopefoundation/Zope/pull/1180>`_.
+
 - Separate ZODB connection information into new ZODB Connections view.
 
 - Move the cache detail links to the individual database pages.
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -53,7 +53,7 @@ from .cookie import getCookieValuePolicy
 
 # DOS attack protection -- limiting the amount of memory for forms
 # probably should become configurable
-FORM_MEMORY_LIMIT = 2 ** 23   # memory limit for forms
+FORM_MEMORY_LIMIT = 2 ** 20   # memory limit for forms
 FORM_DISK_LIMIT = 2 ** 30     # disk limit for forms
 FORM_MEMFILE_LIMIT = 2 ** 12  # limit for `BytesIO` -> temporary file switch
 
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -53,7 +53,7 @@ from .cookie import getCookieValuePolicy
 
 # DOS attack protection -- limiting the amount of memory for forms
 # probably should become configurable
-FORM_MEMORY_LIMIT = 2 ** 20   # memory limit for forms
+FORM_MEMORY_LIMIT = 2 ** 23   # memory limit for forms
 FORM_DISK_LIMIT = 2 ** 30     # disk limit for forms
 FORM_MEMFILE_LIMIT = 2 ** 12  # limit for `BytesIO` -> temporary file switch
 
--- a/src/Zope2/Startup/wsgischema.xml
+++ b/src/Zope2/Startup/wsgischema.xml
@@ -82,27 +82,65 @@
 
   <sectiontype name="dos_protection">
 
-    <description>Defines parameters for DOS attack protection</description>
+    <description>Options for DOS attack protection.
+
+       These options limit the amount of memory and disk resources for
+       the request processing by Zope itself
+       (in contrast to that by the application or the frontend WSGI server).
+
+       Note that the WSGI server (or WSGI middleware) may already have 
+       consumed similar resources. Those resources, too, should
+       likely get limited by corresponding configuration options.
+
+       Zope's own request processing consists of the determination
+       of the request parameters and access to the request body.
+
+       Request parameters can come from a query string and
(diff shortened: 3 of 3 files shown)


d-maurer:
The failing "Windows py3.7" test is:
```
Failure in test test_scored_search (Products.ZCatalog.tests.test_catalog.TestScoring)
...
  File "d:\a\zope\zope\.tox\py37\eggs\products.zcatalog-7.0-py3.7.egg\products\ZCatalog\tests\test_catalog.py", line 1120, in test_scored_search
    self.assertEqual(brains[0].title, '111')
...
AssertionError: '11' != '111'
```

We know that some `ZCatalog` tests are not robust wrt. load conditions and therefore can occasionally fail. But, I have not yet known that the above test belongs to them. I will restart the tests and see what happens.

dataflake (approved):

mauritsvanrees (approved):
The updated documentation LGTM, thanks.

Dieter Maurer wrote:
> @mauritsvanrees #1177 might indicate that `Plone` needs a higher `form-memory-limit` default (currently 1MB). What is your experience?

In my experience, only sites with `plone.restapi` installed (this includes all sites using the default Volto frontend) are affected.  In such a site, if you try to upload a file or image larger than 1MB, you get an error:

```
zExceptions.BadRequest: data exceeds memory limit
```

See also https://github.com/plone/Products.CMFPlone/issues/3848.

On a few sites I have increased the form-memory-limit to 10MB, which at least seems fair for images.  I wondered about doing a patch in `Products.CMFPlone` to update this limit, but then we would have to be careful to not override a manual setting in `zope.conf`.  I would be happy with a higher limit in Zope.

In a pure Plone ClassicUI site without `plone.restapi` installed, with the default settings I can upload a 22MB file without problem.  There the uploaded file is read in a way that does not trigger the limits.

With `plone.restapi` the error already happens when calling `extractCredentials` on the PAS plugin that comes with the package.  The [`json_body` function](https://github.com/plone/plone.restapi/blob/5f9214950a50d4728d7a245bec1358a68c8f2316/src/plone/restapi/deserializer/__init__.py#L8) that is used there should probably be changed along the lines of your suggestions in https://github.com/zopefoundation/Zope/issues/1177#issuecomment-1784889996.  This function is used all over the place though, so this could have unwanted side effects.

I think I will create a PR for `plone.restapi` that catches the `BadRequest` error in the PAS plugin.  That fixes the basic error for ClassicUI at least.

mauritsvanrees:
> Based on the feedback from @mauritsvanrees, the PR now also increases the default for `form-memory-limit` from 1 MB to 8 MB. Hopefully, this will avoid the need for `Plone` to configure this option.

Thanks!

mauritsvanrees:
I am happy with this.  @d-maurer Can you merge?

@dataflake @icemac  A Zope release with this would be good.  Would that be possible today or tomorrow?  I am trying to release Plone 6.0.8 and want to have the increased limit in there, either with a new Zope version or with a temporary patch in CMFPlone.

dataflake:
@mauritsvanrees I don't mind doing releases but I would really appreciate more warning than "today or tomorrow". I am sure releasing Plone 6.0.8 didn't occur to you just today, right?

d-maurer:
Maurits van Rees wrote at 2023-10-31 04:02 -0700:
>On a few sites I have increased the form-memory-limit to 10MB, which at least seems fair for images.  I wondered about doing a patch in `Products.CMFPlone` to update this limit, but then we would have to be careful to not override a manual setting in `zope.conf`.  I would be happy with a higher limit in Zope.

@mauritsvanrees
The parameter protects against potential denial of service attacks
by overloading a system's RAM resources.

I meanwhile learned that `plone.restapi` wants to load
complete file contents into RAM (a design fault in my view).
Should you want to keep this (hopefully only temporarily),
you could override `ZPublisher.HTTPRequest.VALUE_LIMIT`
(currently `Global("FORM_MEMORY_LIMIT")`) by `None`.
This would allow arbitrary large request bodies accessed
via `request["BODY"])`.

mauritsvanrees:
@dataflake Sorry for not communicating earlier.

The problem with the low memory limit was already known (to Plone people) since shortly after the Plone 6.0.7 release from 21 September.  Since then some fixes were done to counter this (in `plone.recipe.zope2instance` and `plone/cookiecutter-zope-instance`), but they still required a configuration change.  I had hoped that someone (most likely a Plone core developer) would have picked this up already, no matter if that would be by a change in Zope or in Plone.  But that did not happen yet.  So much for wishful thinking...

Meanwhile there actually is already a Plone 6.0.8 release candidate since last Thursday.  I wanted to do the final release two days ago.  But finally I decided we should not ship with this low default limit, so I postponed the release (also for one other, totally unrelated problem).

d-maurer:
Maurits van Rees wrote at 2023-11-2 03:44 -0700:
>I am happy with this.  @d-maurer Can you merge?

I would like to still wait a bit.

The current state will still not make all Plone admins happy as
only uploads up to 8 MB will be possible without reconfiguration.
The current state of `plone.restapi` requires that
access to `request["BODY"]` does not have a size protection
(or a very large one (in the order of GB)).

We could remove this size protection in `Zope` or
you could do this in `Products.CMFPlone`.
Or we could add an additional parameter (e.g. `REQUEST_BODY_MEMORY_LIMIT`)
with a quite large default.

We should find consensus about those options before the PR gets
merged.

mauritsvanrees:
`plone.restapi` definitely needs some changes along the lines you have indicated here and elsewhere.

As a quick fix, I think I will open a PR in `plone.restapi` to override `FORM_MEMORY_LIMIT` with a larger value (maybe 16MB) **if** it is at the current default of 1MB.  Then only users of `plone.restapi` are affected, and the configuration options are still all there.

d-maurer:
> As a quick fix, I think I will open a PR in `plone.restapi` to override `FORM_MEMORY_LIMIT` with a larger value (maybe 16MB) **if** it is at the current default of 1MB. Then only users of `plone.restapi` are affected, and the configuration options are still all there.

Maybe, overriding `ZPublisher.HTTPRequest.ZopeFieldStorage.VALUE_LIMIT` is a better option. If `plone.restapi` sets it to `None`, then there is no longer a size limitation for `request["BODY"]` accesses.

d-maurer:
I removed the default change for `form_memory_limit` again as @mauritsvanrees  plans to resolve the Plone related problems elsewhere.

mauritsvanrees:
> Maybe, overriding `ZPublisher.HTTPRequest.ZopeFieldStorage.VALUE_LIMIT` is a better option. If `plone.restapi` sets it to `None`, then there is no longer a size limitation for `request["BODY"]` accesses.

Okay, let's go with that after all.  Thanks.
See https://github.com/plone/plone.restapi/pull/1729

dataflake:
@d-maurer OK, I suggest merging now

@mauritsvanrees Do you still need a Zope 5.8.7 release?

mauritsvanrees:
@dataflake No need for a Zope release, the `plone.restapi` change will be enough for now.  Thanks anyway.

FWIW, I tried the Zope master branch and its versions earlier this week, and all Plone tests passed.

PR #1183, opened 2023-11-06
support form data in PUT requests

### LINKED ISSUE #1182: Clarification on  Handling Image Uploads via HTTP PUT in Zope5 with WSGI Server
### What I did:

I encountered an issue related to the handling of image uploads when using the HTTP PUT method in Zope5 with a WSGI server. Here are the steps I followed to reproduce the problem:

1. Using a Zope5 installation with the WSGI server.
2. Sending an HTTP PUT request to upload an image.
3. Observing that the image data is received in the request body as encoded data.

### What I expect to happen:

I expected that, similar to Zope4 with the ZServer, I would receive the image data as part of the form, and I would be able to access it as an object. This is the behavior I was accustomed to in Zope4.

### What actually happened:

In Zope5 with the WSGI server, when I use the HTTP PUT method to upload an image, the image data is received in the request body, and the body data is encoded. As a result, I am unable to access the image data as a file upload object, which is different from the behavior in Zope4.

- Operating System: Windows/Linux
- Python Version: 3.10.9
- Zope Version: 5.8.3

This report aims to highlight the difference in behavior between Zope4 and Zope5 when using HTTP PUT for image uploads and to understand whether this is the expected behavior for PUT operations.

Additional Observation:
https://github.com/zopefoundation/Zope/blob/5.8.3/src/ZPublisher/HTTPRequest.py#L1425
If i include "PUT" condition here, it is working as expected

Thank you.

d-maurer:
Fixes #1182.

Unlike `multipart`, Zope did not support form data in `PUT` requests. This likely has been a bug: whether a request contains form data is not determined by the request method but by the content type.

This PR brings Zope in line with `multipart`: form data is processed for `POST` and `PUT` requests  controlled by the content type. Like `multipart`, form data is not processed for other request methods (even if the content type indicates form data) because `multipart` would reject those request methods.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.8.7 (unreleased)
 ------------------
 
+- Support form data in ``PUT`` requests (following the ``multipart`` example).
+  Fixes `#1182 <https://github.com/zopefoundation/Zope/issues/1182>`_.
+
 - Separate ZODB connection information into new ZODB Connections view.
 
 - Move the cache detail links to the individual database pages.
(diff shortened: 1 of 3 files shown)


d-maurer:
The "release check" fails because `setup.py" does not yet declare Python 3.12 support -- we can ignore this for the moment.

dataflake (approved):

PR #1184, opened 2023-11-20
Officially support Python 3.12.

icemac:
- [x] Maybe we should also update some other version pins to versions officially supporting Python 3.12.
- [ ] There are many deprecation warnings in Python 3.12 for the latest Chameleon release, maybe we can get a newer one, if it is already fixed there, see https://github.com/zopefoundation/Zope/actions/runs/6926782925/job/18839495396?pr=1184


dataflake:
The deprecation warnings still exist on the latest Chameleon GHA test runs.

+1 on re-running the checkversions script and adding new pins before a release.

dataflake (approved):

mtoepfl:
+1 on updating version pins

icemac:
@dataflake Could you please take over updating the version pins and cutting the release?

dataflake:
Yes, I can do that some time this weekend.

dataflake:
I found time today, Zope 5.9 is now live.

PR #1192, opened 2024-01-09
Fix redirections to URLS with host given as IP-litteral

### LINKED ISSUE #1191: Invalid redirects to URLs with IPv6 literals addresses
### What I did:

When zope redirects to an IPv6 URL, for example with a python script with:

```py
return container.REQUEST.RESPONSE.redirect('http://[fe80::1ff:fe23:4567:890a]:1234')
```

the brackets surrounding the IPv6 litteral are percent-escaped in the `Location` header.

Chrome accepts it for some reason, but most clients do not understand this (firefox, curl, requests, urllib3).

Getting the URL with curl shows how thte `[` and `]` are percent-encoded:

```
...
< location: http://%5Bfe80::1ff:fe23:4567:890a%5D:1234
...
curl: (3) The redirect target URL could not be parsed: Port number was not a decimal number between 0 and 65535
```

Instead of  `http://%5Bfe80::1ff:fe23:4567:890a%5D:1234` ,  `http://[fe80::1ff:fe23:4567:890a]:1234` would have been correct.

The quoting of `netloc` part of the URL was implemented for https://github.com/zopefoundation/Zope/pull/1032 after first quoting the path only in https://github.com/zopefoundation/Zope/pull/586 and this seems to be based on RFC 2396, but that RFC has been obsoleted by RFC 3986 which specifies that the host can be given with surrounding brackets:

      host        = IP-literal / IPv4address / reg-name
      IP-literal = "[" ( IPv6address / IPvFuture  ) "]"

The fix I am considering is to not escape `[` and `]` in host ( I plan to submit a pull request a bit later ).

### What I expect to happen:

`[` and `]` not percent encoded in the host ( `http://[fe80::1ff:fe23:4567:890a]:1234` )

### What actually happened:

`[` and `]` were percent encoded (
  `http://%5Bfe80::1ff:fe23:4567:890a%5D:1234`  )

happens on master branch

perrinjerome:
When redirecting to an URL with an IPv6 host with surrounding brackets,
we should not escape the surrounding brackets.

The patch updates referenced RFC from 2396 to 3986, which obsoletes it
and change the safe characters for the netloc part to allow [ and ].
The RFC specifies that [ and ] are only allowed when they are the first
and last characters, but we don't need to be more specific here, because
using [ or ] in other places of the host is rejected by urlparse above.

Fixes #1191

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 - Officially support Python 3.12.1.
   (`#1188 <https://github.com/zopefoundation/Zope/issues/1188>`_)
 
+- Fix redirections to URLs with host given as IP-literal with brackets.
+  Fixes `#1191 <https://github.com/zopefoundation/Zope/issues/1191>`_.
+
 
 5.9 (2023-11-24)
 ----------------
--- a/src/ZPublisher/HTTPResponse.py
+++ b/src/ZPublisher/HTTPResponse.py
@@ -235,24 +235,24 @@ class HTTPBaseResponse(BaseResponse):
         # To be entirely correct, we must make sure that all non-ASCII
         # characters are quoted correctly.
         parsed = list(urlparse(location))
-        rfc2396_unreserved = "-_.!~*'()"  # RFC 2396 section 2.3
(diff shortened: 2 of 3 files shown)


d-maurer (approved):

dataflake (approved):

perrinjerome:
Thank you for reviewing

PR #1196, opened 2024-02-17
Fix authentication error viewing ZMI with a user defined outside of zope root

### LINKED ISSUE #1195: Authentication error viewing ZMI with a user defined outside of zope root
## BUG/PROBLEM REPORT / FEATURE REQUEST

`manage_page_header` makes a `head` like this:

```html
<!doctype html>
<html lang="en">
<head>

<meta http-equiv="content-type" content="text/html;charset=utf-8" />

<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />

<title>Zope</title>

	<link rel="stylesheet" type="text/css" href="/++resource++zmi/bootstrap-4.6.0/bootstrap.min.css" />
	<link rel="stylesheet" type="text/css" href="/++resource++zmi/fontawesome-free-5.15.2/css/all.css" />
	<link rel="stylesheet" type="text/css" href="/++resource++zmi/zmi_base.css" />
	<script src="/++resource++zmi/jquery-3.5.1.min.js"></script>
	<script src="/++resource++zmi/bootstrap-4.6.0/bootstrap.bundle.min.js"></script>
	<script src="/++resource++zmi/ace.ajax.org/ace.js"></script>
	<script src="/++resource++zmi/zmi_base.js"></script>

```

Since the "Restrict access to static ZMI resources." changes from ce5288a1b (Update ACE editor 1.4.4 to 1.5.0 (#1166), 2023-10-11), the links for zmi require view management screens permission.

Because the link is like this:

```html
	<script src="/++resource++zmi/jquery-3.5.1.min.js"></script>
```

in practice they require view management screens permissions **at the root of zope**.

### What I did:

1. create a folder `subfolder`
2. create a user folder in that folder
3. create a user `u` with manager role in `/subfolder/acl_users`
4. login as `u` and visit `/subfolder/manage_main`

### What I expect to happen:

view management screens

### What actually happened:

401 error while getting CSS and javascripts

this started from 5.9 and is still the case on current master.

---

I'll send a pull request, I think we can change the links generation logic to include the authentication path from the user. In the example case the link would be:

```html
	<script src="/subfolder/++resource++zmi/jquery-3.5.1.min.js"></script>
```

perrinjerome:
Fixes #1195

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 - Fix redirections to URLs with host given as IP-literal with brackets.
   Fixes `#1191 <https://github.com/zopefoundation/Zope/issues/1191>`_.
 
+- Fix authentication error viewing ZMI with a user defined outside of zope root.
+  Fixes `#1195 <https://github.com/zopefoundation/Zope/issues/1195>`_.
+
 
 5.9 (2023-11-24)
 ----------------
--- a/src/App/dtml/manage_page_header.dtml
+++ b/src/App/dtml/manage_page_header.dtml
@@ -10,24 +10,27 @@
 </dtml-let>
 
 <title><dtml-if title_or_id><dtml-var title_or_id><dtml-else>Zope</dtml-if></title>
+<dtml-let basepath="'/'.join([''] + [p for p in (REQUEST['BASEPATH1'], REQUEST.get('AUTHENTICATION_PATH')) if p])">
+
 <dtml-in css_urls>
-	<link rel="stylesheet" type="text/css" href="&dtml-BASEPATH1;&dtml-sequence-item;" />
+	<link rel="stylesheet" type="text/css" href="&dtml-basepath;&dtml-sequence-item;" />
 </dtml-in>
 <dtml-in js_urls>
-	<script src="&dtml-BASEPATH1;&dtml-sequence-item;"></script>
+	<script src="&dtml-basepath;&dtml-sequence-item;"></script>
 </dtml-in>
 
-<link rel="shortcut icon" type="image/x-icon" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon.ico" />
-<link rel="apple-touch-icon" sizes="180x180" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/apple-touch-icon.png" />
-<link rel="icon" type="image/png" sizes="32x32" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon-32x32.png" />
-<link rel="icon" type="image/png" sizes="16x16" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon-16x16.png" />
-<link rel="manifest" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/site.webmanifest" />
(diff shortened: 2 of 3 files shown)


d-maurer:
Jérome Perrin wrote at 2024-2-16 22:23 -0800:
>
>Fixes #1195
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/zopefoundation/Zope/pull/1196

Your PR changes the URL the ZMI resources are accessed with
taking into account the authentication path.

While this solves the problem you have observed,
it seems suboptimal to me: quite complex and less HTTP cache efficient.
I would prefer a different approach: make the necessary resources
public.
@dataflake
I am aware that this goes against
"https://github.com/zopefoundation/Zope/pull/1166#issuecomment-1752469143"
but is it really a security concern to have public access to the CSS and JS
resources of the ZMI?

perrinjerome:
Thank you

> it seems suboptimal to me: quite complex and less HTTP cache efficient. I would prefer a different approach: make the necessary resources public.

Yes, this is a bit complex. Frankly, I would also prefer to make the resources public, I don't see why they need to be protected.

dataflake:
I absolutely don't like publishing a whole bunch of JS out of a basic Zope install that none of us can audit for issues.

dataflake:
To expand on my earlier answer, I don't care about the CSS. I am concerned about the JS. If this was just a few script files we maintained ourselves that would be OK. But we have several megabytes of stuff in there, including the full Ace editor package, bootstrap, jquery and fontawesome. None of us has any idea what all that code does. Just publishing it freely for convenience seems a bad trade-off for me.

Jerome's solution looks OK to me, it preserves the safety barrier. I don't think caching is really a valid concern. We're talking a ZMI, an application that requires authentication everywhere. It doesn't make much sense to try and cache things there.

perrinjerome:
You probably understood that, but I want to clarify this about the caching concerns: My first naive idea was to make the resources relatives to the current folder. This would have fixed the original issue, but it would have been really bad for cache, because each folder visited in the ZMI would use different URLs for js and css. By making the resources relative to the authentication path like it's implemented here, each admin user always uses the same URLs for js and css, so they can be cached in the browser cache.

dataflake (dismissed):

perrinjerome:
I resolved a conflict, it would be good to decide something here. @dataflake approved a previous version of the changes. The argument that it's better to be safe and not publish javascript that we don't know sounds reasonable (at least the risk is limited to ZMI).

 @d-maurer you're also OK ?

I don't really have another idea how to fix this.

d-maurer:
Jérome Perrin wrote at 2024-2-27 01:23 -0800:
>I resolved a conflict, it would be good to decide something here. @dataflake approved a previous version of the changes. The argument that it's better to be safe and not publish javascript that we don't know sounds reasonable (at least the risk is limited to ZMI).
>
> @d-maurer you're also OK ?

I am not convinced by @dataflake 's arguments but do not care
enough. Go ahead.

perrinjerome:
Thank both of you

mauritsvanrees:
This fixes the problem for the normal case, so thanks.  But this gives wrong urls when using virtual hosting.  A Zope Manager gets this link in the head:

    href="/plone/virtual_hosting/++resource++zmi/zmi_base.css"

A Plone manager gets this link in the head:

    href="/plone/virtual_hosting///++resource++zmi/zmi_base.css"

I opened a new issue #1203 and show a partial fix so the urls are correct.

But even with correct links, I get an Unauthorized when using virtual hosting, so the root cause is not really this PR.

PR #1198, opened 2024-02-17
Fix Content-Disposition header for clients without rfc6266 support

perrinjerome:
Since https://github.com/zopefoundation/Zope/pull/893 the Content-Disposition header supports non-ascii filenames, by containing the filename in ascii and the filename in UTF-8, but the ascii version was produced by applying `str` on a `bytes` instance, so it looks like `b'file.txt'` instead of `file.txt`.


d-maurer (approved):

dataflake (approved):

perrinjerome:
Thank you

PR #1204, opened 2024-04-12
Fix authentication error viewing ZMI resource with virtual hosting

### LINKED ISSUE #1203: Authentication error viewing ZMI resource with virtual hosting.
This is similar to issue #1195 which has more links, and which has been solved.  But the solution from PR #1196 makes a problem when using virtual hosting worse.

Setup:

* Zope 5.9, Plone 6.0.10, Python 3.11.
* Zope runs on port 9341.
* The Plone site is at `/plone`
* One user account with the Manager role defined on the Zope root, and another defined on the Plone site with Manager role there.
* nginx runs on port 9340 and passes all traffic on to the Plone site object.  The main line that does this, is this rewrite rule:

```
rewrite ^/(.*)$ /VirtualHostBase/http/$host:9340/plone/VirtualHostRoot/$1 last;
```

With Zope 5.9:

* Zope Manager can access ZMI fine when directly accessing Zope.
* Zope Manager sees ZMI unstyled when going via nginx.  I get an Unauthorized on for example http://localhost:9340/++resource++zmi/zmi_base.css
* Plone Manager sees the ZMI unstyled in both cases.

Issue #1195 recognises the last problem and PR #1196 fixes it.  I copy the fixed `src/App/dtml/manage_page_header.dtml` from Zope master and try again:

* Both the Zope and Plone Manager can see the ZMI fine when directly accessing Zope.  The Plone user obviously cannot access the Zope root, but on the Plone site the resource urls are correctly like http://localhost:9341/plone/++resource++zmi/zmi_base.css
* Both the Zope and Plone Manager see the ZMI unstyled when going via nginx.

The Plone user has this link in the head:

    href="/plone/virtual_hosting///++resource++zmi/zmi_base.css"

The Zope user has this link in the head:

    href="/plone/virtual_hosting/++resource++zmi/zmi_base.css"

I have a patch on the copied dtml file that improves this:

```
< <dtml-let basepath="'/'.join([''] + [p for p in (REQUEST['BASEPATH1'], REQUEST.get('AUTHENTICATION_PATH')) if p])">
---
> <dtml-let authpath="REQUEST.get('AUTHENTICATION_PATH', '').split('virtual_hosting', maxsplit=1)[-1].lstrip('/')"
>           basepath="'/'.join([''] + [p for p in (REQUEST['BASEPATH1'], authpath) if p]).lstrip('/')"
> 	  basepath="'/' + basepath if basepath else ''">
```

This turns the head links for both users into this when using virtual hosting (and correctly leaves the non-virtual host case unchanged):

    href="/++resource++zmi/zmi_base.css"

So this is an improvement for the virtual hosting case.

But the problem still remains: these urls give an Unauthorized when accessed through nginx, for both Manager users:

```
Unauthorized: http://localhost:9340/++resource++zmi/zmi_base.css
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 391, in publish_module
  Module ZPublisher.WSGIPublisher, line 269, in publish
  Module ZPublisher.BaseRequest, line 641, in traverse
  Module ZPublisher.HTTPResponse, line 1016, in unauthorized
zExceptions.unauthorized.Unauthorized: You are not authorized to access this resource.
```

Wait a minute.  This is weird.  I get that Unauthorized error for four ZMI resources and exactly one minute later, for each of those plus one extra resource, the nginx access log reports this:

```
*9 upstream timed out (110: Connection timed out) while reading response header from upstream, client: 10.200.209.1, server: localhost, request: "GET /++resource++zmi/zmi_base.css HTTP/1.1", upstream: "http://127.0.0.1:9346/VirtualHostBase/http/localhost:9340/plone/VirtualHostRoot/++resource++zmi/zmi_base.css", host: "localhost:9340", referrer: "http://localhost:9340/manage_main"
```

And now Zope hangs.  Ah, this is only after I have edited the `error_log` to not filter out `Unauthorized`.  When I restart and filter it out again, Zope no longer hangs.  Maybe this is caused by `Products.PDBDebugMode`, which is installed in this test site.  So never mind that.

Main problem is, for the Zope Manager user:

1. This works: http://localhost:9341/++resource++zmi/zmi_base.css
2. This works: http://localhost:9341/plone/++resource++zmi/zmi_base.css
3. This gives Unauthorized, although it should really be the same as the second one: http://localhost:9340/++resource++zmi/zmi_base.css

Question: why?  And how do we fix this?

perrinjerome:
Fixes #1203

### DIFF
--- a/src/App/dtml/copyright.dtml
+++ b/src/App/dtml/copyright.dtml
@@ -9,13 +9,13 @@
   <link rel="stylesheet" type="text/css" href="&dtml-BASEPATH1;/++resource++zmi/bootstrap-4.1.1/bootstrap.min.css" />
   <link rel="stylesheet" type="text/css" href="&dtml-BASEPATH1;/++resource++zmi/zmi_base.css" />
 
-  <link rel="shortcut icon" type="image/x-icon" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon.ico" />
-  <link rel="apple-touch-icon" sizes="180x180" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/apple-touch-icon.png" />
-  <link rel="icon" type="image/png" sizes="32x32" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon-32x32.png" />
-  <link rel="icon" type="image/png" sizes="16x16" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/favicon-16x16.png" />
-  <link rel="manifest" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/site.webmanifest" />
-  <link rel="mask-icon" href="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/safari-pinned-tab.svg" color="#5bbad5" />
-  <meta name="msapplication-config" content="&dtml-BASEPATH1;/++resource++zmi/logo/favicon/browserconfig.xml"/>
+  <link rel="shortcut icon" type="image/x-icon" href="&dtml-BASEPATH1;/++resource++logo/favicon/favicon.ico" />
+  <link rel="apple-touch-icon" sizes="180x180" href="&dtml-BASEPATH1;/++resource++logo/favicon/apple-touch-icon.png" />
+  <link rel="icon" type="image/png" sizes="32x32" href="&dtml-BASEPATH1;/++resource++logo/favicon/favicon-32x32.png" />
+  <link rel="icon" type="image/png" sizes="16x16" href="&dtml-BASEPATH1;/++resource++logo/favicon/favicon-16x16.png" />
+  <link rel="manifest" href="&dtml-BASEPATH1;/++resource++logo/favicon/site.webmanifest" />
+  <link rel="mask-icon" href="&dtml-BASEPATH1;/++resource++logo/favicon/safari-pinned-tab.svg" color="#5bbad5" />
+  <meta name="msapplication-config" content="&dtml-BASEPATH1;/++resource++logo/favicon/browserconfig.xml"/>
   <meta name="msapplication-TileColor" content="#2d89ef" />
   <meta name="theme-color" content="#ffffff" />
 </head>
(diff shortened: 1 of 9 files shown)


src/zmi/styles/subscriber.py
```diff
@@ -1,23 +1,61 @@
...
+
+
+def prepend_authentication_path(request: HTTPRequest, path: str) -> str:
+    """Prepend the path of the user folder.
+
+    Because ++resource++zmi is protected, we generate URLs relative to the
+    user folder of the logged-in user, so that the user can access the
+    resources.
+    """
+    authenticated_user = request.get("AUTHENTICATED_USER")
```
dataflake, line 16: Please do not pick the user from the `REQUEST`. That's convenient but a request can be manipulated. Do it like this instead:

```
from AccessControl.SecurityManagement import getSecurityManager
<...>
authenticated_user = getSecurityManager().getUser()
```

dataflake (changes requested):

src/zmi/styles/subscriber.py
perrinjerome: Thanks ! I amended the commit accordingly

dataflake (dismissed):

src/App/dtml/copyright.dtml
```diff
@@ -9,13 +9,13 @@
   <link rel="stylesheet" type="text/css" href="&dtml-BASEPATH1;/++resource++zmi/bootstrap-4.1.1/bootstrap.min.css" />
```
mauritsvanrees, line 9: Not really a problem of this PR, but this is actually points to a non-existing bootstrap version.  So let's fix it.

```suggestion
  <link rel="stylesheet" type="text/css" href="&dtml-BASEPATH1;/++resource++zmi/bootstrap-4.6.0/bootstrap.min.css" />
```

mauritsvanrees (approved):
Seems to work. Thanks!

One small inline suggestion, but that is not a problem caused by this PR, so I approve.

src/App/dtml/copyright.dtml
perrinjerome: Thanks, I added this change in the same PR

perrinjerome:
after applying the suggestion of fixing bootstrap url in copyright page, your approvals became "stale", so I had to ask review again.

mauritsvanrees (approved):

perrinjerome:
Thanks you !

Issue #1202, opened 2024-04-02
 Deprecating docstring control: Zope Start may get blocked by Products


drfho:
Hello @d-maurer,
I am sorry to report the following error launching zope:
after merging "**Preliminary step to deprecate docstring based publication control**" https://github.com/zopefoundation/Zope/pull/1197 into main-branch https://github.com/zopefoundation/Zope/commit/3b32e2e07d1cd180caf6dab5db3d58d8460874b8 Zope fails starting with a type error calling [Products.CMFCore](https://github.com/zopefoundation/Products.CMFCore)

I am not sure where the problem may be located: 
a bug due to the code-change of Zope or Products.CMFCore must get compatible with the new code?
Any hints/comments?

## Log
```
~/vpy38_zms/bin$ runwsgi -v ~/instance/zms5_main/etc/zope.ini
__REGISTRY__['confdict'] None
2024-04-02 22:12:54 INFO [chameleon.config:39][MainThread] directory cache: /home/zope/instance/zms5_main/var/cache.
Traceback (most recent call last):
  File "/home/zope/vpy38_zms/bin/runwsgi", line 8, in <module>
    sys.exit(main())
  File "/home/zope/src/zopefoundation/Zope/src/Zope2/Startup/serve.py", line 251, in main
    return command.run()
  File "/home/zope/src/zopefoundation/Zope/src/Zope2/Startup/serve.py", line 189, in run
    app = self.loadapp(app_spec, name=app_name, relative_to=base,
  File "/home/zope/src/zopefoundation/Zope/src/Zope2/Startup/serve.py", line 220, in loadapp
    return loadapp(app_spec, name=name, relative_to=relative_to, **kw)
  File "/home/zope/vpy38_zms/lib/python3.8/site-packages/paste/deploy/loadwsgi.py", line 246, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/home/zope/vpy38_zms/lib/python3.8/site-packages/paste/deploy/loadwsgi.py", line 271, in loadobj
    return context.create()
  File "/home/zope/vpy38_zms/lib/python3.8/site-packages/paste/deploy/loadwsgi.py", line 738, in create
    return self.object_type.invoke(self)
  File "/home/zope/vpy38_zms/lib/python3.8/site-packages/paste/deploy/loadwsgi.py", line 198, in invoke
    app = context.app_context.create()
[...]

PR #1205, opened 2024-04-23
Work around `Products.CMFCore` and `Products.CMFPlone` design bug

d-maurer:
Work around for #1202.

`Products.CMFCore` and `Products.CMFPlone` register non callable constructors with `App.ProductContext.ProductContext.registerClass` (even though the method's docstring clearly states that constructors must be callable).
Formerly, this was not a problem. But with the introduction of explicit zpublishable decoration, constructors are now required to be callable braking `Products.CMFCore` and `Products.CMFPlone`.

The PR introduces the new parameter `resources` of `registerClass`  which should be used instead of `constructors` to register non callable objects. It uses work around code to avoid braking `Products.CMFCore` and `Products.CMFPlone` (and maybe other packages) when they abuse the `constructors` parameter for non callable objects; a `DeprecationWarning` is issued in this case.

In addition, the PR cleans up `registerClass`'s docstring a bit: in the old days, when the product registry could still be persistent (controlled via a parameter), the constructors were required to be picklable. The feature was abandoned long ago (because of many problems) and consequently, constructors need no longer be picklable (indeed, the constructors generated for zpublishable decoration are not picklable). The PR removes the corresponding requirement from the docstring..

Finally, the PR removes the mention from the docstring that the `instance_class`  parameter were not used (which is obviously wrong). But, I am not sure whether this removal is optimal. It is possible that the author wanted to express that `instance_class` is not used directly (to construct instances) but only to provide useful defaults for other parameters. Feedback welcome.

I would like to stress that this is a work around for bugs in other packages. They should get fixed there and the work around code here eventually removed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,11 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 - Fix authentication error viewing ZMI with a user defined outside of zope root.
   Fixes `#1195 <https://github.com/zopefoundation/Zope/issues/1195>`_.
 
+- Work around ``Products.CMFCore`` and ``Products.CMFPlone`` design bug
+  (registering non callable constructors).
+  For details, see
+  `#1202 <https://github.com/zopefoundation/Zope/issues/1202>`_.
+
 
 5.9 (2023-11-24)
 ----------------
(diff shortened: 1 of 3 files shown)


dataflake (approved):

drfho (approved):
works fine / LGTM / many thanks

mauritsvanrees (approved):
I have not looked at the implementation details, but I confirm this fix works.
And I agree this should be fixed in CMFCore and CMFPlone soon.

PR #1211, opened 2024-05-16
Update three packages.

mauritsvanrees:
Plone currently overrides three versions compared to Zope master.
I know you prefer updating the versions once after a Zope release and then test with those until the next Zope release.  But maybe these can be considered.  I don't mind, just a suggestion.

* Latest certifi seems a logical one.
* mr.developer has a small fix: https://github.com/fschulze/mr.developer/blob/master/CHANGES.rst
* zope.testrunner has fixes for Python 3.13 and native namespaces: https://github.com/zopefoundation/zope.testrunner/blob/master/CHANGES.rst

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -15,7 +15,7 @@ Pygments = 2.17.2
 Record = 4.0
 Sphinx = 7.2.6
 alabaster = 0.7.13
-certifi = 2023.11.17
+certifi = 2024.2.2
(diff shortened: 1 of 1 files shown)


dataflake (approved):
Yes, we still prefer version changes after a release so they get some exposure and testing before the next release. You don't want to deal with issues right after a public release. But this looks fine.

PR #1213, opened 2024-05-17
Define custom content type for Page Template (fixes #1212)

### LINKED ISSUE #1212: WebDAV PUT factory behavior differs between platforms
Zope is currently relying on the fact that creating an object through WebDAV PUT with a name ending in `.pt` produces a Zope Page Template. This is no longer true on all platforms. Newer Linux versions (reproducable on the latest Ubuntu and RHEL-based platforms) seem to associate that file name extension with content of type `application/vnd.snesdev-page-table`, so the test for the default PUT factory now fails.

See https://github.com/zopefoundation/zope.contenttype/issues/17

I'm hoping to get this resolved quickly so the Plone folks get a new Zope release (-> #1210). One option would be to add specific name to object type rules into the default PUT factory code instead of deferring to `zope.contenttype`.

dataflake:
Fixes #1212 

I chose to put the `mime.types` file into the PageTemplates product itself because I didn't see a sensible place elsewhere where I can load it during initialization.

### DIFF
--- a/src/webdav/NullResource.py
+++ b/src/webdav/NullResource.py
@@ -114,9 +114,7 @@ class NullResource(Persistent, Implicit, Resource):
 
         if ext == '.dtml':
             ob = DTMLDocument('', __name__=name)
-        elif typ in ('text/html',
-                     'text/xml',
-                     'application/vnd.zopefoundation.pagetemplate'):
+        elif ext in ('.pt', '.zpt') or typ in ('text/html', 'text/xml'):
             ob = ZopePageTemplate(name, body, content_type=typ)
         elif typ.startswith('image/'):
             ob = Image(name, '', body, content_type=typ)
--- a/src/webdav/NullResource.py
+++ b/src/webdav/NullResource.py
@@ -114,7 +114,9 @@ class NullResource(Persistent, Implicit, Resource):
 
         if ext == '.dtml':
             ob = DTMLDocument('', __name__=name)
-        elif typ in ('text/html', 'text/xml'):
+        elif typ in ('text/html',
+                     'text/xml',
+                     'application/vnd.zopefoundation.pagetemplate'):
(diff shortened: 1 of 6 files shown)


d-maurer (dismissed):

src/webdav/NullResource.py
```diff
@@ -114,7 +114,9 @@ def _default_PUT_factory(self, name, typ, body):
 
         if ext == '.dtml':
             ob = DTMLDocument('', __name__=name)
-        elif typ in ('text/html', 'text/xml'):
+        elif typ in ('text/html',
+                     'text/xml',
+                     'application/vnd.zopefoundation.pagetemplate'):
```
d-maurer, line 119: Alternatively, we could apply the same logic as for `.dtml`, e.g. use
```python
elif ext in ('.pt', '.zpt') or typ in (...
```

mauritsvanrees: Relying on the extension might indeed be better.

Otherwise I wonder what happens when the operating system associates `.pt` with that other mime type you found.  Does the new Zope definition win?  Does the OS definition win?  Does it matter on the order in which the two definitions get loaded?

d-maurer:
Maurits van Rees wrote at 2024-5-17 12:32 -0700:
> ...
>Relying on the extension might indeed be better.
>
>Otherwise I wonder what happens when the operating system associates `.pt` with that other mime type you found.  Does the new Zope definition win?

All depends on Python's `mimetypes` module.
This promises that later registered type files take precedence over
former ones. The first type files registered are those provided
by the system.
Thus, the Zope definition will win unless the application
registers additional files with conflicting definitions.

> Does the OS definition win?  Does it matter on the order in which the two definitions get loaded?

In principle, yes: but the system provided files are registered
before any others.

dataflake:
I chose the simplest possible solution.

PR #1214, opened 2024-05-28
Drop Python 3.7 support and update meta templates

dataflake:
Creating as draft because of the macOS issues.


dataflake:
Possibly related: https://github.com/zopefoundation/meta/issues/247

dataflake:
The errors are all a re-manifestation of https://github.com/buildout/buildout/issues/609 and https://github.com/zopefoundation/meta/issues/181. zc.buildout is broken when using it from a Python that has been compiled to use `universal2` architecture on macOS because it has a really dumb algorithm to match the system architecture.

In the past we have worked around it by just excluding GHA tests for Python versions we know wouldn't work anymore. At this point, with newer macOS runners that are all using `universal2`, this isn't enough anymore. Unless we stop testing on macOS altogether.

Nothing is broken for Python installs that are built for the native x86_64 or arm64 architecture. Just universal2.

Not sure what to do right now.

d-maurer:
Jens Vagelpohl wrote at 2024-5-28 02:57 -0700:
>The errors are all a re-manifestation of https://github.com/buildout/buildout/issues/609 and https://github.com/zopefoundation/meta/issues/181. zc.buildout is broken when using it from a Python that has been compiled to use `universal2` architecture on macOS because it has a really dumb algorithm to match the system architecture.
>
>In the past we have worked around it by just excluding GHA tests for Python versions we know wouldn't work anymore. At this point, with newer macOS runners that are all using `universal2`, this isn't enough anymore. Unless we stop testing on macOS altogether.
>
>Nothing is broken for Python installs that are built for the native x86_64 or arm64 architecture. Just universal2.
>
>Not sure what to do right now.

One approach could be to fix `zc.buildout`.

When I remember right, then older `zc.buildout` versions delegated
to `easy_install` (at least I have seen this name often in tracebacks).
An upcoming `zc.buildout` could delegate to a modern installer
(e.g. `pip`) in the hope that this can cope with modern platforms.

dataflake:
Yes, fixing zc.buildout would be great. And making it not use easy_install. I wish someone would do that, I do not want to become a zc.buildout and/or setuptools maintainer myself.

My latest thought was to try and switch the tox template to use pip and straight zope.testrunner instead of installing dependencies and building the convenience test script with buildout. Some of the configuration templates already do it that way.

dataflake:
This PR uses changed meta templates from https://github.com/zopefoundation/meta/pull/248 to work around the macOS issues. Local tox runs are not affected unless you manually add `-universal2` to the end of a tox environment name.

.meta.toml
```diff
@@ -31,6 +31,20 @@ additional-config = [
...
+    "universal2: zope.interface",
+    "universal2: zope.security",
+    "universal2: zope.container",
+    "universal2: Persistence",
+    "universal2: Acquisition",
+    "universal2: AccessControl",
+    "universal2: zodbpickle",
+    "universal2: charset_normalizer",
+    "universal2: MarkupSafe",
+    "universal2: zope.testrunner",
```
icemac, line 47: `zope.testrunner` does not contain any C code, it should not have a `universal2` wheel.

.meta.toml
```diff
@@ -39,11 +53,13 @@ testenv-commands-pre = [
     "{envbindir}/buildout -c {toxinidir}/buildout.cfg buildout:directory={envdir} buildout:develop={toxinidir} buildout:root-directory={toxinidir}",
     ]
 testenv-commands = [
+    "# The universal2 environment only runs Zope's own tests",
+    "universal2: {envdir}/bin/test {posargs:-vc}",
+    "# All other environments run the expanded alltests script",
```
icemac, line 58: Why?

src/Zope2/utilities/mkwsgiinstance.py
```diff
@@ -227,7 +227,7 @@ def get_zope2path(python):
     try:
         output = subprocess.check_output(
             [python, '-c', 'import Zope2; print(Zope2.__file__)'],
-            universal_newlines=True,  # makes Python 3 return text, not bytes
+            text=True,  # makes Python 3 return text, not bytes
```
icemac, line 230:
```suggestion
            text=True,
```
We are on Python-3-only so this comment should no longer be needed.

versions-prod.cfg
```diff
@@ -81,21 +81,3 @@ zope.testbrowser = 6.0
 zope.testing = 5.0.1
 zope.traversing = 5.0
 zope.viewlet = 5.0
-
-[versions:python38]
-# Chameleon >= 4.5 requires Python 3.9
-Chameleon = 4.4.4
```
icemac, line 87: This should still be there even though we currently do not use version 4.5+ otherwise we have to find out it again when updating.

icemac (approved):
LGTM

.meta.toml
dataflake: This is not about creating a universal2 wheel for publishing. It's about running in a universal2 environment, which the macOS GHA runners are.

.meta.toml
dataflake: Because when pre-installing the additional packages in the universal2 environment the alltests script will point to the virtual environment's site-packages folder multiple times, which means all kinds of undesired tests are run (and fail). If you know a better solution I am all ears.

versions-prod.cfg
dataflake: I agree, it just so happens that Chameleon 4.4.4 is the latest version that lets the tests run so the pins would be the same for Python 3.8 and everything else. I put it back in.

There is a newer release of Chameleon but the tests fail. (different issue for a different day)

d-maurer:
Jens Vagelpohl wrote at 2024-5-29 09:00 -0700:
> ...
>Because when pre-installing the additional packages in the universal2 environment the alltests script will point to the virtual environment's site-packages folder multiple times, which means all kinds of undesired tests are run (and fail). If you know a better solution I am all ears.

`zope.testrunner` already has the option `require-unique`.
This means that it maintains a structure able to detect (apparent) test
"duplicate"s.
It would be easy to use this structure to implement the option
`skip-test-duplicates` to not run a test with a test id already known.

dataflake:
> `zope.testrunner` already has the option `require-unique`. This means that it maintains a structure able to detect (apparent) test "duplicate"s. It would be easy to use this structure to implement the option `skip-test-duplicates` to not run a test with a test id already known.

The problem is not duplicate tests. The problem is that it tries to test everything in the virtual environment's site-packages, which doesn't work and is not the desired outcome.

d-maurer:
Jens Vagelpohl wrote at 2024-5-29 09:18 -0700:
>> `zope.testrunner` already has the option `require-unique`. This means that it maintains a structure able to detect (apparent) test "duplicate"s. It would be easy to use this structure to implement the option `skip-test-duplicates` to not run a test with a test id already known.
>
>The problem is not duplicate tests. The problem is that it tries to test everything in the virtual environment's site-packages, which doesn't work and is not the desired outcome.

Likely, there is a solution for this, too:

 * Formerly, you could use "bin/instance test" to test packages
   excluding standard ones.
   It worked with a complex negative module filter, excluding
   all modules which should not get tested

 * Alternatively, we could try the option `ignore-dir`.
   It tells the runner where not to look for tests.

dataflake:
If you want to specify folders to ignore there's the problem that you would need to synthesize what folders to ignore in the tox configuration. This is not a fixed value. It's the site-packages folder of the Python version that is being used to run the tests.

Before spending too much time on this please consider what we're testing. The alltests script tests Zope plus an arbitrary selection of other packages. There's no real rhyme or reason to that selection of additional packages. And obviously those packages are all tested with Zope in their own package test configurations already. And the alltest script would only be replaced by the standard test script for macOS runners. So there's no real loss. The other runners will have tested this already.

.meta.toml
d-maurer: I agree that we can avoid the "foreign" (= not genuine Zope) tests in the `Mac` environment: the tests are executed (even on `Mac`) as part of the checks of their respective project; proper interoperability with Zope is checked in the other (non `Mac`) environments and interoperability problems should not be environment specific.

.meta.toml
```diff
@@ -39,11 +53,13 @@ testenv-commands-pre = [
...
     "# the `layer` argument below suppresses a `ZCatalog` performance test",
     "# which occasionally fails on GITHUB",
-    "{envdir}/bin/alltests --layer '!Products.PluginIndexes' {posargs:-vc}",
+    "!universal2: {envdir}/bin/alltests --layer '!Products.PluginIndexes' {posargs:-vc}",
```
d-maurer, line 61: Did you not want to avoid `alltests` for `universal2`?

By the way: we see above the use of a negative filter (indicated by the leading `!`), applied here to `layer`. Negative filtering is possible for the module, too, and can (among others) be used to exclude tests from dedicated packages (should we want to run `alltests` for `universal2` as well).

constraints.txt
```diff
@@ -2,16 +2,14 @@ AccessControl==6.3
 Acquisition==5.2
 AuthEncoding==5.0
 BTrees==5.2
-Chameleon==4.2.0; python_version == '3.7'
 Chameleon==4.4.4; python_version == '3.8'
 Chameleon==4.4.4; python_version > '3.8'
```
d-maurer, line 6: Same version for `3.8` and `> 3.8`?

requirements-full.txt
```diff
@@ -3,16 +3,14 @@ AccessControl==6.3
 Acquisition==5.2
 AuthEncoding==5.0
 BTrees==5.2
-Chameleon==4.2.0; python_version == '3.7'
 Chameleon==4.4.4; python_version == '3.8'
 Chameleon==4.4.4; python_version > '3.8'
```
d-maurer, line 7: Same version for `3.8` and `> 3.8`?

constraints.txt
dataflake: Yes, please see https://github.com/zopefoundation/Zope/pull/1214#discussion_r1619117320

requirements-full.txt
dataflake: Yes, please see https://github.com/zopefoundation/Zope/pull/1214#discussion_r1619117320

.meta.toml
dataflake: The negative filter means the alltests script is not executed for the universal2 environment. I don't understand the question.

d-maurer: While I get frequently invited to review GHA test related modifications, I am no expert: I have not recognized the `!` as a negative filter.

d-maurer (dismissed):

.meta.toml
dataflake: This is actually a `tox` feature. When I say e.g. `tox -e py310-universal2` tox will run the Python 3.10 tests and what's after the dash in the environment name will be directly visible to the tox configuration and I can filter by this name. This filter can be used in many places, like adding/removing step dependencies or commands.

PR #1194, opened 2024-02-16
Preliminary support for Python 3.13a3

icemac:
Open issues:

- [x] Pin `RestrictedPython` to a version which can be installed on Python 3.13, probably `7.0a2.dev0` or a newly to be created dev release.
- [x] Needs https://github.com/zopefoundation/zodbpickle/pull/83 to be merged and released.
- [x] Need a release of zope.container after https://github.com/zopefoundation/zope.container/pull/53 is merged.

PR #1215, opened 2024-05-30
Preliminary support for Python 3.13, second try

dataflake:
Supercedes #1194 - it was easier to re-do the branch from master.

The Python 3.13 tests reveal another big hurdle for Zope on Python 3.13. zope.testbrowser relies on the WebOb and WebTest projects. None of those have seen any activity for the last 3 years and WebOb is incompatible with Python 3.13 because it uses the now-gone `cgi` module.


setup.py
```diff
@@ -122,6 +122,7 @@ def _read_file(filename):
         'zope.traversing',
         'zope.viewlet',
         'multipart',
+        'legacy-cgi; python_version > "3.12"',
```
dataflake, line 125: This is an experiment. The dependency must go into `zope.testbrowser` instead. See https://github.com/zopefoundation/zope.testbrowser/pull/108. This will need merging and then a release that is pulled in here so this dependency can go.

dataflake: Removed the dependency, it's now moved into zope.testbrowser itself.

dataflake:
So just for clarification, these changes do not declare official Python 3.13 support yet. They just set up everything so test run on Python 3.13.

For declaring official support RestrictedPython needs to be vetted and released - see https://github.com/zopefoundation/RestrictedPython/issues/264

d-maurer (approved):

PR #1217, opened 2024-06-11
update dependencies

perrinjerome:


perrinjerome:
It seems there's an error with latest version of chameleon, this does not seem to be a known problem, is it ?

d-maurer:
> It seems there's an error with latest version of chameleon, this does not seem to be a known problem, is it ?

I know that @malthe has made changes to `z3c.pt` to make it compatible with new `Chameleon` versions (define some interfaces no longer directly defined by `Chameleon`). Not sure, whether `z3c.pt` got a new release after this change (and you use it).

dataflake:
@perrinjerome We usually handle dependency updates for Zope separately after a release or when there's a good technical reason to do so. Also, right now there's going to be more updates to dependency packages than usual so it creates more work to keep bringing PRs with Zope dependency updates. I suggest putting this aside for the moment and concentrating on the dependencies itself.

perrinjerome:
Thanks, we can definitely put this on hold, this was just a draft pull request, initially I wanted to see if we can update `zope.exceptions` here already for https://github.com/zopefoundation/Products.PythonScripts/pull/65 and I tried to update everything to see if that was OK, but all this is not urgent.

The error also happens with latest z3c.pt

PR #1219, opened 2024-06-12
Return 404 instead of IndexError for traversal past the root

### LINKED ISSUE #1218: Traversal with `..` past the root raises IndexError
### What I did:

1. Start Zope on port 8080
2. Try to load `http://localhost:8080/..` using curl: `curl -v --path-as-is 'http://localhost:8080/..'`

(Note, this is hard to reproduce in a browser because it may normalize the path before it's sent to the server. We're seeing this error from bot activity.)

### What I expect to happen:

I'd expect a 404 response since you can't navigate "up" from the root.

### What actually happened:

This raises an IndexError and returns a 500 response.

```
  File "/app/lib/python3.11/site-packages/ZPublisher/WSGIPublisher.py", line 181, in transaction_pubevents
    yield
  File "/app/lib/python3.11/site-packages/ZPublisher/WSGIPublisher.py", line 391, in publish_module
    response = _publish(request, new_mod_info)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/ZPublisher/WSGIPublisher.py", line 269, in publish
    obj = request.traverse(path, validated_hook=validate_user)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/ZPublisher/BaseRequest.py", line 381, in traverse
    del clean[-1]
        ~~~~~^^^^
IndexError: list assignment index out of range
```

Zope 5.9, Python 3.11.3, macOS

davisagli:
Fixes #1218

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,10 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
 - Update to newest compatible versions of dependencies.
 
-- Added CC-BY 4.0 license to the Zope logo. 
+- Added CC-BY 4.0 license to the Zope logo.
+
+- Fix ``IndexError`` on traversal past the root using `..`.
+  (`#1218 <https://github.com/zopefoundation/Zope/issues/1218>`_)
 
 
 5.10 (2024-05-18)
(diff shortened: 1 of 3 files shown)


tseaver (approved):

davisagli:
I updated the branch protection rules to require the "coverage/coveralls (push)" check instead of "coverage/coveralls", since the name seems to have changed.

Issue #1220, opened 2024-07-27
Align docutils version with sphinx_rtd_theme requirement


stevepiercy:
With the upcoming release of `sphinx_rtd_theme` 3.0, [docutils will get bumped to the following](https://github.com/readthedocs/sphinx_rtd_theme/pull/1576/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R48):

```python
	docutils >0.18,<0.22
```

Should we remove the docutils pin in `versions.cfg` and allow `sphinx_rtd_theme` to handle its installation, or should we bump accordingly?

FTR, I ran into a version conflict when trying to install `plone.restapi` with buildout, and it took me a while to hunt down where it fell down:

```
Version and requirements information containing docutils:
  [versions] constraint on docutils: 0.18.1
  Base installation request: 'zest.releaser[recommended]', 'zest.pocompile', 'zestreleaser.towncrier', 'towncrier', 'readme', 'docutils', 'zpretty', 'i18ndude', 'coverage'
  Requirement of readme-renderer: docutils>=0.21.2
  Requirement of pyroma; extra == "recommended": docutils
  Requirement of readme-renderer[md]>=40: docutils>=0.21.2
While:
  Installing dependencies.
Error: The requirement ('docutils>=0.21.2') is not allowed by your [versions] constraint (0.18.1)
```

davisagli:
@stevepiercy From what I see here, Zope's constraint for docutils (`0.18.1`) is compatible with the requirement from `spinx_rtd_theme` (`docutils >0.18,<0.22`).

The problem is that zest.releaser pulls in readme-renderer, and readme-renderer 0.44 requires `docutils>=0.21.2`

One way to resolve this would be to pin readme-renderer to an older version that still supports docutils 0.18.

Another way would be to add support for docutils 0.21.2 in sphinx_rtd_theme, and then update or remove the docutils pin here.

stevepiercy:
> Another way would be to add support for docutils 0.21.2 in `sphinx_rtd_theme`, and then update or remove the docutils pin here.

That's my preferred option, and I expect it will happen soon. See the linked PR in this issue description. I would prefer the removal of the pin from Zope, but maybe there is a valid reason to pin it similar to `sphinx_rtd_theme` v3.0.

dataflake:
We pin all dependencies here, so a pin will stay. We can update it, sure.

dataflake:
[...]

PR #1221, opened 2024-08-07
Change Sphinx theme used for Zope documentation

dataflake:
In a discussion about the never ending annoyance that the currently used `sphinx_rtd_theme` forbids the latest `docutils` (-> https://github.com/zopefoundation/Zope/issues/1220) Steve Piercy suggested two other themes, the `plone-sphinx-theme` that is now used for Plone 6 documentation, and `furo`, another lightweight and modern theme.

This PR updates all compatible dependencies and switches the documentation theme to `furo`. You can test it all locally and look at the `furo`-themed documentation by running `bin/tox -edocs` and then opening the page at `docs/_build/html/index.html`.

This PR is a suggestion, not a "must". I do like the theme and as a maintainer I do like the fact that I have to stop worrying about dependency restrictions that make no sense.

I have taken a quick look at `plone-sphinx-theme` as well, but it explicitly only supports Python 3.9 and up with no fallback for Python 3.8. I respect the Plone team's decision to drop Python 3.8 support, but we're not there yet. Our documentation should build on all Python versions supported by Zope.


constraints.txt
```diff
@@ -11,11 +11,11 @@ MultiMapping==5.0
...
-RestrictedPython==7.1; python_version == '3.10'
-RestrictedPython==7.1; python_version == '3.11'
-RestrictedPython==7.1; python_version == '3.12'
-RestrictedPython==7.1; python_version == '3.8'
-RestrictedPython==7.1; python_version == '3.9'
+RestrictedPython==7.2; python_version == '3.10'
+RestrictedPython==7.2; python_version == '3.11'
+RestrictedPython==7.2; python_version == '3.12'
+RestrictedPython==7.2; python_version == '3.8'
+RestrictedPython==7.2; python_version == '3.9'
```
stevepiercy, line 18: My OCD made me twitch, seeing an alpha-only sort, instead of alphanumeric. More motivation to drop Python < 3.10 lol.

docs/requirements.txt
```diff
@@ -1,5 +1,4 @@
 Sphinx
-sphinx_rtd_theme>1
-docutils<0.19
+furo
```
stevepiercy, line 2: If you pin docutils in `versions.cfg`, should you also pin it here? Same for Sphinx.

versions.cfg
```diff
@@ -13,44 +13,45 @@ Products.BTreeFolder2 = 5.1
...
 zc.recipe.testrunner = 3.0
-zipp = 3.18.2
-zope.testrunner = 6.4
+zipp = 3.19.2
+zope.testrunner = 6.5
 
 [versions:python38]
 # alabaster >= 0.7.14 requires Python 3.9
 alabaster = 0.7.13
+# Sphinx <8 requires docutils <0.21
```
stevepiercy, line 53: Align with pin in `versions.cfg`. See https://github.com/sphinx-doc/sphinx/blob/master/pyproject.toml#L66

```suggestion
# Sphinx 8 requires docutils>=0.20,<0.22
```

versions.cfg
```diff
@@ -13,44 +13,45 @@ Products.BTreeFolder2 = 5.1
...
+docutils = 0.20.1
 # Required dependency on Python 3.8 only
 importlib-resources = 6.4.0
 # Sphinx >= 7.2 requires Python 3.9+
```
stevepiercy, line 57: This line and the next conflict with earlier lines.

stevepiercy (commented):
I'm not sure about the conflicting versions of docutils and Sphinx in these changes. Should they be consistent or different?

versions.cfg
```diff
@@ -50,6 +50,8 @@ zope.testrunner = 6.5
 [versions:python38]
 # alabaster >= 0.7.14 requires Python 3.9
 alabaster = 0.7.13
+# Sphinx <8 requires docutils <0.21
+docutils = 0.20.1
```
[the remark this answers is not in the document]
dataflake: As the section header says, this is for Python 3.8. Sphinx 8 is not supported there, so docutils must remain at the version supported by Sphinx 7, which is 0.20.1

constraints.txt
dataflake: This list is autogenerated from the information in `versions-prod.cfg` and `versions.cfg`. It is not maintained by hand, so this doesn't matter much.

versions.cfg
dataflake: No, I suggest this line stay the way it is. Mentioning "Sphinx <8" ties this to the Sphinx 7.x pin just a few lines down, and there is no need to mention a lower bound for the pin, it's always "the highest that will work".

versions.cfg
dataflake: Not sure what you mean. This is a Python 3.8-specific pin so yes, it will be different from the higher Python versions that do support the latest packages.

dataflake:
> I'm not sure about the conflicting versions of docutils and Sphinx in these changes. Should they be consistent or different?

They are different for Python 3.8 because Sphinx 8 does not support Python 3.8. That means there must be a separate lower pin for Sphinx, which in turn requires a separate lower pin for docutils. The section with the lower pins only applies on Python 3.8.

stevepiercy (approved):
Thanks for the explanation. I overlooked the section header.

constraints.txt
```diff
@@ -49,25 +49,25 @@ zope.cachedescriptors==5.0
...
 zope.event==5.0
-zope.exceptions==5.0.1
+zope.exceptions==5.1
 zope.filerepresentation==6.0
 zope.globalrequest==2.0
 zope.hookable==6.0
 zope.i18n==5.1
 zope.i18nmessageid==6.1.0
-zope.interface==6.4.post2
+zope.interface==7.0.1
```
d-maurer, line 65: Should we wait with the use of `zope.interface>=7.0`? "https://github.com/zopefoundation/zope.interface/issues/315" reports a segmentation fault (for the use of `zope.interface.7.0.1` in `Django`.

requirements-full.txt
```diff
@@ -50,25 +50,25 @@ zope.cachedescriptors==5.0
...
 zope.event==5.0
-zope.exceptions==5.0.1
+zope.exceptions==5.1
 zope.filerepresentation==6.0
 zope.globalrequest==2.0
 zope.hookable==6.0
 zope.i18n==5.1
 zope.i18nmessageid==6.1.0
-zope.interface==6.4.post2
+zope.interface==7.0.1
```
d-maurer, line 66: See  `constraints.txt`.

d-maurer (approved):

constraints.txt
dataflake: Good catch. That error report sounds seriously strange and it's unclear if this is Django-specific, so out of an abundance of caution I reverted that pin.

icemac (approved):
LGTM from reading through the diff. I did not take a look at the result.

mauritsvanrees:
Thanks!

For those who want to compare:

* The Zope 4 docs use the old theme: https://zope.readthedocs.io/en/4.x/
* The Zope 5 docs use the new theme: https://zope.readthedocs.io/en/latest/

dataflake:
If you compare https://zope.readthedocs.io/en/4.x/INSTALL.html and https://zope.readthedocs.io/en/latest/INSTALL.html you can also see how local tables of content are handled. They are stuck in their fixed location on the page with the RTD theme and thus scroll out of view while remaining in view on the right hand side with the `furo` theme.

stevepiercy:
Dark mode and better typography in Furo eases my eye strain, too. The right-side navigation also has scroll-spy, highlighting the current location within the page.

dataflake:
"scroll spy" - cool term. Never heard that before.

stevepiercy:
I first saw it Twitter Bootstrap. https://getbootstrap.com/docs/3.4/javascript/#scrollspy

PR #1224, opened 2024-08-16
Run isort, autopep8 and flake8 in pre-commit.

icemac:
I excluded pre-commit from running locally, so the changes it made show up as a commit here in the PR. (I had to make some little changes which cannot be done automatically (too long line, indentation problems)

Open tasks:

- [ ] ~~Change log~~ this is just an interal change, no need to tell the world.
- [x] Remove `isort`, `autopep8`, `flake8` from `tox.ini` (they need to run only once.)
- [x] Update meta/config, so it can produce the changes in this PR
- [x] require `pre-commit` job in GHA to succeed (via meta/config)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,3 +1,5 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/zope-product
 name: pre-commit
 
 on:
@@ -5,9 +7,15 @@ on:
   push:
     branches:
     - master
+  # Allow to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+env:
+  FORCE_COLOR: 1
+
 jobs:
   pre-commit:
-    name: pre-commit linting
+    name: linting
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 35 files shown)


dataflake (approved):
This is a readable PR, thanks.

By the way, this PR caught some files that are not included in the current tox configuration, like `utils.py` in the project root. So there's a discrepancy between a local tox run and what this PR has done that should be fixed at the same time.

icemac:
I am going to remove these checks from tox, as running them once should be enough and we do not have the burden to maintain it in two places.

dataflake:
There will be something left, either in tox or with instructions how to do it, to execute these checks before a push, right?

jugmac00:
I usually run pre-commit via tox, so there can't be a discrepancy between the local checks and those in CI, and then you even do not need to install pre-commit locally, but you certainly can.

dataflake:
Right, I'd be interested in some invocation - preferably out of tox since that is the de-facto standard we use - to run the same steps locally that the GitHub action does.

dataflake (approved):
By the way, I removed `ubuntu-lint` from the branch protection rule for `master` and added `linting`

icemac:
> By the way, I removed `ubuntu-lint` from the branch protection rule for `master` and added `linting`

This was the last missing piece I just added to https://github.com/zopefoundation/meta/pull/265.

icemac:

PR #1227, opened 2024-08-29
Try out zc.buildout 3.1.0

icemac:


dataflake:
The Windows failure seems to point to an incompatibility between `cffi` 1.17.0 and `setuptools` version 74. I bet if you pin `setuptools<74` it will work.

dataflake:
The `cffi` folks have an open PR, but who knows how long they will take for a release: https://github.com/python-cffi/cffi/pull/118

dataflake:
Is that new pin acceptable? If yes I would add it to the other templates as well.

mauritsvanrees (approved):

dataflake (approved):

icemac:
> Is that new pin acceptable? If yes I would add it to the other templates as well.

So you were right in https://github.com/zopefoundation/meta/pull/274 that `setuptools 74` is not yet usable. So this is fine for now until `cffi` has fixed the bug on their end.

icemac:

PR #1236, opened 2024-10-30
set up `request["PARENTS"]` before request parameter processing

### LINKED ISSUE #1235: Exceptions during request parameter processing are not processed by `error_log`
If an `error_log` object (from `Products.SiteErrorLog`) exists, it should process (and usually record) exceptions raised during request processing, including exceptions raised by request parameter processing. The latter actually does not happen -- for the following reason:

`Products.SiteErrorLog` looks at `request["PUBLISHED"]` and if this does not exist at `request["PARENTS"]` to determine the error context and uses it to determine which `error_log`  object should record the exception.  Unfortunately, `ZPublisher.WSGIPublisher.publish` calls `request.processInputs` (the request parameter processing) before it sets up `request["PARENTS"]`. As a consequence, `SiteErrorLog` can not determine the error context and does not process the exception.

d-maurer:
Fixes #1235.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.11.1 (unreleased)
 -------------------
 
+ - Define ``request["PARENTS"]`` before request parameter processing
+   such that request failure subscribers (such as that of
+   ``Products.SiteErrorLog``) can determine the error context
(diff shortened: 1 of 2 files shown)


d-maurer:
The failing MAXOS checks are unrelated to this change and tackled in #1234.

dataflake (approved):

davisagli (approved):

PR #1240, opened 2024-11-27
Use modern Python import features for product enumeration

### LINKED ISSUE #1239: Zope needs to change the way it enumerate Products?
When Zope loads Products, it uses code like this:

[Link to the code in Zope](https://github.com/zopefoundation/Zope/blob/04b75cf7956425fe3e4e1f4184bac4d377502c59/src/OFS/Application.py#L418)

```python
for index, product_dir in enumerate(Products.__path__):
        product_names = os.listdir(product_dir)
        for product_name in product_names:
            if _is_package(product_dir, product_name):
                # i is used as sort ordering in case a conflict exists
                # between Product names.  Products will be found as
                # per the ordering of Products.__path__
                products.append((0, product_name, index, product_dir))
```

However the [Python Packaging Guide seems to recommend something different](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-namespace-packages)

```python
import importlib
import pkgutil

import myapp.plugins

def iter_namespace(ns_pkg):
    # Specifying the second argument (prefix) to iter_modules makes the
    # returned name an absolute name instead of a relative one. This allows
    # import_module to work without having to do additional modification to
    # the name.
    return pkgutil.iter_modules(ns_pkg.__path__, ns_pkg.__name__ + ".")

discovered_plugins = {
    name: importlib.import_module(name)
    for finder, name, ispkg
    in iter_namespace(myapp.plugins)
}
```

This seems relevant as I'm currently investigating a bug around imports of namespaces packages installed with recent versions of setuptools and pip, where pip deprecated the hardcoded special handling of `setup.py`/`setup.cfg` without a pyproject.toml file present. [See details here](https://ichard26.github.io/blog/2024/11/whats-new-in-pip-24.3/#psa-legacy-editable-installs-are-deprecated)

Looking at the blame of that code didn't quite allow me to understand the reasoning behind it, and it may very well be that in the past this was the correct way to get the contents of namespace packages.

But it seems this may have changed?

I would like your perspective on wether the old way might be better or if we should migrate to the new way and how to test if that brings up any issues.

d-maurer:
Fixes #1239.

Currently, Zope only supports file based module finders during product enumeration (see `OFS.Application.get_products`).
This PR replaces Zope's  proprietary product enumeration by `pkgutil.iter_modules` and thereby supports all module finder types supported by Python.

To reduce the risk for incompatibilities the signature for product handling functions in `OFS.Application` is mostly left unchanged: `product_dir` is replaced by `finder` (though most functions do not use this parameter in their body).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
   in ``constraints.txt``.
   (`#1234 <https://github.com/zopefoundation/Zope/pull/1234>`_)
 
+- Use modern Python import features for product enumeration
+  (`#1239 <https://github.com/zopefoundation/Zope/issues/1239>`_).
+
 
 5.11.1 (2024-11-03)
 -------------------
--- a/src/OFS/Application.py
+++ b/src/OFS/Application.py
@@ -13,8 +13,8 @@
 """Application support
 """
 
-import pkgutil
 import os
+import pkgutil
 import sys
 from logging import getLogger
 from urllib.parse import urlparse
--- a/src/OFS/Application.py
+++ b/src/OFS/Application.py
@@ -13,7 +13,6 @@
 """Application support
 """
 
-import os
 import pkgutil
 import sys
 from logging import getLogger
--- a/src/OFS/Application.py
+++ b/src/OFS/Application.py
@@ -13,6 +13,7 @@
 """Application support
 """
 
+import pkgutil
 import os
 import sys
 from logging import getLogger
@@ -378,7 +379,7 @@ def install_products(app=None):
     folder_permissions = get_folder_permissions()
     meta_types = []
     done = {}
-    for priority, product_name, index, product_dir in get_products():
+    for priority, product_name, index, finder in get_products():
         # For each product, we will import it and try to call the
         # intialize() method in the product __init__ module. If
         # the method doesnt exist, we put the old-style information
@@ -386,7 +387,7 @@ def install_products(app=None):
         if product_name in done:
             continue
         done[product_name] = 1
-        install_product(app, product_dir, product_name, meta_types,
+        install_product(app, finder, product_name, meta_types,
                         folder_permissions)
 
     # Delayed install of packages-as-products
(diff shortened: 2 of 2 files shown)


dataflake:
The code looks find and the tests run, but how do we really know if the new way to load products will produce the same results as the old way, meaning it won't lead to any products not being loaded?

d-maurer:
Jens Vagelpohl wrote at 2024-11-27 01:42 -0800:
>The code looks fine and the tests run, but how do we really know if the new way to load products will produce the same results as the old way, meaning it won't lead to any products not being loaded?

I trust that `pkgutil.iter_modules` does the right thing
for file loaders.

The original code iterated over `Products.__path__` and selected
in the resulting folders
a subfolder as product if it contained an `__init__.p[yco]`.
It disregarded path elements which were not folders;
therefore, it worked only for file loaders.

`pgkutil.iter_modules` promises to return a `ModuleInfo`
object for each top level module (or package) indicated by `__path__`.
If the path element corresponds to a file loader and a subfolder
contains an `__init__.p[yco]`, then this subfolder represents
a package and therefore should result in the generation
of a corresponding `ModuleInfo`.
Thus, the new approach should lead to at least the products
detected earlier.

The new approach may find more products:
those that do not use a file loader and those that are
implemented via an implicit namespace package.

dataflake (approved):
Thanks for the explanation!

dwt:
That looks about how I would have expected. Thanks al lot!

dwt (approved):

mauritsvanrees:
I have a setup in Plone coredev where this code does not work.  But I think the old code also would not work.  The problem is the `Products.__path__` is `lib/python/site-packages/Products`, and this directory is missing `CMFPlone` because this is in a checkout in `src/Products.CMFPlone`.

See https://github.com/plone/plone.testing/issues/104#issuecomment-2659431283

It could be that this is only the case when you have everything installed with `pip` and use `mxdev` to have some checkouts.  Also, it only poses a problem for the tests.  And it is an easy fix in `plone.testing`, basically: if `get_products` does not find a product, treat is as a normal package and call `OFS.metaconfigure.get_packages_to_initialize`.

So I am not creating a bug report, this is just a note that the Products path may not always be what we think.

dwt:
Out of interest, I am currently trying to switch all my python package manager needs over to uv (because speeed!!) and find that they also explicitly mention that as an alternative for mxdev. Would switching to uv for this use calls also happen to solve this problem?

PR #1241, opened 2024-12-03
Update to `zope.interface = 7.2`.

icemac:


icemac:
Hm, the broken MacOS runs: That's the problem I want to solve for `Products.ZCatalog` – I'd not have expected to see it here again.

d-maurer:
Michael Howitz wrote at 2024-12-2 23:39 -0800:
>Hm, the broken MacOS runs: That's the problem I want to solve for `Products.ZCatalog` – I'd not have expected to see it here again.

Maybe, a (pip-) cache issue?

icemac:
@d-maurer wrote:
> Maybe, a (pip-) cache issue?

Nope, deleted all caches but got same result.

dataflake:
The difference between the broken run and the last good run is that this step:

```
py39-universal2: install_deps> python -I -m pip install AccessControl Acquisition 'cffi>=1.17.0rc1' charset_normalizer MarkupSafe Persistence 'setuptools<74' 'wheel>0.37' 'zc.buildout>=3.1' zodbpickle zope.container zope.interface zope.security zope.testrunner -c /Users/runner/work/Zope/Zope/constraints.txt
```

installs `zope.interface` version 7.1.1 both times. During the last good run that matches what buildout wants to pull later and it's not pulled by buildout. During the bad run buildout doesn't like the already-installed version and wants to get it again, and that breaks.

Honestly, I am just sick and tired of these constant issues, I don't want to deal with it anymore. The tests run fine locally. I don't see how the test failures have any bearing on real life use of Zope. I have literally spent days digging into this and pointing out issues in buildout itself when faced with a `universal2` environment and `setuptools`. None of it got any attention.

dataflake:
I just noticed you forgot to update the requirements and constraints files. This may fix it.

dataflake (approved):

icemac:
> I just noticed you forgot to update the requirements and constraints files. This may fix it.

Sorry, I got so used to the automated way these files are updated in `zopetoolkit` and all the other places where I use automatic version updates, that I forgot to do it here manually.

PR #1251, opened 2025-02-09
Adding FORM_PART_LIMIT for overriding MultipartParser.part_limit

drfho:
Complex modelling GUIs may have lot of form values in a grid. With the introduction of multipart==1.2.1
https://github.com/zopefoundation/Zope/blame/a2cbf86fca81836aac194b1075ed0728e8c0c79b/requirements-full.txt#L27
some weeks ago the arbitrary limit of 128 form parameters is getting relevant. Using the new requirement ZPublisher blocks bigger forms. I would like to suggest handling the parts_limit it similar to the other form limits and add a corresponding variable 

```
FORM_PART_LIMIT = 2 ** 10     # limit for individual form parts
```
that can override the 128 limitation.

Ref: https://github.com/zms-publishing/ZMS/issues/334

### DIFF
--- a/src/ZPublisher/HTTPRequest.py
+++ b/src/ZPublisher/HTTPRequest.py
@@ -55,6 +55,7 @@ from .cookie import getCookieValuePolicy
 
 # DOS attack protection -- limiting the amount of memory for forms
 # probably should become configurable
+FORM_PART_LIMIT = 2 ** 10     # limit for individual form parts
 FORM_MEMORY_LIMIT = 2 ** 20   # memory limit for forms
 FORM_DISK_LIMIT = 2 ** 30     # disk limit for forms
 FORM_MEMFILE_LIMIT = 2 ** 12  # limit for `BytesIO` -> temporary file switch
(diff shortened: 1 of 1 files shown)


dataflake:
I think that's a good idea but not ready for merging. This needs tests and it should be configurable in zope.conf under the `dos_protection` section as well, like the other limits. I may have some time over the next few days.

dataflake:
@drfho I'm currently working on making this configurable. Question: I'm not a frontend developer so I don't know how big those forms can get. Would it not be better to choose a lower default (or even the ``multipart`` default of 128) as a safer starting value? My best guess would be that this excessive number of form parameters would be the exception rather than the rule. IMHO an administrator should have to explicitly take action and configure the higher limit if they really need it.

drfho:
> @drfho 128 as a safer starting value? 

128 is really low; 512 would be good and 1024 quite safe. I think the authors of multiparts are not adressing production GUIs but 3rd view front-ends that are published to the public. Supposingly they didn't have administrative GUIs like Zope, Plone etc in mind.

dataflake:
> 128 is really low; 512 would be good and 1024 quite safe. I think the authors of multiparts are not adressing production GUIs but 3rd view front-ends that are published to the public. Supposingly they didn't have administrative GUIs like Zope, Plone etc in mind.

Yes you're right, I didn't think about mile long folder contents lists in the ZMI.

davisagli:
And the grid of checkboxes in manage_access...

dataflake (approved):
Ready to go I think

drfho:
ok, looks perfect, many thanks! 
Merge is done by @dataflake, right?

dataflake:
No, it is done by the person who opened the PR, unless this is an unknown new developer.

PR #1253, opened 2025-03-01
Drop pkg_resources usage

gforcada:
See https://github.com/plone/Products.CMFPlone/issues/4126


dataflake:
I would argue that you should move the `importlib-metadata` version pin from versions.cfg to versions-prod.cfg since it is no longer an indirect but a direct dependency.

dataflake:
(don't forget to run the buildout or to run the `utils.py` script afterwards to rebuild the requirements and constraints files)

dataflake (approved):
I made those last changes myself, go ahead and merge

dataflake:
I undid my changes - it seems the `importlib-metadata` non-standard library package is indeed just an indirect dependency.

gforcada:
If we are using recent enough python versions we don't need `importlib-metadata` as `importlib.metadata` (note the dot) is in Python standard library.

`importlib-metadata` (with dash) is useful for getting access to APIs on newer python versions if you are stuck on older python versions.

See the table on its PyPI page: https://pypi.org/project/importlib-metadata/

PR #1254, opened 2025-03-11
Add configuration switch to turn off the built-in XML-RPC support

dataflake:
Zope always had support for XML-RPC and other than registering a utility to decide whether to allow it there was no simple way to just turn it off. This PR adds a zope.conf switch to do just that.

The new configuration option is still "on" by default so no one will be surprised. You have to explicitly set it to "off" to disable XML-RPC requests. At that point those are answered by a BadRequest.

I want to add this switch because very few people actually use XML-RPC and I am seeing an increase in malicious and/or spam requests that are detected as XML-RPC in Zope's simple logic (POST with content type `text/xml`) and then cause uncontrolled exceptions in the XML parsing. IMHO everyone should turn this off unless it is really needed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 5.12.1 (unreleased)
 -------------------
 
+- Add configuration switch to turn off the built-in XML-RPC support.
+
 - Add configuration switch for the maximum allowed number of form fields.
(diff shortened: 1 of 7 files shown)


ale-rt (approved):
LGTM, I would bump the minor version as well.

dataflake:
> LGTM, I would bump the minor version as well.

Good catch, thanks!

gforcada (approved):

PR #1255, opened 2025-03-13
Update zope.publisher and zope.testrunner for pkg_resources fixes

mauritsvanrees:
I know we don't usually update version pins except right after a Zope release, but for `pkg_resources` fixes I think we should make an exception.


dataflake (approved):
Oh, never mind, I looked at the release notes and it's fine.

dataflake:
I just triggered rebuilding the published configurations at https://zopefoundation.github.io/Zope/

PR #1257, opened 2025-03-20
Add preliminary support for Python 3.14

icemac:


icemac:
I have no idea where this ast.Str comes from: I searched zopefoundation and Chameleon 4.4.4 which is still used – no success.

So I am giving up here now. Anyone to take over?

dataflake:
`ast.Str` comes from Chameleon 4.4.4. The references were fixed in version 4.6.0. We cannot move to that version until https://github.com/malthe/chameleon/issues/438 is fixed - unfortunately I have not gotten any response at all.

dataflake:
This is now done, all remaining work for a Python 3.14-compatible release continues in other tasks like #1259.

PR #1265, opened 2025-09-25
Update to PEP-420 versions of zope.* namespace.

icemac:
Update to current zope.meta.


icemac:

PR #1269, opened 2025-10-26
Small release process improvements

dataflake:
I added an explicit step to delete documentation build artifacts because the current rules in `MANIFEST.in` will include a bunch of that. I have inadvertently generated release tarballs that are way too large (12.6 instead of 7.7 MB) due to including documentation build files for a long time.

I also switched from using `zopepy` to using `buildout setup` because `zopepy` requires running the buildout in place, which I rarely do anymore. I also removed the `egg_info -Db ''` part, I don't think it serves a useful purpose. If you believe it is important please let me know.

### DIFF
--- a/docs/maintenance.rst
+++ b/docs/maintenance.rst
@@ -64,6 +64,11 @@ Steps for creating a new Zope release
 
 - If the tests succeed, commit the changes.
 
+- Clear out documentation build artifacts to prevent adding these files to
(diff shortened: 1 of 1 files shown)


icemac:
I am always using `zest.releaser` to cut and update releases so I have no idea what purpose `egg_info -Db ''` could serve.

PR #1252, opened 2025-02-11
Upgrade to Chameleon 4.6.0.

mauritsvanrees:
We should use latest Chameleon (only possible on Python 3.9+) to avoid deprecation warnings for some changes in the `ast` module.  In Python 3.14 these will become errors.
See [chameleon PR](https://github.com/malthe/chameleon/pull/434).

Zope itself also uses some deprecated `ast` code in `Products.PageTemplates`.  And `z3c.pt` does it as well.  Maybe it needs to be fixed there first:

```
...Zope/src/Products/PageTemplates/engine.py:312: DeprecationWarning:
  ast.Str is deprecated and will be removed in Python 3.14;
  use ast.Constant instead
    type=ast.Str(self.type),
.../Zope/src/Products/PageTemplates/engine.py:313: DeprecationWarning:
  ast.Str is deprecated and will be removed in Python 3.14;
  use ast.Constant instead
    expression=ast.Str(self.expression),
...z3c/pt/expressions.py:178: DeprecationWarning:
  ast.Str is deprecated and will be removed in Python 3.14;
  use ast.Constant instead
    component = ast.Str(part)
```

At any rate the Zope tests currently fail with latest Chameleon:

```
Failure in test testBatchIteration (Products.PageTemplates.tests.testChameleonTalesExpressions.ChameleonTalesExpressionTests.testBatchIteration)
Traceback (most recent call last):
  File "/Users/maurits/.pyenv/versions/3.13.1/lib/python3.13/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/Users/maurits/.pyenv/versions/3.13.1/lib/python3.13/unittest/case.py", line 651, in run
    self._callTestMethod(testMethod)
  File "/Users/maurits/.pyenv/versions/3.13.1/lib/python3.13/unittest/case.py", line 606, in _callTestMethod
    if method() is not None:
  File "/Users/maurits/community/plone-coredev/6.2/src/Zope/src/Products/PageTemplates/tests/testHTMLTests.py", line 177, in testBatchIteration
    self.assert_expected(self.folder.t, 'CheckBatchIteration.html')
  File "/Users/maurits/community/plone-coredev/6.2/src/Zope/src/Products/PageTemplates/tests/testHTMLTests.py", line 96, in assert_expected
    assert not t._v_errors, 'Template errors: %s' % t._v_errors
[...]

PR #1272, opened 2025-11-12
Use PEP 420-compatible `zc` namespace packages

dataflake:
Upgrade `zc` namespace packages and `plone.versioncheck`. Removed obsolete Python 3.9 pins.


CHANGES.rst
```diff
@@ -7,8 +7,10 @@ since the branch point at Zope 4.1.2.
 The change log for the previous version, Zope 4, is at
 https://github.com/zopefoundation/Zope/blob/4.x/CHANGES.rst
 
-5.14 (unreleased)
------------------
+6.0 (unreleased)
```
icemac: Should we actually call it version 6.0 or would 5.14 be enough.

For a version 6 we should also apply for a new trove classifier, I think.

dataflake: I think that depends on what we do about horse-with-no-namespace.

If you add that as a dependency then a Zope upgrade won't break any deployment and it would be OK to call it 5.14.

If you don't add it as a dependency and choose to e.g. just document it as a band-aid if something breaks then it should be 6.0 because it will break existing deployments without additional steps like adding horse-with-no-namespace.

dataflake:
I get that buildout 5 would be a requirement for Zope 5.14/6.0. But at the same time earlier Zope releases must never ever use buildout 5 because their deployments would break immediately.

There is no buildout pin in Zope itself or in the build/install documentation we have. Almost everyone will have their deployments without buildout version pin and as soon as they do a simple `pip install -U zc.buildout`, boom.

I admit I never realized this would be the case because I didn't read closely enough. I dimly remember asking if horse-with-no-namespace should actually be a dependency of buildout itself, but didn't get any support there. That would have solved this issue.

mauritsvanrees:
I actually expect most people *will* have `zc.buildout` pinned.  It is what Plone does in its versions anyway.  But other projects may differ.

Using `zc.buildout` 5 should work fine on older Zope versions as well.  I just tried this on a Plone 6.0 buildout and it worked fine.

But in the Plone 6.0 core development buildout it will fail, because it uses checkouts of several packages that use `pkg_resources` namespaces.  For example, we of course have `Products.CMFPlone` as central package.  In a checkout this will fail because other `Products.*` packages will have been installed  by `zc.buildout` 5 as native namespaces.

With any `pip` version, the same should still work: pip has been basically installing all packages as native namespaces for years, and if you use a development/editable install with `pkg_resources` this still works fine.  You will only get ModuleNotFound errors when two editable package both use a different namespace style.

In other words, if you use `pip` to install Zope, `horse-with-no-namespace` is normally not needed.  So ideally, Zope should not add this dependency.

This also suggests that if `zc.buildout` 5 wants to be on the same level as `pip`, it should include `horse-with-no-namespace`.

I did not want to do this until now, because not all projects will need this.  And I can imagine that there are corner cases where `horse-with-no-namespace` does the wrong thing.  Or that all is well now, but a future `setuptools` version will break something that the horse does not expect.

I will make a PR for `zc.buildout`.  Probably pin `setuptools<81` as well.  This could land in `zc.buildout` 5.1.0.  Let's see.

mauritsvanrees:
> ... or an admonition for developers to either stick with buildout<5

For clarity: this won't work if they use any packages that share a pkg_resources-style namespace with what we have in Zope, even if everything is installed from PyPI.  Practically speaking: Plone 6.0 and 6.1 on `zc.buildout` 4 won't work with Zope 5.14/6.0.  Which may be okay: I was not planning on updating Plone 6.0/6.1 to use the new Zope.

Big picture overview of Plone versions:

* Plone 6.0 is in security support mode, so no big changes are planned here anyway.  So whatever we do in Zope or buildout does not have a direct effect.
* Plone 6.1 is in maintenance mode.  Since 6.1.2 it is using native namespaces for `five` and `z3c`.  I would prefer to keep it to those.  Theoretically `zope.*` should be okay, but those are a lot of packages.  `zc.*` would not be that many, but this would include `zc.buildout`, and I want to avoid forcing people to use a new major buildout version in a bugfix release of Plone 6.1.x.  Using this new Zope version with native namespaces for `Products.*` is not an option.  Which is fine.
* Plone 6.2 is under development, pre-alpha.  The intention is to use native namespaces everywhere.  So definitely this new Zope version.  And when using `zc.buildout` we will say you *must* use version 5.  And we will advice installing `horse-with-no-namespace`; at the very least that will be our first advice in case of a ModuleNotFound error.

So if this will be called Zope 5.14, both Plone 6.0 and 6.1 will stick to 5.13.  And then I hope that there can be a 5.13.x branch for occasional security fixes.

I am trying out adding `horse-with-no-namespace` as dependency of `zc.buildout`.  As expected, the biggest problem is fixing the doctests, mostly because they get some extra output from the horse. :-/

mauritsvanrees:
"Use the horse, Luke!"

Couldn't resist.

dataflake:
P.S.: If it's Zope 6 I would really like it to include Python 3.14 from the start, but there's still no solution to the Chameleon issue...

davisagli:
My 2 cents as the creator of horse-with-no-namespace: I always intended it to be something you can install temporarily in dev environments while you're working on converting all namespaces to PEP 420, and not a permanent production dependency.

dataflake:
@davisagli Yes, I absolutely get it. Cynics would call it a hack. But for many people it will be a godsend. That includes production deployments that suddenly break where there's not much time for diagnosis and the hack is a quick lifeline.

dataflake:
@mauritsvanrees @icemac I do think this is OK for merging so that we have it for #1252, right? It doesn't contain anything controversial, anyway.

PR #1282, opened 2026-01-14
Update zope.interface to 8.2 release.

icemac:
This seems to be necessary, to get https://github.com/zopefoundation/Products.ZCatalog/actions/runs/20894894972/job/60031842321 green again because we have a version conflict there.


dataflake:
So the main reason this happened is the extra dependency installation stuffed into some tox testenv sections for macOS. The Zope version pins were almost always specifying the latest deoendency releases, but right now there is a later zope.interface release out. This gets pre-installed by tox as unpinned dependency, which then collides with the Zope version pin for zope.interface later on.

I just did a manual check to see if this extra treatment for macOS is still necessary and it seems at least in Products.ZCatalog it is not. See the changes at https://github.com/zopefoundation/Products.ZCatalog/compare/dataflake/macos_compatibility_test?expand=1 (don't create the actual PR, just eyeball the changes, this needs cleanup), the test results are at https://github.com/zopefoundation/Products.ZCatalog/actions/runs/20987002165

The only other place where this special treatment is actually used is Zope itself. I will do a manual check there as well and if it works then I will remove this special treatment from the zope.meta templates, Products.ZCatalog and Zope.

So please leave this PR open, if my experiment works I will delete it because it's just a workaround.

dataflake:
The tests were successful so I will close this PR, it is superceded by https://github.com/zopefoundation/Zope/pull/1283 which just gets rid of the specialized macOS treatment.

PR #1283, opened 2026-01-14
Remove specialized tox test configuration for GitHub

dataflake:
It is no longer needed and can cause problems, see explanation in https://github.com/zopefoundation/Zope/pull/1282


icemac (approved):
Cool to see that this workaround is no longer needed!

PR #1290, opened 2026-03-23
Use ZODB new ``class_factory`` parameter to register class factory

### LINKED ISSUE #420: Class Factory is not really configurable from Zope
This is a bug in the integration between Zope and ZODB, in the way `classFactory` is set by Zope. `classFactory` is a function responsible for returning the class when loading pickles and it handling the cases where the class cannot be found.

ZODB comes with a [default class factory](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/DB.py#L849-L851) using [`ZODB.broken.find_global`](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/broken.py#L136-L223), which basically returns a `ZODB.broken.Broken` class in case of `ImportError`.

As the comment in ZODB's default class factory explains, Zope rebinds this method at runtime, by [patching the db instance during startup](https://github.com/zopefoundation/Zope/blob/36cfb1a4d7603e0950a4befa833a245d1d9584c4/src/Zope2/App/startup.py#L120). Zope implementation of [class factory is slightly different](https://github.com/zopefoundation/Zope/blob/36cfb1a4d7603e0950a4befa833a245d1d9584c4/src/Zope2/App/ClassFactory.py#L17-L22), it returns an `OFS.Uninstalled.Broken` class in case of any `Exception`.

This difference in exception handling caused a bug in my application code, Zope's version handles the case where the module python file exists on disk but cannot be loaded. In my case the python file was present on disk, but not supposed to be used, it was for another python version and loading it with this python version caused a `SyntaxError` - not an `ImportError`. When using Zope's class factory, the syntax error was caught and a broken object was returned, but when using ZODB's class factory was used, the error propagated. I could observe the two behaviors a bit randomly after investigation I figure out that not all ZODB connections are using Zope's class factory, the Zope approach of patching does not full work.

The problematic timeline with the Zope integration is during Zope startup:

* When [loading configuration](https://github.com/zopefoundation/Zope/blob/36cfb1a4d7603e0950a4befa833a245d1d9584c4/src/Zope2/Startup/datatypes.py#L237), Zope will implicitly open a database.
* At the end of `DB.__init__`, a connection is opened to ensure the root object exists, in [`src/ZODB/DB.py` lines 468-475](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/DB.py#L468-L475).
* When a `Connection` is created, it stores a direct reference to `db.classFactory` in its `ObjectReader`, in [`src/ZODB/Connection.py` line 198](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/Connection.py#L198). The `ObjectReader` stores this as `self._factory` and uses it to resolve classes when unpickling, in [`src/ZODB/serialize.py` lines 473-488](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/serialize.py#L473-L488), so at this point, this initial connection's `_reader._factory` holds a reference to the **default** `DB.classFactory` method from ZODB (which delegates to `ZODB.broken.find_global` and only handles `ImportError`)
* The `ContextManager.__exit__` commits and then calls `conn.close()`, which returns the connection to `self.pool` via `DB._returnToPool()`, [`src/ZODB/DB.py` lines 1028-1033](https://github.com/zopefoundation/ZODB/blob/21ae6a771901a1e4c71c71f873e89f8cd4beff41/src/ZODB/DB.py#L1028-L1033)
* Later, Zope patches the db instance, which will affect all new connections.

After this sequence, the pool contains a connection with ZODB's class factory and new connections will use Zope's class factory and the behavior will be different depending on which connection from the pool is used.

This seems to have changed with in ZODB 5.0.0, with https://github.com/zopefoundation/ZODB/commit/03b81838a52c4d723c719fb3bed7d7e21a8cd594 .

It's also worth noting that Zope's provide support for [configuring the class factory in zcml](https://github.com/zopefoundation/Zope/blob/36cfb1a4d7603e0950a4befa833a245d1d9584c4/src/Zope2/Startup/wsgischema.xml#L64-L71) but this is not effective at all because the patching happening later.

To fix this, I'm thinking that we could change the DB constructor in ZODB so that it accepts an explicit `class_factory` parameter and change Zope to use this new parameter. Zope's ineffective ZCML to configure the class factory could also be moved to ZODB.

perrinjerome:
- [x] I signed and returned the [Zope Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the zopefoundation GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Developer guidelines](https://www.zope.dev/developer/guidelines.html).
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] If needed, I added new tests for my changes.
- [x] If needed, I added documentation for my changes.
- [x] I included a change log entry in my commits.

-----

If your pull request closes an open issue, include the exact text below, immediately followed by the issue number. When your pull request gets merged, then that issue will close automatically.

Together with https://github.com/zopefoundation/ZODB/pull/421 
Closes https://github.com/zopefoundation/ZODB/issues/420


perrinjerome:
still draft because it will need a new ZODB version

perrinjerome:
I included changes to version-prod.cfg and the corresponding generated constraints files to depend on the new ZODB version

dataflake (approved):
LGTM

drfho:
FYI: Merge results in issue https://github.com/zopefoundation/Zope/issues/1291

perrinjerome:
@dataflake thank you very much for making the releases.

PR #1294, opened 2026-06-01
Update to Python 3.15

icemac:
Current blockers:

- [ ] https://github.com/shibukawa/imagesize_py/issues/85 `imagesize` requires `Python < 3.15`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,6 +31,7 @@ jobs:
         - ["3.12", "py312"]
         - ["3.13", "py313"]
         - ["3.14", "py314"]
+        - ["3.15", "py315"]
         - ["3.11", "docs"]
         - ["3.11", "coverage"]
         exclude:
@@ -66,3 +67,32 @@ jobs:
         uvx coveralls --service=github
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build package artifacts
+      if: >
+        matrix.os[0] == 'ubuntu'
+        && matrix.config[1] == 'py314'
+      run: |
+        rm -f dist/*
+        pip install -U packaging "setuptools >= 78.1.1,< 82" wheel twine
+        python setup.py sdist
+        python setup.py bdist_wheel
+
+    - name: Upload package wheel
+      if: >
+        matrix.os[0] == 'ubuntu'
+        && matrix.config[1] == 'py314'
+      uses: actions/upload-artifact@v7
+      with:
+        name: Zope.whl
+        path: dist/*whl
+
+    - name: Upload package source distribution
+      if: >
+        matrix.os[0] == 'ubuntu'
+        && matrix.config[1] == 'py314'
+      uses: actions/upload-artifact@v7
+      with:
+        name: Zope.tar.gz
+        path: dist/*gz
+
--- a/.meta.toml
+++ b/.meta.toml
@@ -2,7 +2,7 @@
 # https://github.com/zopefoundation/meta/tree/master/src/zope/meta/zope-product
 [meta]
 template = "zope-product"
-commit-id = "ba10c93f"
+commit-id = "516c594a"
 
 [python]
 with-pypy = false
@@ -10,7 +10,7 @@ with-docs = true
 with-sphinx-doctests = false
 with-windows = true
 with-macos = true
-with-future-python = false
+with-future-python = true
 with-free-threaded-python = false
 
 [coverage]
--- a/constraints.txt
+++ b/constraints.txt
@@ -19,7 +19,11 @@ Products.BTreeFolder2==6.0
 Products.ZCatalog==7.3
 Pygments==2.20.0
 Record==4.2
-RestrictedPython==8.1
+RestrictedPython==8.1; python_version == '3.10'
+RestrictedPython==8.1; python_version == '3.11'
+RestrictedPython==8.1; python_version == '3.12'
+RestrictedPython==8.1; python_version == '3.13'
+RestrictedPython==8.3a1.dev0; python_version > '3.13'
 Sphinx==8.1.3; python_version == '3.10'
 Sphinx==9.0.4; python_version == '3.11'
 Sphinx==9.1.0; python_version > '3.11'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 # https://github.com/zopefoundation/meta/tree/master/src/zope/meta/zope-product
 [build-system]
 requires = [
-    "setuptools >= 78.1.1,< 81",
+    "setuptools >= 78.1.1,< 82",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -11,7 +11,11 @@ MultiMapping==5.1
 Paste==3.10.1
 PasteDeploy==3.1.0
 Persistence==5.4
-RestrictedPython==8.1
+RestrictedPython==8.1; python_version == '3.10'
+RestrictedPython==8.1; python_version == '3.11'
+RestrictedPython==8.1; python_version == '3.12'
+RestrictedPython==8.1; python_version == '3.13'
+RestrictedPython==8.3a1.dev0; python_version > '3.13'
 WSGIProxy2==0.5.1
 WebOb==1.8.9
 WebTest==3.0.7
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 # Generated with zope.meta (https://zopemeta.readthedocs.io/) from:
 # https://github.com/zopefoundation/meta/tree/master/src/zope/meta/zope-product
-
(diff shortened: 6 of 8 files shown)


icemac:
cc @dataflake Just FYI. I do not know which problems will show up when `imagesize` is resolved.

mauritsvanrees:
Where is the `imagesize` package used?  My `grep` skills are failing me.  Well, I see the version pin, I just don't see it anywhere else.  I also don't see the package in `bin/*` in a Plone buildout.

dataflake:
> Where is the `imagesize` package used? My `grep` skills are failing me. Well, I see the version pin, I just don't see it anywhere else. I also don't see the package in `bin/*` in a Plone buildout.

Any added pin nowadays is usually in response to buildout complaining about an unpinned dependency. In this case it's a Sphinx 9.1.0 dependency. Commenting out the pin and re-running the Zope buildout will always tell you.

dataflake:
P.S.: I think the script `/bin/imagesize` doesn't show because the buildout configuration that pulls in Sphinx explicitly specifies to only build `bin/sphinx-build`:


```ini
[sphinx]
=> make-docs
recipe = zc.recipe.egg
eggs =
    Zope[docs]
    Sphinx
scripts =
    sphinx-build
```

dataflake:
I am proposing a simple fix to make this work and prevent a package that sees very little maintenance (`imagesize`) from taking our release process hostage:

If we remove the buildout steps `sphinx` and `make-docs` it will all work. I don't know what function these steps still have, if you want to build out the documentation you can simply use `bin/tox -e docs` - this is actually our documented way of doing so, see https://zope.readthedocs.io/en/latest/INSTALL.html#building-the-documentation.

Only slightly related side note: I wonder how in the world `Sphinx` declares Python 3.15 support when they also declare a direct dependency on `imagesize`, which prevents installing the whole thing.

If you all agree with this fix then this PR is ready for review.

icemac:
> If we remove the buildout steps `sphinx` and `make-docs` it will all work. I don't know what function these steps still have, if you want to build out the documentation you can simply use `bin/tox -e docs` - this is actually our documented way of doing so, see https://zope.readthedocs.io/en/latest/INSTALL.html#building-the-documentation.

👍 

> Only slightly related side note: I wonder how in the world `Sphinx` declares Python 3.15 support when they also declare a direct dependency on `imagesize`, which prevents installing the whole thing.

They pin in their `uv.lock` version 1.4.1, see https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/uv.lock#L338 That version has no upper Python version bound.

dataflake (approved):

mauritsvanrees (approved):
Makes sense.

mauritsvanrees:
FWIW, `imagesize` 2.0.1 has been released, which declares Python 3.15 support.

dataflake:
@mauritsvanrees Updating all dependencies today I found another oddity on Python 3.10 and 3.11, full traceback see below. Python 3.12 and up don't show this problem.

The latest `anyio` release declares a dependency `typing_extensions>=4.16.0`. But `setuptools` <81 - in this case version 80.9.0 - has a vendor import of version 4.12.2. You can see at the top of the log output how the build dependency `setuptools<82,>=78.1.1` is installed. The already installed setuptools version 80.9.0 fulfills the requirement so it's left untouched. I am guessing this existing install is done by `tox` itself as part of creating the venv. If I force-update the venv to `setuptools<82,>=81` and get version 81.0.0 everything works again, or if I stick with `anyio==4.14.2`.

```
py310: install_deps> python -I -m pip install 'setuptools<82,>=78.1.1' wheel zc.buildout
py310: commands_pre[0]> .tox/py310/bin/buildout -c /Users/jens/src/zope/Zope/buildout.cfg buildout:directory=/Users/jens/src/zope/Zope/.tox/py310 buildout:develop=/Users/jens/src/zope/Zope buildout:root-directory=/Users/jens/src/zope/Zope
/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/zc/buildout/__init__.py:33: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import PkgResourcesDeprecationWarning
mr.developer: Creating missing sources dir /Users/jens/src/zope/Zope/.tox/py310/develop.
While:
  Installing.
  Loading extensions.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/zc/buildout/buildout.py", line 2376, in main
    getattr(buildout, command)(args)
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/zc/buildout/buildout.py", line 728, in install
    self._load_extensions()
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/zc/buildout/buildout.py", line 1298, in _load_extensions
    ep.load()(self)
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2776, in load
    self.require(*args, **kwargs)  # type: ignore[arg-type]
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2804, in require
    items = working_set.resolve(reqs, env, installer, extras=self.extras)
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/pkg_resources/__init__.py", line 892, in resolve
    dist = self._resolve_dist(
  File "/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/pkg_resources/__init__.py", line 938, in _resolve_dist
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (typing-extensions 4.12.2 (/Users/jens/src/zope/Zope/.tox/py310/lib/python3.10/site-packages/setuptools/_vendor), Requirement.parse('typing_extensions>=4.16.0; python_version < "3.15"'), {'anyio'})
py310: exit 1 (0.76 seconds) /Users/jens/src/zope/Zope> .tox/py310/bin/buildout -c /Users/jens/src/zope/Zope/buildout.cfg buildout:directory=/Users/jens/src/zope/Zope/.tox/py310 buildout:develop=/Users/jens/src/zope/Zope buildout:root-directory=/Users/jens/src/zope/Zope pid=86425
  py310: FAIL code 1 (1.48=setup[0.72]+cmd[0.76] seconds)
  evaluation failed :( (1.49 seconds)
```

mauritsvanrees:
@dataflake On current Zope master, when I change `anyio` to 4.15.1 in `versions.cfg` and `constraints.txt`, and run `tox -e py310`, all is fine.

dataflake:
Did you do `rm -rf .tox/py310` so the venv gets rebuilt?

mauritsvanrees:
Ah, but I already get `setuptools` 81:

```
$ ls -1d .tox/py310/lib/python3.10/site-packages/setuptools*
.tox/py310/lib/python3.10/site-packages/setuptools
.tox/py310/lib/python3.10/site-packages/setuptools-81.0.0.dist-info
```

When I edit `tox.ini` to use `setuptools >= 78.1.1,< 80` it indeed fails.

mauritsvanrees:
This is the first setuptools version where it works for me:

```
[testenv]
skip_install = true
deps =
    setuptools==80.10.1
    zc.buildout
    wheel
```

I would say it is fine to use this as minimum in `tox.ini`.
Well, spelled as `setuptools >=80.10.1,<82`.

dataflake:
Did you manage to reproduce the issue on a clean Zope master when tox rebuilds the Python 3.10/3.11 venv?

mauritsvanrees:
> Did you manage to reproduce the issue on a clean Zope master when tox rebuilds the Python 3.10/3.11 venv?

On a clean Zope master it runs fine, because I get setuptools 81.

I guess it depends on how you have installed Python, which setuptools is available there by default, maybe which tox version you have.

mauritsvanrees:
Actually, as a test this works too, but it may be a bridge too far with the zc.buildout alpha version:

```
[testenv]
skip_install = true
deps =
    setuptools==84
    zc.buildout==6.0.0a1
    wheel
```

That buildout version is needed as minimum for the optional `setuptools-latest` env though:

```
[testenv:setuptools-latest]
basepython = python3
deps =
    git+https://github.com/pypa/setuptools.git\#egg=setuptools
    zc.buildout>=6.0.0a1
    wheel
```

dataflake:
This is all very odd.

The version of `setuptools` in my manually compiled Python installs seems to make no difference here, I get the same results even after manually updating the setuptools version there.

When `tox` (latest version) creates the py310/py311 environments it just always grabs setuptools 80.9.0 while py312-315 get 81.0.0.

Anyway, the GHA runners don't seem to have this issue and the tests are green there. Good enough for now. Thank you for your ideas!

